PR #54, opened 2019-12-16
The brand picker and the actual brand can become off-sync if the brand is changed from another tab

alexasselin008:
## Issue
The brand picker and the actual brand can become off-sync if the brand is changed from another tab.
image

## What I did

Modified the BrandPickerDocsWidget to react to changes from the local storage. I made sure that the brand wasn't applied on the first mount since we don't want to overwrite css variables with the values that we're already applied.

## How to test

- Is this testable with Jest or Chromatic screenshots? No

### DIFF
--- a/storybook/addons/brand-picker/brand-picker-docs-widget.jsx
+++ b/storybook/addons/brand-picker/brand-picker-docs-widget.jsx
@@ -15,7 +15,6 @@ export function BrandPickerDocsWidget() {
     };
 
     const handleSelectBrand = brand => {
-        applyBrand(brand);
         setCurrentBrand(brand);
     };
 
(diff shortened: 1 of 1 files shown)


patricklafrance:
@alexasselin008 peux-tu fixer le build svp.

storybook/addons/brand-picker/brand-picker-docs-widget.jsx
```diff
@@ -13,10 +15,16 @@ export function BrandPickerDocsWidget() {
     };
 
     const handleSelectBrand = brand => {
-        applyBrand(brand);
         setCurrentBrand(brand);
     };
 
+    useEffect(() => {
+        if(previousBrand.current && previousBrand.current.id !== currentBrand.id) {
```
patricklafrance: Le if ne me semble pas respecter nos guidelines, pourtant le build ne la pas flagguer?

Issue #66, opened 2019-12-19
A searchable dropdown shouldn't be fluid by default


patricklafrance:
### Describe the bug

A searchable dropdown shouldn't be fluid when the `fluid` prop is not set.

As we can see in the following screenshot, this is not the default behavior of SUI:

![image](https://user-images.githubusercontent.com/794579/71133876-2967cb80-21ca-11ea-8d6c-dcf1360d87c1.png)

https://semantic-ui.com/modules/dropdown.html#search-selection

### Steps to reproduce

```
        <Dropdown
            placeholder="Select language"
            selection
            search
            options={[
                {
                    "key": "Arabic",
                    "text": "Arabic",
                    "value": "Arabic"
                },
                {
                    "key": "Chinese",
                    "text": "Chinese",
                    "value": "Chinese"
                },
                {
                    "key": "Danish",
                    "text": "Danish",
                    "value": "Danish"
                },
                {
                    "key": "Dutch",
                    "text": "Dutch",
                    "value": "Dutch"
                },
                {
                    "key": "English",
                    "text": "English",
                    "value": "English"
                },
                {
                    "key": "French",
                    "text": "French",
                    "value": "French"
                },
                {
                    "key": "German",
                    "text": "German",
                    "value": "German"
                },
                {
                    "key": "Greek",
                    "text": "Greek",
                    "value": "Greek"
                },
                {
                    "key": "Hungarian",
                    "text": "Hungarian",
                    "value": "Hungarian"
                },
                {
                    "key": "Italian",
                    "text": "Italian",
                    "value": "Italian"
                },
                {
                    "key": "Japanese",
                    "text": "Japanese",
                    "value": "Japanese"
                },
                {
                    "key": "Korean",
                    "text": "Korean",
                    "value": "Korean"
                },
                {
                    "key": "Lithuanian",
                    "text": "Lithuanian",
                    "value": "Lithuanian"
                },
                {
                    "key": "Persian",
                    "text": "Persian",
                    "value": "Persian"
                },
                {
                    "key": "Polish",
                    "text": "Polish",
                    "value": "Polish"
                },
                {
                    "key": "Portuguese",
                    "text": "Portuguese",
                    "value": "Portuguese"
                },
                {
                    "key": "Russian",
                    "text": "Russian",
                    "value": "Russian"
                },
                {
                    "key": "Spanish",
                    "text": "Spanish",
                    "value": "Spanish"
                },
                {
                    "key": "Swedish",
[...]

Issue #62, opened 2019-12-17
Clearable dropdown doesn't display the right icon when a value is selected


patricklafrance:
### Describe the bug

A clearable dropdown should display a "clear" icon when a value is selected. Instead, an arrow is displated.

(Write your answer here.)
![image](https://user-images.githubusercontent.com/794579/71029234-f4317f80-20dc-11ea-9f4f-8fb2c4599240.png)

Here's what it looks like with native SUI:

![image](https://user-images.githubusercontent.com/794579/71029249-01e70500-20dd-11ea-9d6a-02c591cdcb0b.png)

https://react.semantic-ui.com/modules/dropdown/#types-clearable

### Steps to reproduce

```
        <Dropdown
            placeholder="Gender"
            selection
            clearable
            options={[
                {
                    key: "Male",
                    text: "Male",
                    value: "Male"
                },
                {
                    key: "Female",
                    text: "Female",
                    value: "Female"
                }
            ]}
        />
```

### Reproducible demo

https://sg-orbit.netlify.com/?path=/docs/components-dropdown--default-story#usage

Issue #61, opened 2019-12-17
Searchable dropdown placeholder should have less emphasis when the search input is focused


patricklafrance:
### Describe the bug

A searchable SUI dropdown placeholder should have less emphasis when the input is focused.

Here what it looks like in our toolkit:

![image](https://user-images.githubusercontent.com/794579/71028901-35755f80-20dc-11ea-8d03-ea79ba704228.png)

Here what it looks like with the default SUI:

![image](https://user-images.githubusercontent.com/794579/71028929-445c1200-20dc-11ea-863b-a39edf6618d2.png)

### Reproducible demo

https://sg-orbit.netlify.com/?path=/docs/components-dropdown--default-story#usage

Issue #60, opened 2019-12-17
Searchable dropdown search input is not vertically centered


patricklafrance:
### Describe the bug

A SUI search dropdown input is not vertically centered.

![image](https://user-images.githubusercontent.com/794579/71028609-9f413980-20db-11ea-9872-7d31fe3962f4.png)

### Steps to reproduce

```
        <Dropdown
            placeholder="Gender"
            selection
            search
            options={[
                {
                    key: "Male",
                    text: "Male",
                    value: "Male"
                },
                {
                    key: "Female",
                    text: "Female",
                    value: "Female"
                }
            ]}
        />
```

### Reproducible demo

https://sg-orbit.netlify.com/?path=/docs/components-dropdown--default-story#usage

PR #97, opened 2020-01-08
Feature/dropdown semantic

fraincs:
Issue:
https://github.com/gsoft-inc/sg-orbit/issues/60, https://github.com/gsoft-inc/sg-orbit/issues/61, https://github.com/gsoft-inc/sg-orbit/issues/62, https://github.com/gsoft-inc/sg-orbit/issues/66

## What I did

Fixed multiple bug fixes on Semantic UI Dropdown themes


patricklafrance:
The X button to clear the value of a dropdown should be consistent with our other components, it should like this:

image

View: https://sg-orbit.netlify.com/?path=/docs/components-search-input-results-set--default-story#selected-value

patricklafrance:
I believe a "standard dropdown" should have have less emphasis when navigating through the choices.

Same issue as https://github.com/gsoft-inc/sg-orbit/issues/61 but for the "standard"

patricklafrance:
Otherwise it's 🔥 good job!

fraincs:
> The X button to clear the value of a dropdown should be consistent with our other components, it should look like this:
> 
> image
> 
> View: https://sg-orbit.netlify.com/?path=/docs/components-search-input-results-set--default-story#selected-value

The problem here is that clearable dropdowns are using a background image and not a button to close. This means we cannot mimic the same behaviour.

fraincs:
> I believe a "standard dropdown" should have have less emphasis when navigating through the choices.
> 
> Same issue as #61 but for the "standard"

Since it this case it's a choice that has been made and not typing over an input, I think we should keep doing it the Semantic way : https://react.semantic-ui.com/modules/dropdown/

patricklafrance:
> > I believe a "standard dropdown" should have have less emphasis when navigating through the choices.
> > Same issue as #61 but for the "standard"
> 
> Since it this case it's a choice that has been made and not typing over an input, I think we should keep doing it the Semantic way : https://react.semantic-ui.com/modules/dropdown/

Not sure I see the difference between this case and the searchable? Maybe in SUI react it's a bug and it's not the actual intention? Anyway, seems weird to me but if you want to keep it like this I won't block you! :)

patricklafrance:
@fraincs do we need additional chromatic test cases for the dropdown?

fraincs:
> @fraincs do we need additional chromatic test cases for the dropdown?

Since no new components have been added I would say no. Could we add more, probably, but this would not be linked to this story.

Issue #127, opened 2020-01-17
Basic circular labels are not the right size


fraincs:
### Describe the bug

When basic, labels are applied a border, which affects elements sizes. A default circular label is 44px instead of 40.

A possible solution would be to use a box shadow, like buttons do, instead of a border on basic labels since box shadow does not affect an element width.

fraincs:
Fixed in PR https://github.com/gsoft-inc/sg-orbit/pull/138

Issue #103, opened 2020-01-09
Ghost buttons should support icons


patricklafrance:
### Describe the bug

Ghost buttons should support icons.

![image](https://user-images.githubusercontent.com/794579/72110094-632d7d00-3305-11ea-9263-655864438d55.png)

Issue #102, opened 2020-01-09
Circular button sizes should also work for non icon buttons


patricklafrance:
### Describe the bug
Sized forr circular buttons works fine for circular buttons that contains an icon. Doesn't work for circular buttons that contains only text.

![image](https://user-images.githubusercontent.com/794579/72104819-3cb61480-32fa-11ea-9b1c-e0270d5f9e62.png)

Issue #68, opened 2019-12-19
Button component sizes should follow in place naming convention


patricklafrance:
Current sizes for buttons are "short" and "tall":

```
<Button className="short">Short</Button>
<Button className="tall">Tall button</Button>
```

Other SUI components follow "small", "medium", "large" ... naming convention. Buttons should do the same.

fraincs:
Should we target a major release?

PR #138, opened 2020-01-21
Semantic ui buttons revamp

fraincs:
Issue:

Buttons styles were all over the place.

## What I did

Revamp all the css from the semantic ui theme button. Added way more Chromatic tests.

Fixed
https://github.com/gsoft-inc/sg-orbit/issues/127
https://github.com/gsoft-inc/sg-orbit/issues/103
https://github.com/gsoft-inc/sg-orbit/issues/102
https://github.com/gsoft-inc/sg-orbit/issues/68


patricklafrance:
Looks great!

Could you please add the matching issues in the PR description?

patricklafrance:
The following cases are broke in the docs:

image

patricklafrance:
Also, please remove the dimension classes on the Icons in the docs examples since it's not required anymore.

patricklafrance:
The naked chromatic snapshot should includes a few "custom colors" override to make the styles can be overrided

patricklafrance:
For groups example in the docs, I understand that the groups are not 100% ready but look at the icon with a tag, it the margins are not good. From what I understand a button with a label should work at this point?

fraincs:
> For the group example in the docs, I understand that the groups are not 100% ready but look at the icon with a tag, the margins are not good. From what I understand a button with a label should work at this point?
> 
> You should also add a few chromatic tests for this.

Would you suggest a separate test?

fraincs:
> The naked chromatic snapshot should includes a few "custom colors" override to make the styles can be overrided

I would create another shot, in order not to add unnecessary complexity in the tests.

packages/react-components/components/button/tests/chromatic/buttons.jsx
```diff
@@ -0,0 +1,138 @@
+import { Button } from "@orbit-ui/react-button/src";
+import { CalendarIcon, CalendarIcon24, SignoutIcon24 } from "@orbit-ui/icons";
+import { Label } from "@orbit-ui/react-label/src";
+
+export const ThemeButtons = props => {
```
patricklafrance: Rename to Buttons

packages/react-components/components/button/tests/chromatic/buttons.jsx
```diff
@@ -0,0 +1,138 @@
...
+    return (
+        <>
```
patricklafrance: Fragment is not required here

Issue #147, opened 2020-01-24
Medium input, button and label font-size should be 16px


patricklafrance:
![image](https://user-images.githubusercontent.com/794579/73073953-80765580-3e86-11ea-8b93-faf51563bc8c.png)

patricklafrance:
@fraincs I updated the issue title. After talking to juju we decided to update the font-size of all medium basic components from 18px to 16px.

fraincs:
Perfect, makes sense.

fraincs:
This means we keep large at 18px?

patricklafrance:
Yes

Issue #113, opened 2020-01-15
A label with a nested button should alter is right padding


patricklafrance:
### Describe the bug

When a label have a nested button, it should alter is right padding to 8px

### Expected behavior

*Theme*

Add a CSS class "button" to the label.

*Label component*

Add an `button` prop that will add the CSS class.

-->

(Write a link to your reproducible demo)

fraincs:
I think we should add a `with-button` class to prevent a clash with the semantic button classes.

Issue #112, opened 2020-01-15
A label with a nested icon should alter is right padding


patricklafrance:
### Describe the bug

When a label have a nested icon, it should alter is right padding to 8px

### Expected behavior

*Theme*

Add a CSS class "icon" to the label.

*Label component*

Add an `icon` prop that will add the CSS class.

-->

(Write a link to your reproducible demo)

fraincs:
What about when icon is at the start of a label?

https://sg-orbit.netlify.com//?path=/docs/components-label-standard--default-story#icon

Issue #106, opened 2020-01-14
Mini labels should not have a margin-right when not in a tag


fraincs:
### Describe the bug

Labels, when in a button, have a margin right of 8px, this is not an intended behaviour and is expected only when in a tag label.

### Steps to reproduce

1. Use a Label with size mini in a button
2. It has a right margin

### Expected behavior

No margin is applied to a label in a button

### Actual behavior

![image](https://user-images.githubusercontent.com/361632/72355579-c0368380-36b5-11ea-9911-f0eab10c4ae4.png)

patricklafrance:
Wondering if a label in a button should always have a LEFT margin of 8px?

fraincs:
A label in a button can be positioned on the right or left

![image](https://user-images.githubusercontent.com/361632/72362420-24127980-36c1-11ea-9425-97882756fed3.png)

Since Semantic buttons does not wrap the text in a button in an element it is complicated to manage. An option would be to accept a text property on our button and wrap it ourself.

patricklafrance:
Do we really want to support both right and left labels? What's our position on this from a design system standpoint?

Do we have any existing use cases?

patricklafrance:
Here's a suggestion:

- A tag in a button have a right margin of 8px (this is what you intent to do here)
- A label in a button have a left margin of 8px

PR #154, opened 2020-01-28
milestones Labels

fraincs:
Issue:
https://github.com/gsoft-inc/sg-orbit/issues/106
https://github.com/gsoft-inc/sg-orbit/issues/112
https://github.com/gsoft-inc/sg-orbit/issues/113
https://github.com/gsoft-inc/sg-orbit/issues/147

## What I did

- Unified font-sizes for medium components.
- Added a prop to Label to pass an Icon / this accepts an icon position too.
- Adjust a label padding / margin when it contains a button / icon.

### DIFF
--- a/packages/react-components/components/button/tests/chromatic/button.chroma.jsx
+++ b/packages/react-components/components/button/tests/chromatic/button.chroma.jsx
@@ -1,5 +1,6 @@
 import { Button } from "@orbit-ui/react-button/src";
 import { Buttons } from "./components";
+import { Label } from "@orbit-ui/react-label/src";
 import { createChromaticSection, paramsBuilder, storiesOfBuilder } from "@utils";
 
 import styles from "./styles.module.css";
@@ -74,7 +75,7 @@ stories()
              <div className="flex flex-row">
                  <div className="flex flex-column items-start">
                      <Button.Group>
-                         <Button>One</Button>
+                         <Button>One <Label size="mini">12</Label></Button>
                          <Button>Two</Button>
                          <Button>Three</Button>
                      </Button.Group>
--- a/packages/react-components/components/button/tests/chromatic/button.chroma.jsx
+++ b/packages/react-components/components/button/tests/chromatic/button.chroma.jsx
@@ -194,10 +194,30 @@ stories()
                          <Button>Two</Button>
                          <Button>Three</Button>
                      </Button.Group>
-                     <Button.Group toggle basic>
-                         <Button>One</Button>
+                     <Button.Group basic>
+                         <Button disabled>One</Button>
+                         <Button disabled>Two</Button>
+                         <Button disabled>Three</Button>
+                     </Button.Group>
+                     <Button.Group basic>
+                         <Button disabled>One</Button>
                          <Button>Two</Button>
                          <Button>Three</Button>
+                     </Button.Group>
+                     <Button.Group basic>
+                         <Button>One</Button>
+                         <Button active>Two</Button>
+                         <Button>Three</Button>
+                     </Button.Group>
+                     <Button.Group basic>
+                         <Button active>One</Button>
+                         <Button>Two</Button>
+                         <Button>Three</Button>
+                     </Button.Group>
+                     <Button.Group basic>
+                         <Button>One</Button>
+                         <Button>Two</Button>
+                         <Button active>Three</Button>
                      </Button.Group>
                      <Button.Group basic widths="7">
                          <Button>1</Button>
--- a/packages/react-components/components/label/src/label.jsx
+++ b/packages/react-components/components/label/src/label.jsx
@@ -1,5 +1,5 @@
 import { Ref, Label as SemanticLabel } from "semantic-ui-react";
-import { bool, func, object, oneOfType, string } from "prop-types";
+import { bool, func, node, object, oneOfType, oneOf, string } from "prop-types";
 import { forwardRef } from "react";
 import { isNil } from "lodash";
 import { mergeClasses, throwWhenUnsupportedPropIsProvided } from "@orbit-ui/react-components-shared";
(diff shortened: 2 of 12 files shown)


packages/react-components/components/button/tests/chromatic/button.chroma.jsx
```diff
@@ -74,7 +75,7 @@ stories()
              <div className="flex flex-row">
                  <div className="flex flex-column items-start">
                      <Button.Group>
-                         <Button>One</Button>
+                         <Button>One <Label size="mini">12</Label></Button>
```
patricklafrance: I though that a button would enforce the size of the label based on is own size. I'm I too early, it's not done yet?

packages/react-components/components/label/src/label.jsx
```diff
@@ -11,6 +11,18 @@ const propTypes = {
...
+    /**
+     * A label can contain a button.
+     */
+    button: bool,
+    /**
+     * A label can contain an icon.
+     */
+    icon: node,
+    /**
+     * A icon can be positionned.
```
patricklafrance: An icon can appear on the left or right.

packages/react-components/components/label/src/label.jsx
```diff
@@ -22,10 +34,11 @@ const propTypes = {
 };
 
 const defaultProps = {
-    naked: false
+    naked: false,
+    button: false
```
patricklafrance: Add a default iconPosition ?

packages/react-components/components/label/src/label.jsx
```diff
@@ -39,10 +52,13 @@ export function PureLabel({ naked, className, forwardedRef, children, ...props }
...
         const classes = mergeClasses(
             naked && "naked",
+            button && "with-button",
+            icon && "with-icon",
+            iconPosition === "right" && "with-icon-right",
             className
         );
 
-        return <SemanticLabel className={classes} {...props}>{children}</SemanticLabel>;
+        return <SemanticLabel className={classes} {...props}>{icon && iconPosition !== "right" ? icon : null}{children}{icon && iconPosition === "right" ? icon : null}</SemanticLabel>;
```
patricklafrance: This code is fine, however hard to understand and scale, I would prefer something similar to the following:

```
const renderLabelContent = () => {
    if (!isNil(icon)) {
         if (iconPosition === "right") {
              return <>{children}{icon}</>
         }

         return <>{icon}{children}</>
    }

    return children;
};

const renderLabel = () => {
    const classes = mergeClasses(...);

    return <SemanticLabel className={classes} {...props}>{renderLabelContent()}</SemanticLabel>
};
```

P.S. you might not need the fragments in `renderLabelContent`

packages/react-components/components/button/tests/chromatic/button.chroma.jsx
fraincs: How would you do this? Pass a prop down the tree?

packages/react-components/components/label/src/label.jsx
patricklafrance: IMO, we should also add an "icon" class on the Icon component and in SUI theme we should style `.icon` instead of `svg`.

packages/react-components/components/button/tests/chromatic/button.chroma.jsx
patricklafrance: This is easy to do with `cloneElement`

packages/react-components/components/label/stories/standard/standard.stories.mdx
```diff
@@ -51,10 +51,10 @@ A label can include a button.
 
 <Preview>
     <Story name="button">
-        <Label>
+        <Label button>
```
patricklafrance: Should we take a similar approach to the icons and accept a Button component here? I am not sure how usage will be if we do this.

There is a few advantages though:

- We could enforce the button type e.g. ghost
- We could enforce the button size e.g. tiny

While writing this I am wondering if maybe instead we should only accept the property `buttonIcon` and enforce everything else?

patricklafrance:
I am wondering if the prop `tag` should generate the nested label required to form a tag? 

Maybe we should extract Tag in it's own component? Usage feels a bit off if you have to compose 2 nested label components to form a tag.

What do you think?

fraincs:
> I am wondering if the prop `tag` should generate the nested label required to form a tag?
> 
> Maybe we should extract Tag in it's own component? Usage feels a bit off if you have to compose 2 nested label components to form a tag.
> 
> What do you think?

Makes sense. Design and code wise.

packages/react-components/components/label/tests/chromatic/label.chroma.jsx
```diff
@@ -46,76 +51,76 @@ stories()
...
                  <div className="flex flex-column items-start">
                      <Label tag size="large">
                          <Label circular empty size="tiny" color="red" />
                              Group added
                      </Label>
-                     <Label tag size="large">
+                     <Label tag size="large" button>
                          <Label circular empty size="tiny" color="red" />
                              Group added
                          <Button icon circular size="tiny" ghost>
```
patricklafrance: Oh and BTW, when I hover on this ghost button, visually nothing change, bug?

fraincs: That's because it's on a gray background ;)

packages/react-components/components/button/tests/chromatic/button.chroma.jsx
patricklafrance: Extracted in a separate issue: https://github.com/gsoft-inc/sg-orbit/issues/156

packages/react-components/components/label/stories/standard/standard.stories.mdx
patricklafrance: Extracted to: https://github.com/gsoft-inc/sg-orbit/issues/158

Issue #161, opened 2020-01-29
Checkboxes are not centered when used without a label


fraincs:
### Describe the bug

When using a checkbox without a label the checkbox has no height and aligns itself to the top.

### Steps to reproduce

See Storybook doc https://sg-orbit.netlify.com/?path=/docs/components-checkbox--labelled#no-label

### Expected behavior

![image](https://user-images.githubusercontent.com/361632/73379227-e2bcc500-428f-11ea-8ec2-477d4926228d.png)

### Actual behavior

![image](https://user-images.githubusercontent.com/361632/73379240-e6504c00-428f-11ea-9a37-9b045168f13f.png)

PR #163, opened 2020-01-30
css for checkbox

fraincs:
Issue:
https://github.com/gsoft-inc/sg-orbit/issues/161

## What I did

When no label is present in a checkbox semantic adds a class of fitted. I'm taking advantage or this and remove the centering of the checkbox in this situation.

How to test : 
Visit Storybook Doc in Checkbox


patricklafrance:
The changes looks good however I don't see any additional chromatic tests. Do we have a chromatic test for this variant of the checkbox?

fraincs:
> The changes looks good however I don't see any additional chromatic tests. Do we have a chromatic test for this variant of the checkbox?

We never had, I can add some though.

patricklafrance:
Yes please!

Issue #167, opened 2020-02-03
Support TextArea


patricklafrance:
### Is your feature request related to a problem?

We currently only offer theming for basic input. The textarea are not styled.

Semantic UI React offers an "addon" TextArea component. This means that SUI theme doesn't offer any styling for textarea.

I believe we should add a TextArea component to the @orbit-ui/react-inputs project and offer proper styling.

fraincs:
Are we planning on getting specs from designers?

patricklafrance:
If it's not the same as for the text input, of course.

PR #180, opened 2020-02-10
Issue/167

fraincs:
Issue:
https://github.com/gsoft-inc/sg-orbit/issues/167 https://github.com/gsoft-inc/sg-orbit/issues/181 https://github.com/gsoft-inc/sg-orbit/issues/198 https://github.com/gsoft-inc/sg-orbit/issues/200

TextArea was not supported in Orbit.

## What I did

Created a component TextArea which is a HOC of Semantic UI TextArea component. This component supports success, transparent, error, autofocusdelay

## How to test

Both.

Yes, documentation has been created.


packages/react-components/components/input/src/textarea.jsx
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable react/forbid-foreign-prop-types */
```
patricklafrance: you dont need to disable this rule since you dont override prop-types

packages/react-components/components/input/src/textarea.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+const defaultProps = {
+    autofocus: false,
+    error: false,
+    success: false,
+    focused: false,
+    transparent: false
+};
+
+function getTextAreaElement(textAreaRef) {
```
patricklafrance: you dont need this since you already have you ref element

packages/react-components/components/input/src/textarea.jsx
```diff
@@ -0,0 +1,111 @@
...
+    return textAreaRef.current;
+}
+
+function focus(textAreaRef) {
+    if (!isNil(textAreaRef.current)) {
+        getTextAreaElement(textAreaRef).focus();
```
patricklafrance: dont need to call getTextAreaElement, you already have the ref element that you need to focus

packages/react-components/components/input/src/textarea.jsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export function PureTextArea(props) {
+    const { autofocus, autofocusDelay, className, disabled, error, success, focused, transparent, children, forwardedRef, ...rest } = props;
+
+    throwWhenUnsupportedPropIsProvided(props, UNSUPPORTED_PROPS, "@orbit-ui/react-input");
+
+    const textAreaRef = useRef();
+
+    useImperativeHandle(forwardedRef, () => getTextAreaElement(textAreaRef));
```
patricklafrance: you don't have to use useImperativeHandle in this case, simply pass forwardRef to `Ref` instead of creating a proxy ref.

```
<Ref innerRef={forwardRef} ...
```

packages/react-components/components/input/src/textarea.jsx
```diff
@@ -0,0 +1,111 @@
...
+        };
+    }, [autofocus, autofocusDelay, disabled, textAreaRef]);
+}
+
+export function PureTextArea(props) {
+    const { autofocus, autofocusDelay, className, disabled, error, success, focused, transparent, children, forwardedRef, ...rest } = props;
+
+    throwWhenUnsupportedPropIsProvided(props, UNSUPPORTED_PROPS, "@orbit-ui/react-input");
+
+    const textAreaRef = useRef();
```
patricklafrance: Dont need to create a new Ref in this case.

patricklafrance:
Should we support `size`?

We already have `rows`  but it's not the same. `size` would also alter `font-size`. If you support `size` do we want to support `rows` ?

fraincs:
> Should we support `size`?
> 
> We already have `rows` but it's not the same. `size` would also alter `font-size`. If you support `size` do we want to support `rows` ?

rows is a default attribute of textarea

fraincs:
> Should we support `size`?

This is something to discuss with designers, I think it doesn't make sense height wise, but font sizes should match the input's font size, a different line-height might be a good idea too.

fraincs:
> Should we support `size`?

We should follow the same padding, font-size, as same size input, I will adjust before closing this PR.

fraincs:
A Jest test is failing, this seems to be the same issue as this: https://github.com/testing-library/jest-dom/issues/53

patricklafrance:
> A Jest test is failing, this seems to be the same issue as this: [testing-library/jest-dom#53](https://github.com/testing-library/jest-dom/issues/53)

Yes, we already use the [user-event](https://github.com/testing-library/user-event) library to mitigates that kind of problem.

I wonder if this is the same problem in your case though since the component you are testing is explicitly focusing on a dom element, this is not the same as triggering a click event and expecting the target element to have focus.

packages/react-components/components/input/stories/textarea/textarea.stories.mdx
```diff
@@ -133,6 +133,28 @@ A textarea can have a minimum number of rows.
...
+### Size
+
+A textarea can vary in size.
+
+Available sizes are: `"small"`, `"medium"` and `"large"`.
+
+<Preview>
+    <Story name="size">
+        <div className="flex flex-row items-end items-space-between">
+            <div className="flex-auto">
```
patricklafrance: Do you need flex-auto here?

In fact, I added it for the date picker but I don't understand why we need `flex-auto.`

patricklafrance:
Like an input, IMO a textarea shouldn't be fluid by default?

It should support a `fluid` prop.

packages/react-components/components/input/stories/textarea/textarea.stories.mdx
fraincs: See flex auto explanation on MDN : The item is sized according to its width and height properties, but grows to absorb any extra free space in the flex container,  this is the equivalent of "flex: 1 1 auto"

soit flex grow 1 shrink1 basis auto

Issue #199, opened 2020-02-19
Complex Dropdowns with image should implicitly set an avatar size


fraincs:
### Describe the bug

When using a complex dropdown with an avatar, the size of it has to be set by the end user.

![image](https://user-images.githubusercontent.com/361632/74868059-b55db700-5323-11ea-89c2-5b0ebbc9987d.png)

### Steps to reproduce

1. Open Storybook in Dropdown / Complex test.
2. Image avatar size have to be set explicitly.

### Expected behavior

Image avatar size follow a defined scale.

Issue #183, opened 2020-02-12
Dropdown selection list should be the same size of it's parent


fraincs:
Dropdown selection list should be the same size of it's parent: e.g. a small dropdown (32px high) should  have it's selection 32 px high

![image](https://user-images.githubusercontent.com/361632/74370028-686a6580-4da4-11ea-822b-dfc891b59f2c.png)

### Steps to reproduce

1. Open a small dropdown
2. The selection is 40px 

### Expected behavior
The selection is 32px 

### Reproducible demo

https://sg-orbit.netlify.com/?path=/docs/components-dropdown-standard--default-story#size

fraincs:
fixed in https://github.com/gsoft-inc/sg-orbit/pull/201

Issue #172, opened 2020-02-07
Dropdown component should support the size "large"


patricklafrance:
### Describe the bug

Currently most of our inputs support the sizes "small", "medium" and "large". The dropdown only support "small" and "medium".

We should also support "large" to make we offer a consistant experience

fraincs:
Fixed in https://github.com/gsoft-inc/sg-orbit/pull/201

Issue #160, opened 2020-01-29
Change Dropdown color bar behaviour


Gcaya:
### Describe the bug

Only a selected reason of creation should have a colored vertical bar to it's left. Currently any focused option in the dropdown displays the bar.

### Steps to reproduce

(Write your steps here:)

1. Go on the groups page
2. Open a group in the list
3. Make sure there is no reason of creation selected
4. Click on the EditableLabel to choose a reason of creation
5. Move around with arrow keys

### Expected behavior

Only the currently selected item should display a vertical colored bar. The bar should't move either with which ever option is focused. Plz don't mind the difference between "Internal project" and "Internal collaboration" it should display the same text.
![image](https://user-images.githubusercontent.com/8711024/73372905-efd4b680-4285-11ea-8bfb-422098d4431c.png)

### Actual behavior
![CBcNU8ivsf](https://user-images.githubusercontent.com/8711024/73374413-1f84be00-4288-11ea-9e55-693f07155cd4.gif)

(Write what happened. Please add screenshots!)

alexasselin008:
The parameter used for the dropdown are 

```
<Dropdown
        className="small"
        placeholder="Choose a reason of creation"
        open
        selection
        clearable=true
        onChange={handleChange}
        onOpen={handleOpen}
        onClose={handleClose}
        value={value}
        options={options}
        selectOnNavigation={false} 
        selectOnBlur={false}
    />
```

patricklafrance:
@fraincs can you confirm with the design team what is the correct behavior please?

patricklafrance:
@Gcaya I added "minor" and "blocking" tag to our tags list. Can you please add one of these to your issue based on how quick your issues need to be resolved.

patricklafrance:
Did you talk about this with the designers @fraincs ?

Gcaya:
No, only with @alexasselin008 who proposed to open this issue.

patricklafrance:
Sorry @Gcaya my message was intended to @fraincs since he was supposed to talk to the designers about this particular issue.

fraincs:
We agreed with designers, will be fixed on milestone Dropdowns

fraincs:
Fixed in https://github.com/gsoft-inc/sg-orbit/pull/201

PR #201, opened 2020-02-19
Support for size variants - better highlight on hover / navigation

fraincs:
Issue:

- [Change Dropdown color bar behaviour](https://github.com/gsoft-inc/sg-orbit/issues/160)
- [Dropdown selection list should be the same size of it's parent](https://github.com/gsoft-inc/sg-orbit/issues/183)
- [Support for large dropdowns](https://github.com/gsoft-inc/sg-orbit/issues/172)

## What I did

- CSS changes in order to accommodate a large dropdown variant.
- CSS to change height of dropdown selection in order to follow the same height as it's parent/trigger
- Added a small avatar in complex large dropdowns (see https://react.semantic-ui.com/elements/image/) a better way to deal with this would be to fix this [issue](https://github.com/gsoft-inc/sg-orbit/issues/199)
- CSS to support hovered but not selected option

## How to test

Documentation has been updated.
Chromatic tests are updated and split in two chromatic tests default and complex


storybook/stories/semantic-ui/react/dropdown/standard.stories.mdx
```diff
@@ -256,14 +256,15 @@ A dropdown can take the width of its container.
 
 A dropdown can vary in size.
 
-Available sizes are: "small" and "medium".
+Available sizes are: `small`, `medium` and `large`.
```
patricklafrance: Change for:

"Can be set to: "small", "large" or omitted." to be consistant with the other pages.

image

patricklafrance:
Can you create issues with the next steps please. My guess is something like this:

Create a custom dropdown HOC that will:

- Support a `size` prop
- Customize the default behavior of the dropdown `selectOnBlur={false}` & `selectOnNavigation={false}`

patricklafrance:
Also please accept the chromatic changes.

fraincs:
> Can you create issues with the next steps please. My guess is something like this:
> 
> Create a custom dropdown HOC that will:
> 
> * Support a `size` prop
> * Customize the default behavior of the dropdown `selectOnBlur={false}` & `selectOnNavigation={false}`

There's already an issue for the default behavior HOC should't we expand on this one?

patricklafrance:
> > Can you create issues with the next steps please. My guess is something like this:
> > Create a custom dropdown HOC that will:
> > 
> > * Support a `size` prop
> > * Customize the default behavior of the dropdown `selectOnBlur={false}` & `selectOnNavigation={false}`
> 
> There's already an [issue](https://github.com/gsoft-inc/sg-orbit/issues/184) for the default behavior HOC should't we expand on this one?

Ok forget what I wrote then!

Issue #214, opened 2020-02-21
Dropdown action item are not styled


patricklafrance:
### Describe the bug

Dropdown item of type action have been added but they are not styled.

![image](https://user-images.githubusercontent.com/794579/75074997-e0363f80-54ca-11ea-8b92-1c864682639b.png)

Available in our SB at: `?path=/story/components-dropdown-standard--actions`

patricklafrance:
Not sure if it will be usefull but a CSS class `action` have been added on dropdown item that are actions

fraincs:
Useful. Need to discuss design with designers.

Issue #212, opened 2020-02-21
Dropdown item should support having icons in his content


patricklafrance:
### Describe the bug

SUI dropdown item support having custom content. We could leverage this to allow having icons in dropdown items content:

![image](https://user-images.githubusercontent.com/794579/75059774-a1dd5800-54ab-11ea-9648-0d32851f76fe.png)

Currently, it's working with the following code:

```
        <Dropdown
            selection
            options={[createMale(), createFemale(), { content: <span>With icon <CalendarIcon /></span>, value: "Frank the tank" }]}
            defaultOpen
        />
```

But the styling is off:

![image](https://user-images.githubusercontent.com/794579/75059824-bc173600-54ab-11ea-91aa-f943f26ce486.png)

Live example available in our SB: `?path=/story/chromatic-dropdown-item--icon-in-content`

fraincs:
This can also be done using floated content https://react.semantic-ui.com/modules/dropdown/#content-floated-content

Issue #211, opened 2020-02-21
Inline dropdown doesn't have any focus style


patricklafrance:
### Describe the bug

An inline dropdown should have focus style when a user tab to the dropdown with the keyboard. Currently nothing happen and the user doesn't have any clue what the focused item is.

fraincs:
openOnFocus could be the default behaviour for inline dropdowns, this is what semantic does

patricklafrance:
I don't think we want things popping while the user navigate with tabs.

All our other components only open when they are focused and the user press on the space bar.

fraincs:
Currently tabbing then pressing enter does nothing, this is something to think about.

patricklafrance:
@fraincs yes, you should create a separated issue for this.

Issue #210, opened 2020-02-21
Dropdown item should support description


patricklafrance:
### Is your feature request related to a problem?

Semantic UI dropdown item support having a description in an item:

![image](https://user-images.githubusercontent.com/794579/75059161-63936900-54aa-11ea-8336-f4699d03c08a.png)

```
    <Dropdown.Menu>
      <Dropdown.Header icon='tags' content='Filter by tag' />
      <Dropdown.Divider />
      <Dropdown.Item description='2 new' text='Important' />
      <Dropdown.Item description='10 new' text='Hopper' />
      <Dropdown.Item description='5 new' text='Discussion' />
    </Dropdown.Menu>
```

We currently support the feature but it's not style:

![image](https://user-images.githubusercontent.com/794579/75059226-802fa100-54aa-11ea-9f04-99abed6b8721.png)

There is a live example in our SB: `?path=/story/chromatic-dropdown-item--description`

PR #228, opened 2020-03-02
Feature/dropdown hoc update

fraincs:
## Issue
- Inline dropdown were missing a focus state https://github.com/gsoft-inc/sg-orbit/issues/211
- Loading Dropdowns were not showing a loading icon https://github.com/gsoft-inc/sg-orbit/issues/202
- Icons added to the content of a dropdown option were not centered https://github.com/gsoft-inc/sg-orbit/issues/212
- Clearable dropdowns icon was too big https://github.com/gsoft-inc/sg-orbit/issues/204
- A description to a dropdown was misaligned and unstyled https://github.com/gsoft-inc/sg-orbit/issues/210
- Icon before a dropdown option were not vertically aligned https://github.com/gsoft-inc/sg-orbit/issues/209
- Dropdown action was not styled https://github.com/gsoft-inc/sg-orbit/issues/214

## What I did

- A default focus state has been added to inline dropdown.
- Added the same loading animation as input's respecting sizes to loading dropdowns.
- Icons are now centered via flex when in a dropdown content.
- Changed the mask size of the clearable icon.
- Removed the float that was set by Semantic UI to descriptions and used flex instead.
- Icons are now centered via flex when at the start of a dropdown option.
- Added a text-decoration underline and a primary-500 color to dropdown actions. Only the text is clickable since it's not a choice but an external link .

## How to test

Chromatic tests have been added.

Yes it has been updated.


patricklafrance:
image

Link in dropdown item content should also be styled

patricklafrance:
This is weird? Would be solved easily by separating the input and the results

image

patricklafrance:
Personally, I still don't understand why we remove the input borders when the dropdown is focused.

image

packages/semantic-ui-theme/semantic/src/themes/sharegate/modules/dropdown.overrides
```diff
@@ -215,3 +292,99 @@
     height: 2rem;
     max-height: 2rem;
 }
+
+/* Content | With Icon */
+.ui.selection.dropdown .menu > .item {
```
patricklafrance: Are you sure you want to target only `.ui.selection` dropdown? Since items feature are supported for all kind of dropdown it seems limitative. For example, would it also work with an inline dropdown?

Same comments for the other selectors below

patricklafrance:
In the following story: https://deploy-preview-228--sg-orbit.netlify.com/?path=/docs/components-dropdown-inline--default-story#default

The vertical alignment of the arrow icon seems off. Also the icon is blurry.

patricklafrance:
Since the dropdown items size seems to vary base of the dropdown size, how do we handle the size of the icons provided in the content of a dropdown item?

Should it be managed by the toolkit?

fraincs:
> Personally, I still don't understand why we remove the input borders when the dropdown is focused.
> 
> image

This was a choice from the designer back when borders were evil.

fraincs:
> image
> 
> Link in dropdown item content should also be styled

I've added an underline on hover and defaulted to marine500

fraincs:
> Since the dropdown items size seems to vary base of the dropdown size, how do we handle the size of the icons provided in the content of a dropdown item?
> 
> Should it be managed by the toolkit?

This is already being taken care by the toolkit

Issue #245, opened 2020-03-06
Support multiple values for select components


patricklafrance:
### Is your feature request related to a problem?

SUI `<Dropdown />` component support selecting multiple values.

Our HOC, `<Select />` should also support having multiple values. It kind of work but the styling is off:

![image](https://user-images.githubusercontent.com/794579/76126197-7a6aad00-5fcc-11ea-8340-579dde1af0a2.png)

Issue #242, opened 2020-03-04
Inline dropdown should support error state


patricklafrance:
### Describe the bug

An inline dropdown doesn't support error state:

```
            <Dropdown
                placeholder="Gender"
                inline
                error
                options={GENDERS}
                className="mr5"
            />
```

![image](https://user-images.githubusercontent.com/794579/75925330-86bd0180-5e36-11ea-8f17-378c1b0f0022.png)

![image](https://user-images.githubusercontent.com/794579/75925589-efa47980-5e36-11ea-8e4f-bc523d6ec7b0.png)

fraincs:
I totally agree. No error state for dropdowns have been agreed upon with designers, this is something to do before going further.

patricklafrance:
Ok, but it has to be done ASAP since the standard select already support error. If we add a functionnality to the select it must work for every variants, otherwise we remove the functionnality.

fraincs:
the standard select supports error but nothing has been validated

patricklafrance:
got it, I don't mind removing it from the docs then until it's validated.

That being said, I believe all our inputs controls should come with support for error state.

Issue #241, opened 2020-03-04
Non selection dropdown should at least have basic support


patricklafrance:
### Describe the bug

Non `selection` dropdown should at least have basic support.

For the following selection dropdown: 

![image](https://user-images.githubusercontent.com/794579/75924775-8a03bd80-5e35-11ea-8cdd-1dcfea5e3adb.png)

We get the following without `selection`:

![image](https://user-images.githubusercontent.com/794579/75924823-9ee05100-5e35-11ea-907d-0cc9464f9403.png)

Issue #230, opened 2020-03-02
Dropdown arrow seems uncentered


fraincs:
This seem to be caused by the height of the arrow causing antialiasing when centered.

![image](https://user-images.githubusercontent.com/361632/75716740-cc929200-5c9d-11ea-9101-503d8c97b880.png)

![image](https://user-images.githubusercontent.com/361632/75716674-aa990f80-5c9d-11ea-8582-bdeb81b6631f.png)

### Describe the bug

The SVG arrow in dropdown is being antialased.

### Steps to reproduce

1. Inspect a dropdown

### Expected behavior

The arrow seem centered.

fraincs:
A reexport of the SVG has been made, with a round of decimal to 5. This seemed to do the trick, very hard to tell at such tiny heights.

PR #257, opened 2020-03-18
Milestones 5/dropdown

fraincs:
## Issues:
https://github.com/gsoft-inc/sg-orbit/issues/230
https://github.com/gsoft-inc/sg-orbit/issues/241
https://github.com/gsoft-inc/sg-orbit/issues/242
https://github.com/gsoft-inc/sg-orbit/issues/245

## What I did

- Exported the dropdown arrow svg in order to mitigate it's blurry appearance.
- Added support for non selection dropdown.
- Added support for errors in inline dropdowns.
- Added support for different font-size in inline dropdowns.
- Added support multiple values for select components.

## How to test

- Is this testable with Jest or Chromatic screenshots? **Chromatic** yes
- Does this need an update to the documentation? Could benefit from it.


packages/react-components/components/select/src/select.jsx
```diff
@@ -45,7 +50,7 @@ const renderAction = ({ content, key, className, ...rest }, index) => {
 };
 
 export function PureSelect(props) {
-    const { options, actions, inline, forwardedRef, ...rest } = props;
+    const { options, actions, inline, multiple, forwardedRef, ...rest } = props;
```
patricklafrance: You dont need to add `multiple` here since you don't use it, it would be provided to the select component through `...rest`.

packages/react-components/components/select/src/select.jsx
```diff
@@ -69,6 +74,7 @@ export function PureSelect(props) {
             options={renderOptions()}
             selection={!inline}
             inline={inline}
+            multiple={multiple}
```
patricklafrance: You dont need to add `multiple` here since you don't use it, it would be provided to the select component through `...rest`.

packages/react-components/components/select/src/select.jsx
```diff
@@ -21,6 +21,10 @@ const propTypes = {
      * A select can have a list of actions after his items.
      */
     actions: arrayOf(shape(ACTION_SHAPE)),
+    /**
+     * @ignore
+     */
+    multiple: bool,
```
patricklafrance: You dont need to add `multiple` here since you don't use it, it would be provided to the select component through `...rest`.

packages/react-components/components/select/src/select.jsx
```diff
@@ -32,7 +36,8 @@ const propTypes = {
 };
 
 const defaultProps = {
-    inline: false
+    inline: false,
+    multiple: false
```
patricklafrance: You dont need to add `multiple` here since you don't use it, it would be provided to the select component through `...rest`.

packages/semantic-ui-theme/semantic/src/definitions/modules/dropdown.less
```diff
@@ -563,10 +563,11 @@ select.ui.dropdown {
   position: absolute;
   z-index: 2;
 }
+
 .ui.search.dropdown > .text {
   cursor: text;
   position: relative;
-  left: @textCursorSpacing;
```
patricklafrance: if there is a variable, can't we instead update the value of the variable in sharegate theme?

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/image.overrides
```diff
@@ -1,3 +1,38 @@
 /*******************************
     Site Overrides
 *******************************/
+
```
patricklafrance: Does the size should be applied to ALL images instead of scoping to avatar?

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/label.overrides
```diff
@@ -283,3 +283,30 @@
 .ui.label.highlight {
     font-weight: 600;
 }
+
+/* CONTENT  | In Select */
```
patricklafrance: Not sure about this. I don't think a label should be aware of the fact that a multi-select component if consuming it?

I believe this code should be moved.

packages/semantic-ui-theme/semantic/src/themes/sharegate/modules/dropdown.overrides
```diff
@@ -1,397 +1,553 @@
...
+    border-radius: 0.25rem;
+    overflow-x: hidden;
+    overflow-y: auto;
 }
-.ui.dropdown.inline + .dropdown-icon > .icon {
-    width: 1em !important;
-    height: 1em !important;
-    fill: @marine200;
+
+/* LAYOUT | Dropdown */
```
patricklafrance: How come you have "LAYOUT | Dropdown" twice?

patricklafrance:
Looks good!

Let me finish something and work a little bit on the multi values select before merging please.

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/image.overrides
fraincs: To me an avatar should not have the same size as avatar which tend to be smaller by it's nature. Image size and avatar are something to discuss with designers.

packages/semantic-ui-theme/semantic/src/themes/sharegate/modules/dropdown.overrides
fraincs: derp

patricklafrance:
We should also support: https://react.semantic-ui.com/modules/dropdown/#types-multiple-search-selection

Repro available in SB: ?path=/story/chromatic-select-multiple-values--searchable

Issue #239, opened 2020-03-04
Should Orbit UI offers a button variant that is styled as a link?


patricklafrance:
Should Orbi UI offers a button variant that is styled as a link?

patricklafrance:
After a discussion with the designers, YES!

https://www.figma.com/file/VATs1Zu2vEXqIBicdwP3fI/Orbit---design-system?node-id=0%3A2523

PR #288, opened 2020-03-29
Added a link variant to the button component

patricklafrance:
Issue: https://github.com/gsoft-inc/sg-orbit/issues/239

## How to test

### DIFF
--- a/packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/button.overrides
+++ b/packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/button.overrides
@@ -8,7 +8,7 @@
 .ui.button,
 .ui.buttons .button {
     min-height: @scaleHotel;
-    display: flex;
+    display: inline-flex;
     border-radius: calc(@scaleHotel / 2);
     align-items: center;
     justify-content: center;
@@ -357,12 +357,6 @@
     background-color: @transparent;
 }
 
-/* SKIN | Basic | Secondary */
-.ui.basic.secondary.buttons .button,
-.ui.basic.secondary.button {
-    color: @marine500 !important;
-}
-
 /* SKIN | Ghost | Primary */
 .ui.ghost.primary.button,
 .ui.ghost.primary.buttons .button {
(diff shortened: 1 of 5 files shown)


packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/button.overrides
```diff
@@ -8,7 +8,7 @@
 .ui.button,
 .ui.buttons .button {
     min-height: @scaleHotel;
-    display: flex;
+    display: inline-flex;
```
patricklafrance: What do you think @fraincs ? With `display: flex` the button itself was not taking 100% of it's parent, but it's container was.

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/button.overrides
```diff
@@ -1020,6 +1019,34 @@
...
+    box-shadow: none !important;
+}
+
+/* SKIN | Link | Basic | STATE | Active */
+.ui.button.link.basic:focus,
+.ui.button.link.basic:hover,
+.ui.button.link.basic:active {
+    text-decoration: none;
+    color: var(--primary-600) !important;
+}
```
patricklafrance: What do you think @fraincs ?

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/button.overrides
fraincs: all good that's the use of inline-flex

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/button.overrides
```diff
@@ -1020,6 +1019,34 @@
...
+    background: none !important;
+    color: var(--primary-600);
+    font-weight: 400;
+}
+
+/* SKIN | Link | STATE | Active */
+.ui.button.link:focus,
+.ui.button.link:hover,
+.ui.button.link:active {
+    text-decoration: underline;
```
fraincs: In the best world we would not use      text-decoration: underline;  but a border bottom

Issue #286, opened 2020-03-27
Radio button having a label but no text doesn't render well when checked


patricklafrance:
### Describe the bug

![image](https://user-images.githubusercontent.com/794579/77804673-817a5d80-7056-11ea-8cd5-ff6b8f534f3f.png)

### Steps to reproduce

Repro in SB: ?path=/story/chromatic-radio-checked--no-text

Issue #285, opened 2020-03-27
Indeterminate checkbox doesn't render well for small and large size


patricklafrance:
### Describe the bug

![image](https://user-images.githubusercontent.com/794579/77802925-4e35cf80-7052-11ea-89f4-9caf7303d2c6.png)

### Steps to reproduce

Repro in SB:

- ?path=/story/chromatic-checkbox-indeterminate--text
- ?path=/story/chromatic-checkbox-indeterminate--no-text

Issue #271, opened 2020-03-24
A select item with an icon doesn't follow text / icon ratio


patricklafrance:
### Describe the bug

![image](https://user-images.githubusercontent.com/794579/77432169-16116100-6db4-11ea-999e-27c8bdba68e4.png)

A select item icon size adapt to the select size but the text doesn't. Therefore, it breaks our standard text /icon ratio.

Available options:

- Don't adapt the icon size to the select size
- Adapt the text size to the select size

patricklafrance:
After talking with design, we will adapt the text size to the select size

fraincs:
So the text size will be the same as the value size?

Issue #196, opened 2020-02-18
Input left padding is not validated by designers


fraincs:
The padding on the left of an input (when loading) is a necessity but never been discussed with designers. The behaviour inherits some Semantic UI defaults.

### Steps to reproduce

1. Visit https://sg-storybook.netlify.com/?path=/story/chromatic-input--sizes
2. Some paddings are not on our scale (e.g. 37px)

patricklafrance:
Also have a look at SearchInput & Dropdown

PR #291, opened 2020-03-30
Bundle/dropdowns

fraincs:
Issues:
https://github.com/gsoft-inc/sg-orbit/issues/196
https://github.com/gsoft-inc/sg-orbit/issues/271
https://github.com/gsoft-inc/sg-orbit/issues/285
https://github.com/gsoft-inc/sg-orbit/issues/286

## What I did

- Followed a common pattern for padding of elements inside an input/select/dropdown.
- Followed a common pattern for icon size inside an input/select/dropdown.
- Fixed Indeterminate checkbox sizing and position when not the default size.
- Fixed radio button selected dot when only a label is used as a label.

## How to test

- Is this testable with Jest or Chromatic screenshots? Yes


packages/react-components/src/checkbox/tests/chromatic/shared-stories.jsx
```diff
@@ -11,23 +11,23 @@ function Icons({ element, ...rest }) {
 
     return (
         <>
-            <div className="flex mb8">
+            <div className="flex items-start mb8">
```
patricklafrance: This is a detail, but can you use `ìtems-end` instead? I used `items-end` over `items-start` almost everywhere.

packages/react-components/src/input/src/input.jsx
```diff
@@ -96,6 +95,11 @@ function throwWhenMutuallyExclusivePropsAreProvided({ button, icon, iconPosition
 
 export function PureInput(props) {
     const { autofocus, autofocusDelay, className, fluid, icon, iconPosition, button, size, loading, disabled, children, forwardedRef, ...rest } = props;
+    const SIZES_TO_BUTTON = {
+        [SMALL]: "tiny",
+        [MEDIUM]: "small",
+        [LARGE]: "medium"
```
patricklafrance: Instead of hardcoding "tiny", "small", "medium" you should also used constant from `"../../shared`

packages/semantic-ui-theme/semantic/src/themes/sharegate/collections/form.overrides
```diff
@@ -99,18 +99,19 @@
 
 /* LAYOUT | Sizes */
 .ui.textarea {
-    padding: 1rem;
+    padding: 1rem 1.25rem;
```
patricklafrance: Should you use our CSS instead?

I am getting worried that if we update a foundation CSS variable value one day, all our apps will be updated but Orbit-UI toolkit will not since we don't use the CSS vars.

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/input.overrides
```diff
@@ -42,36 +45,58 @@
 
 /* Icons */
 .ui.icon.input.small > input {
-    padding-right: calc(var(--scale-foxtrot) + @iconPadding) !important;
+    padding-right: var(--scale-golf)!important;
```
patricklafrance: Ça fait du bien, c'était assez dégeux mon affaire haha

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/input.overrides
```diff
@@ -42,36 +45,58 @@
...
-    padding-right: calc(var(--scale-golf) + @iconPadding) !important;
+    padding-right: var(--scale-india) !important;
 }
 
 .ui.icon.input > svg.icon {
     top: 50%;
     transform: translateY(-50%);
     position: absolute;
-    right: var(--scale-bravo);
+    right: 10px;
```
patricklafrance: Make me wonder... shouldn't get these value from a calc based on our vars?

For example, if the padding is 40px which means `--scale-hotel` we could get the value from the following calc: `calc((var(--scale-hotel) / 2) / 2)`

packages/semantic-ui-theme/semantic/src/themes/sharegate/modules/dropdown.overrides
```diff
@@ -622,38 +665,54 @@
...
+.ui.loading.dropdown.large > i.icon::before,
+.ui.loading.dropdown.large > i.icon::after {
+    width: 1.5rem;
+    height: 1.5rem;
+}
+
 /*  STATE | Loading | SIZE | Small */
 .ui.loading.dropdown.small > i.icon {
-    height: 1.25rem;
+    height: 1rem;
```
patricklafrance: Could use `--scale-delta` instead of `1rem`

packages/semantic-ui-theme/semantic/src/themes/sharegate/collections/form.overrides
fraincs: Yes indeed

packages/semantic-ui-theme/semantic/src/themes/sharegate/elements/input.overrides
fraincs: dans ce cas c'est plus icon size / 2 et non /2 /2 mais je te suis

Issue #277, opened 2020-03-25
With the select component, a selected item with an avatar should display the avatar in the selected value.


patricklafrance:
### Describe the bug

With the select component, a selected item with an avatar should display the avatar in the selected value.

![image](https://user-images.githubusercontent.com/794579/77556075-65c45b00-6e8e-11ea-81b4-6dd3f04990b0.png)

https://www.figma.com/file/VATs1Zu2vEXqIBicdwP3fI/Orbit---design-system?node-id=0%3A1062

PR #300, opened 2020-04-03
Select option with avatar should display the avatar when selected

patricklafrance:
Issue: https://github.com/gsoft-inc/sg-orbit/issues/277


packages/react-components/src/select/src/item.jsx
```diff
@@ -15,9 +15,9 @@ const AVATAR_SHAPE = {
 };
 
 const SIZES_TO_AVATAR = {
-    [SMALL]: "tiny",
-    [MEDIUM]: "small",
-    [LARGE]: "small"
+    [SMALL]: MINI,
```
fraincs, line 18: Better

Issue #323, opened 2020-04-13
Inline select different icon size when used outsied a text block


Gcaya:
### Describe the bug

We want to use the inline select but not within a text block. When we do so, there is a difference in size between the icon presented to the left of the component and those inside the select list.

### Steps to reproduce

1. Render an inline select outside of a text block and with options containing icons
2. Click the select component and see the difference

### Expected behavior

I was expecting the icons both to the left and in the select list to be the same size.

### Actual behavior

Nothing extreme but you can clearly see the difference in size.

![image](https://user-images.githubusercontent.com/8711024/79150305-85b4b380-7d96-11ea-8660-ff9762ae1047.png)

### Deadline

If possible, this would be needed at the end of April.

PR #326, opened 2020-04-16
loading inline

fraincs:
Issue:

https://github.com/gsoft-inc/sg-orbit/issues/323

## How to test

### DIFF
--- a/packages/react-components/src/select/stories/inline.stories.mdx
+++ b/packages/react-components/src/select/stories/inline.stories.mdx
@@ -104,6 +104,23 @@ An inlined select can be embedded in a block.
     </Story>
 </Preview>
 
+### Loading
+
+An inlined select can show that it is currently loading data.
+
+<Preview>
+    <Story name="loading">
+        <div><Select
+            placeholder="select an astronaut"
+            inline
+            loading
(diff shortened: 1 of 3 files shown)


packages/react-components/src/select/stories/inline.stories.mdx
```diff
@@ -104,6 +104,23 @@ An inlined select can be embedded in a block.
     </Story>
 </Preview>
 
+### Loading
+
+An inlined select can show that it is currently loading data.
+
+<Preview>
+    <Story name="loading">
+        <div><Select
```
patricklafrance, line 113: You don't need the <div></div>

patricklafrance:
As you said, it looks weird because it's not centered for all font-size.

I am not sure what to think of this, we need it but at the same time it doesn't look very professional to release something that is not centered.

packages/react-components/src/select/stories/inline.stories.mdx
fraincs: the what

patricklafrance: the div

Issue #189, opened 2020-02-14
Focus states are the same as active and hover


fraincs:
This is problematic in situation where you click on a button that acts as a toggle, unselecting it does not change its look.
This is also problematic for accessibility reasons.

This applies to all components that can be clicked/focused. (e.g. input, buttons)

Explorations have been made with designers although nothing final.

![eHewCuCCVH](https://user-images.githubusercontent.com/361632/74544398-109f3c00-4f15-11ea-976c-efc686dac1b1.gif)

### Steps to reproduce

1. Hover a button
2. Click it
3. There is no difference between active and hover

### Expected behavior

An active element has a different style.

PR #338, opened 2020-04-30
Feature/focus states

fraincs:
Issue:
[189](https://github.com/gsoft-inc/sg-orbit/issues/189)


patricklafrance:
An active button seems to look like an focused button:

image

I am not sure this is what we want. However, if this is what we want, this seems inconsistant with a button group an active button group:

image

patricklafrance:
Date picker calendar clear button is missing a part of the focus indicator:

image

Also true for the apply button:

image

patricklafrance:
Seems weird that an inline date picker:

image

And a dropdown menu:

image

Are different, seems like both are some kind of underline texte with an arrow.

patricklafrance:
A focused Select doesn't seems to have a green shadows like the other components:

image

I would expect something like a radio:

image

patricklafrance:
I would expect the inline select to be more like the dropdown menu when focused.

We seems to have 3 different styles for basically the same thing:

Inline select:
image

Dropdown menu:
image

Inline date picker:
image

patricklafrance:
I believe a multiple values select should support tabbing to the selected values label.

Might be out of scope for this PR but please create a new issue if this is out of scope:

image

patricklafrance:
Transparent select focused style seems different than inline select style, is it what we want?

Transparent select:
image

Inline select:
image

fraincs:
> An active button seems to look like a focused button:
> 
> image
> 
> I am not sure this is what we want. However, if this is what we want, this seems inconsistant with an active button group:
> 
> image

actually there is no way to do a proper active state on dark buttons, except maybe using a lighter color, a black border would not show.

fraincs:
> > An active button seems to look like a focused button:
> > image
> > I am not sure this is what we want. However, if this is what we want, this seems inconsistant with an active button group:
> > image
> 
> actually there is no way to do a proper active state on dark buttons, except maybe using a lighter color, a black border would not show.

the other issue is that it's impossible to follow the same logic on the group button due to the way they are engineered.

fraincs:
I removed the outline on active state, as for the group button the logic is that on active we darken the color. Since the color is white, we darken the border color. Makes sense?

patricklafrance:
Do we want to have a different style for an hovered item and the currently "keyboard" selected item?

image

patricklafrance:
The avatar size seems quite small, how did you choose the size?

image

fraincs:
> The avatar size seems quite small, how did you choose the size?
> 
> image

The maximum that can fit in. Since we use the same size logic as dropdown menu.

fraincs:
> Do we want to have a different style for an hovered item and the currently "keyboard" selected item?
> 
> image

Nothing has been set in stone with designers. I can talk to them ASAP,

PR #360, opened 2020-05-20
Hotfix/may 19

fraincs:
Issues:
- When navigating via keyboard the Selected item was changing font-weight, this was not optimal.
- An upward Select and Dropdown was showing a downward arrow.

## What I did
- The selected / hover style has been reworked in the style of react-select.
- The Select / Dropdown arrow is now rotated when the prop upward is true.

## How to test
?path=/story/components-select-standard--default-story
?path=/story/components-select-standard--upward


patricklafrance:
Are we sure about primary 900 on black text? Also how come text is still bold?

image

patricklafrance:
This one looks nice, but I am not 100% sure about the 500  one. Could we instead use for example 50 and 300 instead of 50 and 500?

image

fraincs:
> This one looks nice, but I am not 100% sure about the 500 one. Could we instead use for example 50 and 300 instead of 50 and 500?
> 
> image

image
I took inspiration from here.

Issue #364, opened 2020-05-22
All texts should default to marine 500


fraincs:
All texts should default to marine 500, some texts do not have a default color and default to pure black.

PR #381, opened 2020-06-18
Hotfix/12

fraincs:
Issue:
- Empty Select was giving it's parent an height to high.
- [364](https://github.com/gsoft-inc/sg-orbit/issues/364)

## What I did

- Added a fallback color for all ```ui and o-ui``` components.
- Set marine 500 to be the default for all components using ui and o-ui.

## How to test
/?path=/docs/components-select-standard--default-story


packages/semantic-ui-theme/semantic/src/themes/sharegate/globals/site.overrides
```diff
@@ -1,3 +1,7 @@
 /*******************************
     Site Overrides
 *******************************/
+.ui,
+.o-ui {
+    color: var(--marine-500);
+}
```
patricklafrance, line 7: I'm fine if we keep it like this, but I want to point up the fact that we are adding this to SUI theme and for `o-ui` aren't SUI related.

Also we will at some point not use the SUI theme anymore.

fraincs: This is a valid point, where would you put it?

patricklafrance: I don't have the answer yet.

PR #388, opened 2020-07-06
Revamp tag

fraincs:
Issue:
Lozenge / Tag / Badge were missing styles.

## What I did

CSS for Lozenge / Tag / Badge


packages/react-components/src/badge/src/Badge.css
```diff
@@ -96,6 +110,15 @@
     opacity: 0.4;
 }
 
+/* ICON */
+.o-ui.badge.icon {
+    display: inline-flex;
```
patricklafrance, line 115: Why would you need to alter an icon display to flex? It can't have any content.

packages/react-components/src/badge/src/Badge.jsx
```diff
@@ -51,7 +51,7 @@ export function InnerBadge(props) {
             )}
             ref={forwardedRef}
         >
-            {content}
+            <span>{content}</span>
```
patricklafrance, line 54: Yeah, I am still again this.

I think when possible we should avoir extra element.

I have a hard time buying it's required to correctly center the text since in the previous version of Orbit UI we we're able to do it without the extra SPAN and this is same use case except the root element is a DIV instead of a BUTTON.

https://sg-storybook.netlify.app/?path=/docs/components-button--default-story

packages/react-components/src/lozenge/src/Lozenge.css
```diff
@@ -2,4 +2,51 @@
...
+.o-ui.lozenge.with-right-icon {
+    padding-right: calc(var(--scale-bravo) / 2);
+}
+
+/* WITH ICON LEFT | ICON */
+.o-ui.lozenge.with-left-icon .icon {
+    margin-right: 0.125rem;
+}
+
+.o-ui.lozenge.with-right-icon .icon {
```
patricklafrance, line 45: I guess you should also add a comment here for the right icon

packages/tachyons/src/core/_outlines.css
```diff
@@ -1,6 +1,7 @@
-.outline {
+/* Clash with outline in orbit ui */
```
patricklafrance, line 1: What do we do with this?

If I remember correctly this is a class that we provide for the apps.

packages/react-components/src/badge/src/Badge.jsx
fraincs: It works there since we don't have a badge smaller that 16px. It fails at 14px since the content is 15px.

fraincs: Looking at Atlassian they have the same issue but avoid it by not having a badge smaller than 16px.

packages/tachyons/src/core/_outlines.css
fraincs: Yes but it clashes with .ui....outline, I am pretty sure it is not used in Apps. I'd rather have them set it on their side if needed. Thoughts?

packages/react-components/src/badge/src/Badge.css
fraincs: We needed it when I tried to go overboard with a background and all. Not needed anymore.

packages/tachyons/src/core/_outlines.css
patricklafrance: Would be safest way to handle this until we know more about it, so sure.

packages/react-components/src/badge/src/Badge.jsx
patricklafrance: Do we need a badge having 14px? What is the use case?

packages/tachyons/src/core/_outlines.css
fraincs: When in a Tag but I guess 16px wouldn't be bad either

packages/react-components/src/badge/src/Badge.jsx
fraincs: When in a Tag but I guess 16px wouldn't be bad either

packages/tachyons/src/core/_outlines.css
fraincs: I'll remove it and note in release notes.

PR #391, opened 2020-07-13
Allow bool and number as valid proptype for SelectItem value

AlexandreArpin:
Issue:
Semantic UI's Dropdown.Item allow for bool, number and string as valid values.

When providing a number as the value for a SelectItem in Orbit, you get a PropType warning.

The workaround right now is to cast to string and parse around the SelectItem but ideally, Orbit's Select would support for bool and number values as well as string.

## What I did
Modify the PropType of SelectItem to allow for bool and number to be passed.

We know Semantic UI DropdownItem support all 3 proptype values (https://react.semantic-ui.com/modules/dropdown/), so it should simply work.

## How to test
I _tried_ adding a story for my bug/usecase, I'm not sure if we want to include a story for this particular usecase.

### DIFF
--- a/packages/react-components/src/select/src/SelectItem.jsx
+++ b/packages/react-components/src/select/src/SelectItem.jsx
@@ -1,6 +1,6 @@
 import { SelectContext } from "./SelectContext";
 import { Dropdown as SemanticDropdown } from "semantic-ui-react";
-import { arrayOf, element, oneOf, oneOfType, shape, string } from "prop-types";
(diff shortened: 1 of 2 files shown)


patricklafrance:
Thanks for the PR.

I will move the story to our regression test (`SelectItem.chroma.jsx`) later on.

AlexandreArpin:
:| 

Le name de ma story a pas été changer, ça crash netlify

Faudrait patcher `<Story name="disabled">`  pour `<Story name="number">`

PR #392, opened 2020-07-20
Fix multiple stories having the same name

AlexandreArpin:
Introduced in #391 

Just want to make sure master is stable!


patricklafrance:
Thanks for the fix!

PR #417, opened 2020-09-24
Outline error fix

fraincs:
Issue:
Outline in error states were not looking good.

## What I did
Added an alert focus error color.


patricklafrance:
I would prefer if you use the term `focus-ring` instead of `outline` since this is the generally accepted term to describe this.

patricklafrance (approved):

PR #423, opened 2020-09-30
30/09/2020

fraincs:
Issues:

- focus states especially in inputs were not inner focus
- direct child of links were not inheriting the parent border radius
- icon link were not the same height of its children

## What I did

- added a concept of small and medium inset focus ring
- added a concept of `--o-ui-focus-ring-color` to easily override the ring color on a component basis
- set icon display mode to block in order to stop line heights changing their height

### DIFF
--- a/packages/react-components/src/index.css
+++ b/packages/react-components/src/index.css
@@ -1,9 +1,15 @@
 :root {
-    --o-ui-focus-ring: 0 0 0 2px var(--white), 0 0 0 4px var(--primary-500);
     --o-ui-disabled-opacity: 0.4;
     --o-ui-pill-border-radius: 999px;
     --o-ui-rounded-border-radius: var(--scale-alpha);
     --o-ui-circular-border-radius: 100%;
+}
+
+[class^="o-ui-"] {
+    --o-ui-focus-ring-inset-sm: 0 0 0 1px var(--o-ui-focus-ring-color);
+    --o-ui-focus-ring-inset-md: 0 0 0 2px var(--o-ui-focus-ring-color);
+    --o-ui-focus-ring: 0 0 0 2px var(--white), 0 0 0 4px var(--o-ui-focus-ring-color);
+    --o-ui-focus-ring-color: var(--primary-500);
 }
 
 /* ROTATE */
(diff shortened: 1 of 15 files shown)


patricklafrance:
The image in the card should still have rounded corners.

image

patricklafrance:
The chromatic tests associated to Link should also be updated to reflect the new inherit border-radius feature. I belive these tests still have a shape on `<Image>` components.

packages/react-components/src/index.css
```diff
@@ -1,11 +1,17 @@
...
-    --o-ui-focus-ring: 0 0 0 2px var(--white), 0 0 0 4px var(--primary-500);
     --o-ui-disabled-opacity: 0.4;
     --o-ui-pill-border-radius: 999px;
     --o-ui-rounded-border-radius: var(--scale-alpha);
     --o-ui-circular-border-radius: 100%;
 }
 
+[class^="o-ui-"] {
+    --o-ui-focus-ring-inset-sm: 0 0 0 1px var(--o-ui-focus-ring-color);
+    --o-ui-focus-ring-inset-md: 0 0 0 2px var(--o-ui-focus-ring-color);
```
patricklafrance, line 10: no `lg` support?

packages/react-components/src/index.css
```diff
@@ -1,11 +1,17 @@
 :root {
-    --o-ui-focus-ring: 0 0 0 2px var(--white), 0 0 0 4px var(--primary-500);
     --o-ui-disabled-opacity: 0.4;
     --o-ui-pill-border-radius: 999px;
     --o-ui-rounded-border-radius: var(--scale-alpha);
     --o-ui-circular-border-radius: 100%;
 }
 
+[class^="o-ui-"] {
```
patricklafrance, line 8: I guess you added this for specifity? If so, shouldn't we add all the vars here?

packages/react-components/src/index.css
fraincs: we do not need one, a 8px radii is way too much

packages/react-components/src/index.css
fraincs: for specificity as well to be able to use custom properties in custom properties. All of our component specific custom properties could be indeed prefixed.

packages/react-components/src/index.css
fraincs: actually it would be 4 lets do it

packages/react-components/src/index.css
patricklafrance: yes they should be!

patricklafrance (approved):

PR #427, opened 2020-10-09
Bouton rewrite

fraincs:
Issue:
- Button CSS was repetitive.
- Outline were not working on any background other than white

## What I did
- Removed repetitive CSS declarations
- Used another technique to display an outline on focus.


packages/react-components/src/button/src/Button.css
```diff
@@ -18,12 +18,14 @@
     cursor: pointer;
     outline: none;
     font-weight: 500;
+    justify-content: center;
```
patricklafrance, line 21: Doesn't make sense to me to set `justify-content` when the element display is not `flex` or `grid` ?

packages/react-components/src/index.css
```diff
@@ -1,16 +1,13 @@
-:root {
-    --o-ui-disabled-opacity: 0.4;
-    --o-ui-pill-border-radius: 999px;
-    --o-ui-rounded-border-radius: var(--scale-alpha);
-    --o-ui-circular-border-radius: 100%;
-}
-
 [class^="o-ui-"] {
     --o-ui-focus-ring-inset-sm: 0 0 0 1px var(--o-ui-focus-ring-color);
```
patricklafrance, line 2: Are these var still used since you updated how the focus ring is done?

packages/react-components/src/checkbox/src/Checkbox.jsx
```diff
@@ -152,7 +151,7 @@ export function InnerCheckbox(props) {
             {...rest}
             {...wrapperProps}
         >
-            <VisuallyHidden {...inputProps} />
+            <input {...inputProps} />
```
patricklafrance, line 154: Why? I would really prefer if you don't do this.

packages/react-components/src/button/src/Button.css
```diff
@@ -33,25 +35,71 @@
...
+}
+
+/* STATES | FOCUS | FOCUS RING */
+.o-ui-button:after {
+    opacity: 0;
+    content: "";
+    top: -2px;
+    left: -2px;
+    right: -2px;
+    bottom: -2px;
```
patricklafrance, line 52: How come our focus ring is now that complex? Used to be a simple box-shadow.

These focus ring position values should be extracted to vars in `index.css`

packages/react-components/src/button/src/Button.css
```diff
@@ -429,11 +364,14 @@ a.o-ui-button {
...
-.o-ui-button-ghost:focus,
-.o-ui-button-ghost.o-ui-button-focus {
-    box-shadow: var(--o-ui-focus-ring-inset-md);
-    z-index: 2;
-    outline: none;
+.o-ui-button-ghost:focus:after,
+.o-ui-button-ghost.o-ui-button-focus:after {
+    opacity: 1;
+    /* inner */
+    top: 1px;
```
patricklafrance, line 371: These position values should also be extracted to `index.css`

packages/react-components/src/checkbox/src/Checkbox.jsx
patricklafrance: React should throw an exception stating it doesn't know to render the prop `as`.  I am not sure why I don't see this exception in SB preview

packages/react-components/src/button/src/Button.css
```diff
@@ -33,25 +35,71 @@
...
+    content: "";
+    top: -2px;
+    left: -2px;
+    right: -2px;
+    bottom: -2px;
+    display: block;
+    position: absolute;
+    box-shadow: 0 0 0 2px var(--o-ui-focus-ring-color);
+    border-radius: var(--o-ui-pill-border-radius);
+    transition: opacity .2s ease-in;
```
patricklafrance, line 57: Should also be a CSS var?

packages/react-components/src/checkbox/src/Checkbox.css
```diff
@@ -240,13 +261,24 @@
...
 /* SIZING | LARGE */
 /* SIZING | LARGE | LABEL */
 .o-ui-checkbox-lg .o-ui-checkbox-box:before {
     width: var(--scale-foxtrot);
     height: var(--scale-foxtrot);
 }
 
+.o-ui-checkbox-lg input {
+    width: 32px;
+    height: 32px;
```
patricklafrance, line 279: Should use `--scale-golf`

packages/react-components/src/button/src/Button.css
fraincs: This was meant to be as generic as possible, I agree that I'd rather put it on `.o-ui-button:not(.o-ui-icon-button)` which is where we set buttons to a display of `inline-grid`

packages/react-components/src/index.css
fraincs: `--o-ui-rounded-border-radius` is still used for inner focus-rings (the ones that are not offset)

packages/react-components/src/checkbox/src/Checkbox.jsx
fraincs: > Why? I would really prefer if you don't do this and keep VisuallyHidden.
The only way to have an offset on a checkbox since both pseudos are already in use for the checkmark and the border.

packages/react-components/src/button/src/Button.css
fraincs: Since box-shadow does not supports an offset, on any background other than white you would see a 2px solid white border.

packages/react-components/src/checkbox/src/Checkbox.css
fraincs: the logic is 2x the border width (2 sides to a box) + 2 times the offset

packages/react-components/src/checkbox/src/Checkbox.css
```diff
@@ -1,4 +1,6 @@
 .o-ui-checkbox {
+    --o-ui-focus-ring-stroke-width: 2px;
+    --o-ui-indeterminate-stroke-width-md: 5px;
```
patricklafrance, line 3: Why `-md` suffix for the for the indeterminate but not for `--o-ui-focus-ring-stroke-width`. Don't care if we add it or remove it but if it should be consistant if there isn't any good reason not to be.

packages/react-components/src/checkbox/src/Checkbox.css
```diff
@@ -7,7 +9,7 @@
     line-height: 1;
     max-width: 100%;
     width: max-content;
-    height: max-content;
```
patricklafrance, line 10: Why did you remove the max-content? Ideally we would keep this for every components

packages/react-components/src/checkbox/stories/Checkbox.stories.mdx
```diff
@@ -84,7 +84,7 @@ A checkbox can be indeterminate.
 
 <Preview>
     <Story name="indeterminate">
-        <Checkbox defaultIndeterminate>Milky Way</Checkbox>
+        <Checkbox size="lg" defaultIndeterminate>Milky Way</Checkbox>
```
patricklafrance, line 87: Derp here with `size="lg"`

packages/react-components/src/index.css
```diff
@@ -1,16 +1,16 @@
...
     --o-ui-focus-ring-inset-md: 0 0 0 2px var(--o-ui-focus-ring-color);
     --o-ui-focus-ring-inset-lg: 0 0 0 4px var(--o-ui-focus-ring-color);
     --o-ui-focus-ring: 0 0 0 2px var(--white), 0 0 0 4px var(--o-ui-focus-ring-color);
+    --o-ui-focus-ring-offset: 2px;
+    --o-ui-focus-ring-inner-offset: 1px;
```
patricklafrance, line 7: `inner-offset` doesn't make much sense to me.

packages/react-components/src/index.css
```diff
@@ -1,16 +1,16 @@
...
-    --o-ui-disabled-opacity: 0.4;
-    --o-ui-pill-border-radius: 999px;
-    --o-ui-rounded-border-radius: var(--scale-alpha);
-    --o-ui-circular-border-radius: 100%;
-}
-
 [class^="o-ui-"] {
     --o-ui-focus-ring-inset-sm: 0 0 0 1px var(--o-ui-focus-ring-color);
     --o-ui-focus-ring-inset-md: 0 0 0 2px var(--o-ui-focus-ring-color);
     --o-ui-focus-ring-inset-lg: 0 0 0 4px var(--o-ui-focus-ring-color);
```
patricklafrance, line 4: We do have an `outset` and an `inset` focus-ring. Therefore I would expect to have var names that contains `inset` & `ouset`. We do have vars for `inset` but I don't see any for `outset`.

packages/react-components/src/input/src/Input.css
```diff
@@ -188,7 +187,7 @@
 }
 
 /* BUTTON */
-.o-ui-input-button {
+.o-ui-input-has-button .o-ui-input-button {
```
patricklafrance, line 190: Not sure why would need to add `has-button` here? The class `o-ui-input-button` would only be available there IS a button. There IMO it's redundant and useless.

packages/react-components/src/index.css
fraincs: this is opposed to the outer offset, how would you name this variable since it's use is only to "insert" the focus ring in the element.

packages/react-components/src/index.css
fraincs: That is because the outset focus ring is not using a box shadow

patricklafrance: Well as far as I understand all this we do have a var like `--o-ui-focus-ring-offset: 2px;` that exist to be used as part of an `outset` focus-ring?

`--o-ui-focus-ring-outset-offset: 2px`

packages/react-components/src/index.css
patricklafrance: Make sense, but how was I supposed to understand this by looking at these var names?

`--o-ui-focus-ring-offset: 2px;` should somehow include outer then

packages/react-components/src/index.css
patricklafrance: After reviewing other comments it should be something like:

```
--o-ui-focus-ring-outset-outer-offset: 2px;
--o-ui-focus-ring-outset-inner-offset: 1px;
```

packages/react-components/src/index.css
fraincs: totall;

PR #442, opened 2021-01-14
Theme/darkmode

fraincs:
Issue:

Dark Theme was not implemented.

## What I did

- Added indirections to colours
- Added a concept of `color inherit` to the Text and Heading Component in order to have a default to marine900 while being able to have a text that adapts it's colour, see Alert Component
- Many other bug fixes.


packages/foundation/src/atoms/core/colors/brand.css
```diff
@@ -1,35 +1,38 @@
-:root {
+[class^="o-ui-"] {
```
patricklafrance, line 1: I don't think something in foundation should be scope with "o-ui". I do see our foundation as our current implementation or our future design tokens so it should be developed under the same principles.

packages/foundation/src/atoms/core/foundations/index.css
```diff
@@ -0,0 +1,11 @@
+.o-ui-background {
```
patricklafrance, line 1: Shouldn't be any "o-ui" in foundation.

I don't think either there should be CSS classes in foundation, only vars

packages/foundation/src/atoms/core/index.css
```diff
@@ -1,4 +1,5 @@
 @import "./colors/index.css";
+@import "./foundations/index.css";
```
patricklafrance, line 2: A nested "foundations" folder under a "foundation" project seems like a bad name.

packages/foundation/tests/shadows.chroma.jsx
```diff
@@ -10,7 +10,7 @@ stories()
     .add("default",
          () =>
              <Inline gap={12}>
-                 <div className="w13 h13 shadow-100 mb8" />
-                 <div className="w13 h13 shadow-200" />
+                 <div className="w13 h13 shadow-100 bg-white mb8" />
```
patricklafrance, line 13: Why? If you need to set a background-color shouldn't it use an indirection?

packages/react-components/src/accordion/src/Accordion.css
```diff
@@ -103,6 +104,7 @@
     right: var(--scale-bravo);
     bottom: calc(var(--scale-bravo) - 2px);
     border: 2px transparent solid;
+    transition: var(--o-ui-focus-ring-transition);
```
patricklafrance, line 107: Do we still need this since we reject the idea of delaying the focus ring?

packages/react-components/src/accordion/src/Accordion.css
```diff
@@ -117,5 +119,5 @@
 
 .o-ui-accordion-trigger:focus::before,
 .o-ui-accordion-trigger-focus::before {
-    border-color: var(--primary-500);
+    border-color: var(--brand-border);
```
patricklafrance, line 122: why "brand" instead of "primary" ?

packages/react-components/src/alert/src/Alert.css
```diff
@@ -67,20 +66,42 @@
...
+    color: var(--critical-900);
 }
 
 .o-ui-alert-critical .o-ui-alert-icon {
-    fill: var(--alert-900);
+    fill: var(--critical-900);
+}
+
+/* DARK MODE */
+/* @todo  Note add an indirection here - we should not be aware that we are using Sunray and al. */
```
patricklafrance, line 88: Seems like you need to do something here.

packages/react-components/src/button/src/Button.css
```diff
@@ -34,6 +34,7 @@
     display: block;
     position: absolute;
     box-shadow: 0 0 0 2px var(--o-ui-focus-ring-color);
+    /* box-shadow: 0 0 0 2px currentColor; */
```
patricklafrance, line 37: To remove

packages/react-components/src/button/src/Button.css
```diff
@@ -56,22 +57,22 @@ a.o-ui-button {
 
 /* SOLID */
 .o-ui-button-solid {
-    color: var(--white);
-    background-color: var(--marine-500);
+    color: var(--static-white);
```
patricklafrance, line 60: static? It honnestly start to get confusing.

packages/react-components/src/button/src/Button.css
```diff
@@ -56,22 +57,22 @@ a.o-ui-button {
...
+    background-color: var(--foreground-color-5);
```
patricklafrance, line 61: I am not sure as a developper how I will understand this code 1-2-3-4-5 it's starting to be a lot.

I could manage with 1 or 2 variations but 5 is too many.

packages/react-components/src/button/src/Button.css
```diff
@@ -83,82 +84,95 @@ a.o-ui-button {
...
 .o-ui-button-solid:active.o-ui-button-disabled,
 .o-ui-button-solid.o-ui-button-active:disabled {
-    background-color: var(--marine-500);
+    background-color: var(--foreground-color-5);
 }
 
 /* SOLID | PRIMARY */
 .o-ui-button-solid.o-ui-button-primary {
-    background-color: var(--primary-500);
+    background-color: var(--foreground-color-brand);
```
patricklafrance, line 100: Again I would have expected `--foreground-color-primary`. Maybe it's me who doesn't understand the system well.

packages/react-components/src/button/src/Button.css
```diff
@@ -169,8 +183,14 @@ a.o-ui-button {
 /* OUTLINE | HOVER */
 .o-ui-button-outline:hover,
 .o-ui-button-outline.o-ui-button-hover {
-    box-shadow: 0 0 0 1px var(--marine-700) inset;
-    color: var(--marine-700);
+    box-shadow: 0 0 0 1px var(--quaternary-border-hover) inset;
```
patricklafrance, line 186: What the hell is "quaternary" lol ?

patricklafrance: I feel something like "shadow-border-..." might be easier to understand

packages/react-components/src/button/src/Button.css
```diff
@@ -182,162 +202,163 @@ a.o-ui-button {
...
+    box-shadow: 0 0 0 1px var(--quaternary-border) inset;
+    color: var(--secondary-text);
+    background-color: var(--foreground-color-1);
 }
 
 /* OUTLINE | PRIMARY */
 .o-ui-button-outline.o-ui-button-primary {
-    box-shadow: 0 0 0 1px var(--primary-500) inset;
-    color: var(--primary-500);
+    box-shadow: 0 0 0 1px var(--brand-border) inset;
```
patricklafrance, line 222: I guess the weird thing is that to me "primary" always meant the brand color. Why does it change now?

packages/react-components/src/button/src/Button.css
```diff
@@ -182,162 +202,163 @@ a.o-ui-button {
...
+    color: var(--static-white);
+    background-color: var(--foreground-color-brand);
 }
 
 /* OUTLINE | SECONDARY */
 .o-ui-button-outline.o-ui-button-secondary {
-    box-shadow: 0 0 0 1px var(--cloud-200) inset;
-    color: var(--cloud-500);
+    box-shadow: 0 0 0 1px var(--tertiary-border) inset;
```
patricklafrance, line 243: I don't think "tertiary", "quaternary" is the way to go. It reminds me too much of spacing chart "alpha", "bravo" mess. Might be even worst since it is prone to typing issue.

packages/react-components/src/button/src/Button.css
```diff
@@ -182,162 +202,163 @@ a.o-ui-button {
...
 }
 
 /* OUTLINE | SECONDARY | ACTIVE */
 .o-ui-button-outline.o-ui-button-secondary:active,
 .o-ui-button-outline.o-ui-button-secondary.o-ui-button-active {
-    box-shadow: 0 0 0 1px var(--cloud-700) inset;
-    color: var(--white);
+    box-shadow: 0 0 0 1px var(--primary-border-active) inset;
+    background-color: var(--foreground-color-2-active);
+    color: var(--static-white);
```
patricklafrance, line 258: Why not "--white" ? Why the "static" prefix?

packages/react-components/src/button/src/Button.css
```diff
@@ -389,20 +410,24 @@ a.o-ui-button {
 
 /* SOLID | LOADING | PRIMARY */
 .o-ui-button-loading.o-ui-button-solid.o-ui-button-primary:before {
-    border: 2px solid var(--primary-800);
-    border-top-color: var(--white);
+    border: 2px solid rgba(256, 256, 256, 0.5);
```
patricklafrance, line 413: why no var?

packages/react-components/src/button/src/Button.css
```diff
@@ -389,20 +410,24 @@ a.o-ui-button {
...
+    border: 2px solid var(--cloud-500);
+    border-top-color: var(--primary-text);
 }
 
-/* SOLID | LOADING | DANGER */
 .o-ui-button-loading.o-ui-button-solid.o-ui-button-danger:before {
-    border: 2px solid var(--alert-300);
-    border-top-color: var(--white);
+    --opacity-alert-500: 0.5;
+    border: 2px solid rgba(256, 256, 256, 0.5);
```
patricklafrance, line 429: why no var?

packages/react-components/src/button/src/Button.jsx
```diff
@@ -125,11 +125,11 @@ export function InnerButton(props) {
         },
         icon: {
             size: condensed ? size : embeddedIconSize(size),
-            className: "o-ui-button-left-icon"
+            className: "o-ui-button-icon o-ui-button-left-icon"
```
patricklafrance, line 128: This modification is wrong.

This is the Button component not the IconButton component. The original code is correct it should be:

```
icon: {
            size: condensed ? size : embeddedIconSize(size),
            className: "o-ui-button-left-icon"
}
```

packages/react-components/src/button/src/Button.jsx
```diff
@@ -125,11 +125,11 @@ export function InnerButton(props) {
...
         },
         text: {
             size: condensed ? condensedTextSize(size) : size,
-            className: "o-ui-button-text",
+            className: "o-ui-button-text o-ui-text-color-inherit",
```
patricklafrance, line 132: I am not sure why we would need this here?

I don't have a problem with text-color-inherit but I am not sure it should be applied de-factor. I see it more for special cases like an Alert

packages/react-components/src/button/src/IconButton.css
```diff
@@ -62,6 +62,12 @@
 
 /* INHERIT */
 .o-ui-button-inherit.o-ui-button-solid .o-ui-button-icon {
-    fill: var(--white);
+    fill: var(--static-white);
     opacity: .9;
 }
+
+.o-ui-icon-button .o-ui-button-icon,
```
patricklafrance, line 69: Not sure about this selector? It seem related to this?

image

packages/react-components/src/button/src/IconButton.css
```diff
@@ -62,6 +62,12 @@
...
+.o-ui-button-icon .o-ui-icon {
```
patricklafrance, line 70: Every selector in this file should start with `.o-ui-icon-button`

packages/react-components/src/checkbox/src/Checkbox.css
```diff
@@ -46,38 +51,34 @@
     left: 50%;
     width: var(--scale-echo);
     height: var(--scale-echo);
+    mask: url("~@orbit-ui/icons/dist/icon-check-24.svg") center center no-repeat;
```
patricklafrance, line 54: I guess this is a new strategy that I don't understand yet but it seems weird to render the check icon when the checkbox is not yet checked.

packages/react-components/src/icons/src/Icon.jsx
```diff
@@ -15,18 +15,28 @@ const propTypes = {
...
     /**
      * Default slot override.
      */
     slot: string
 };
 
+export function getIconClass(color) {
+    const colorClass = color ? `color-${color}`: "";
+
+    return colorClass;
```
patricklafrance, line 31: That type of code should be directly in the `cssModule`.  I understand you duplicate it twice but it is super simple code.

packages/react-components/src/input/src/Input.css
```diff
@@ -217,3 +216,11 @@
 .o-ui-input-loading.o-ui-input input {
     padding-right: var(--o-ui-padding-loading);
 }
+
+/* DARK MODE */
+.o-ui-dark-color-scheme .o-ui-input input::selection,
+.o-ui-dark-color-scheme .o-ui-input textarea::selection {
+    /* add indirection - I guess selection color makes sense here or in fact primary colors should not be inverted only their inderections */
```
patricklafrance, line 223: Seems like a TODO

packages/react-components/src/input/src/NumberInput.css
```diff
@@ -23,13 +23,14 @@
...
-    width: 150px;
+    /* 150px @ 16px */
+    width: 9.365rem;
 }
 
 /* SPINNER */
 .o-ui-number-input-spinner {
     display: flex;
-    visibility: hidden;
+    /* visibility: hidden; */
```
patricklafrance, line 33: Are you sure?

packages/react-components/src/layout/tests/chromatic/Stack.chroma.jsx
```diff
@@ -10,57 +11,57 @@ function stories(segment) {
 stories()
     .add("default", () =>
         <Stack>
-            <div className="bg-primary-500">Alpha</div>
-            <div className="bg-primary-500">Bravo</div>
-            <div className="bg-primary-500">Charlie</div>
+            <div className="bg-primary-500"><Text>Alpha</Text></div>
```
patricklafrance, line 14: We shouldn't need the Text component here. I mean the devs who use our components should be able to render text without using the <Text /> component.

patricklafrance: Basically the Text component exist for usecases where the componens render multiple children for example :

```
                    <Button>
                        <SignoutIcon />
                        <Text>Button</Text>
                    </Button>
```

Because we use the Text component in this usecase, the Button component can internally differentiate the text and the icon and augment them accordingly.

packages/react-components/src/layout/tests/chromatic/Stack.chroma.jsx
```diff
@@ -10,57 +11,57 @@ function stories(segment) {
...
+            <div className="bg-primary-500"><Text>Bravo</Text></div>
+            <div className="bg-primary-500"><Text>Charlie</Text></div>
         </Stack>
     )
     .add("align start", () =>
         <Stack align="start">
-            <div>Alpha</div>
+            <div><Text>Alpha</Text></div>
```
patricklafrance, line 21: No Text component

packages/react-components/src/layout/tests/chromatic/Stack.chroma.jsx
```diff
@@ -10,57 +11,57 @@ function stories(segment) {
...
             <div>Bravo</div>
             <div>Charlie</div>
         </Stack>
     )
     .add("align end", () =>
         <Stack align="end">
-            <div>Alpha</div>
-            <div>Bravo</div>
-            <div>Charlie</div>
+            <div><Text>Alpha</Text></div>
```
patricklafrance, line 28: No Text component

packages/react-components/src/layout/tests/chromatic/Stack.chroma.jsx
```diff
@@ -10,57 +11,57 @@ function stories(segment) {
...
+            <div><Text>Charlie</Text></div>
         </Stack>
     )
     .add("vertical align start", () =>
         <Stack verticalAlign="start" style={{ height: "200px" }}>
-            <div>Alpha</div>
-            <div>Bravo</div>
-            <div>Charlie</div>
+            <div><Text>Alpha</Text></div>
+            <div><Text>Bravo</Text></div>
```
patricklafrance, line 43: etc.. etc..

packages/react-components/src/link/src/Link.css
```diff
@@ -2,7 +2,8 @@
     display: inline-block;
     border: none;
     background-color: var(--transparent);
-    color: var(--marine-900);
+    color: var(--primary-text);
+    /* add inherit */
```
patricklafrance, line 6: TODO here

packages/react-components/src/link/src/Link.css
```diff
@@ -180,62 +189,63 @@
 
 /* VARIATIONS */
 /* PRIMARY */
-.o-ui-link-primary {
-    color: var(--primary-500);
+.o-ui-link.o-ui-link-primary {
```
patricklafrance, line 192: I don't think you need `.o-ui-link` here

packages/react-components/src/link/src/Link.css
```diff
@@ -180,62 +189,63 @@
...
-    color: var(--primary-700);
+    color: var(--brand-text-hover);
 }
 
 /* PRIMARY | STATE | FOCUS */
 .o-ui-link:focus.o-ui-link-primary,
 .o-ui-link-focus.o-ui-link-primary {
-    background-color: var(--primary-50);
-    box-shadow: 0 0 0 2px var(--primary-50);
+    background-color: var(--foreground-color-brand-light);
```
patricklafrance, line 205: Seems wrong to use a foreground color to set a background color?

packages/react-components/src/link/src/TextLink.jsx
```diff
@@ -107,7 +107,7 @@ export function InnerTextLink(props) {
         },
         text: {
             size,
-            className: "o-ui-link-text"
+            className: "o-ui-link-text o-ui-text-color-inherit"
```
patricklafrance, line 110: Hmmm? Since you always render `o-ui-text-color-inherit` why don't you set it in the CSS then?

packages/react-components/src/list/src/ListItem.jsx
```diff
@@ -8,6 +8,10 @@ const propTypes = {
      * A list item can vary in size.
      */
     size: oneOf(["inherit"]),
+    /**
+     * Default classes
+     */
+    className: string,
```
patricklafrance, line 14: We don't document the "className" prop since it's implicit that every component support it.

packages/react-components/src/list/src/ListItem.jsx
```diff
@@ -31,6 +35,7 @@ export function InnerListItem({
                 rest,
                 {
                     size,
+                    className:"o-ui-text-color-inherit",
```
patricklafrance, line 38: I believe it should go in the CSS since it's not dynamic

packages/react-components/src/lozenge/src/Lozenge.css
```diff
@@ -33,7 +33,16 @@
...
+    color: var(--static-white);
+}
+
+.o-ui-lozenge.o-ui-lozenge-primary .o-ui-lozenge-text {
+    color: var(--static-white);
+}
+
+.o-ui-lozenge.o-ui-lozenge-primary .o-ui-lozenge-icon {
+    fill: currentColor;
+    color: inherit;
```
patricklafrance, line 47: I don't fully understand your inherit strategy, I have a feeling it might not be fully consistant across all the components?

If I am following well in this case you should have provided `color: "inherit` to the Icon component instead of doing it in CSS

packages/react-components/src/paragraph/src/Paragraph.jsx
```diff
@@ -42,7 +46,8 @@ export function InnerParagraph(props) {
                     size,
                     className: cssModule(
                         "o-ui-p",
-                        normalizeSize(size)
+                        normalizeSize(size),
+                        "o-ui-text-color-inherit"
```
patricklafrance, line 50: 3 things:

1- How does paragraph `inherit` and `color: "inherit"` props cope together?
2- You don't even use the color prop you added since you always render the "o-ui-text-color-inheri" class instead.

packages/react-components/src/text/src/Text.jsx
```diff
@@ -3,10 +3,12 @@ import "./Text.css";
...
 import { forwardRef } from "react";
-import { mergeProps, normalizeSize, slot, useStyleProps } from "../../shared";
+import { mergeClasses, mergeProps, normalizeSize, slot, useStyleProps } from "../../shared";
 
-export function getTextClass(size) {
-    return `o-ui-text-${normalizeSize(size)}`;
+export function getTextClass(size, color) {
+    const colorClass = color ? `o-ui-text-color-${color}`: "";
+
+    return `o-ui-text-${normalizeSize(size)} ${colorClass} `;
```
patricklafrance, line 11: Do these things directly in the component render function.

packages/react-components/src/theme-provider/src/ThemeProvider.css
```diff
@@ -1,6 +1,191 @@
...
+    --shadow-lightness: 3%;
+    --shadow-opacity: 1;
+    /* FOCUS */
+    --o-ui-focus-ring-color: var(--primary-800);
+}
+
 /* APRICOT */
-/* APRICOT | LIGHT */
-.o-ui-apricot-theme.o-ui-light-color-scheme {
+.o-ui-apricot-theme {
```
patricklafrance, line 188: you shouldn't remove `.o-ui-light-color-scheme`. Without a color scheme it shouldn't work.

packages/react-components/src/theme-provider/src/ThemeProvider.css
```diff
@@ -42,23 +212,8 @@
...
-    --primary-600: black;
-    --primary-700: black;
-    --primary-800: black;
-    --primary-900: black;
-}
-
 /* DESKTOP */
-/* DESKTOP | LIGHT */
-.o-ui-desktop-theme.o-ui-light-color-scheme {
+.o-ui-desktop-theme {
```
patricklafrance, line 216: you shouldn't remove `.o-ui-light-color-scheme`. Without a color scheme it shouldn't work.

packages/react-components/src/button/src/Button.css
patricklafrance: Also, it creates a discrepancy between how Tachyons is structured:

image
image

As you can see in the screenshots, tachyons does handle primary as a product and brand as "marine", "moonstone" etc..

packages/react-components/src/layout/tests/chromatic/Stack.chroma.jsx
patricklafrance: After discussion we might want to do this.

packages/foundation/src/atoms/core/colors/brand.css
fraincs: We can't leave this at root for 2 reasons, first root could clash with other imported CSS using  custom properties at root too, second using root has a drawback ` we can’t rely on inheritance for functions that need to be re-calculated. That’s because each property value is computed once per element (in our case, the html root element), and then the computed value is inherited. By defining our function at the document root, we don’t make the entire function available to descendants — only the computed result of our function.`

tl:dr: for a custom property to be available everywhere and used in other custom properties declarations it needs to be attached to a common class and not root.

packages/foundation/src/atoms/core/foundations/index.css
fraincs: these vars aka custom properties are set in themeProvider, I agree these should be atomic classes exposed by Tachyons down the line, I am removing this folder until I expose these classes as discussed.

packages/foundation/tests/shadows.chroma.jsx
fraincs: These indirection classes are not created. If we expose these classes it makes sense to use `foreground-color-1` instead.

packages/react-components/src/button/src/Button.css
fraincs: This is not a novel concept, it's a colour that won't bulge based on a theme. Google uses it on Youtube CSS, there is more examples in the wild. How would you make sure a colour doesn't change?

A solid button does not change it's text colour when in dark mode hence it cannot use `--white```. Thoughts?

packages/react-components/src/button/src/Button.css
fraincs: White is modified by the Theme. Static White is a colour that won't change ever.

packages/react-components/src/button/src/Button.jsx
fraincs: I think we talked about it and it was alright, since `button-icon` implies it's the icon in the button.

packages/react-components/src/checkbox/src/Checkbox.css
fraincs: that way the checkmark is already in place and won't have a weird transition

packages/react-components/src/link/src/Link.css
fraincs: it was needed for specificity reasons
```
.o-ui-link {
...    color: var(--primary-text);
...
}
```

since link has a color set we need more specificity.

fraincs: Since we now set the marine900 at the themeprovider level this won't be necessary.

packages/react-components/src/lozenge/src/Lozenge.css
fraincs: Provide color: "inherit to a component - not anymore
o-ui-text-color-inherit - not anymore
we are lefe with these occasions where we need an icon to be the same color as the text while reacting to hover / focus / active states.

packages/react-components/src/theme-provider/src/ThemeProvider.css
fraincs: Both were the same that's why, I will be more explicit in my declarations.

packages/react-components/src/button/src/Button.css
fraincs: There is no logic, it's months of design piling up. A cleaning could be done with the product designers.

packages/react-components/src/accordion/src/Accordion.css
fraincs: Primary is used in a hierarchy/importance context.

packages/react-components/src/button/src/Button.css
fraincs: https://spectrum.adobe.com/page/color/#Static-colors see here

fraincs: or here `var(--yt-spec-static-brand-white)` from youtube

packages/foundation/src/atoms/core/colors/brand.css
patricklafrance: I don't think the first argument is a valid one. Since this is a foundation, this is expected to be available. It is the responsability of the consumer apps to ensure their CSS properties doesn't clash with the foundation.

Do you have a specific use case which we currently have that's doesn't work without scoping these properties?

patricklafrance: If we must scope these properties, I believe the way to go is to prefix the property names with "o-ui", not scoping to [class^="o-ui-"].

packages/foundation/tests/shadows.chroma.jsx
patricklafrance: As discussed in a call yesterday, we should create them!

packages/react-components/src/accordion/src/Accordion.css
patricklafrance: As discussed in a call yesterday, this might be confusing.

packages/react-components/src/button/src/Button.css
patricklafrance: Well IMO if I see `--white` I know it doesn't change.

If I see `--foreground-color` I know it might change depending on the current theme

packages/react-components/src/button/src/Button.jsx
patricklafrance: Yes you're right

packages/foundation/tests/shadows.chroma.jsx
fraincs: Where would you define them? In Tachyons? If so any name idea, Alias?

packages/react-components/src/button/src/Button.css
fraincs: Here the box-shadow is used to create a border, visually it's a border (1px solid) and not a shadow per se, this is why I used the border variable.

packages/foundation/src/atoms/core/colors/brand.css
fraincs: `
.o-ui-checkbox-invalid .o-ui-checkbox-box::before {
    --o-ui-focus-ring-color: var(--critical-border);
}
`

here is an example, redefining the focus ring color that is used here `--o-ui-focus-ring-inset-md: 0 0 0 2px var(--o-ui-focus-ring-color);` that way you don't have to remember / redeclare the ` 0 0 0 2px` part and by mistake change the default value.

packages/foundation/tests/shadows.chroma.jsx
fraincs: done, might need to be moved and expand upon later on after discussions with designers

docs/getting-started/Tachyons.stories.mdx
```diff
@@ -91,15 +91,15 @@ Since Tachyons original colors have been replaced with [Orbit Design colors](?pa
 
 The following classes are available:
 
-- `b--[color-name]-[50-900]` (ex. `b--sunray-50`)
-- `hover-b--[color-name]-[50-900]` (ex. `hover-b--sunray-50`)
-- `focus-b--[color-name]-[50-900]` (ex. `focus-b--sunray-50`)
+- `b--[color-name]-[50-900]` (ex. `b--o-ui-sunray-50`)
+- `hover-b--[color-name]-[50-900]` (ex. `hover-b--o-ui-sunrayunray-50`)
```
patricklafrance, line 95: Why add "o-ui" on Tachyons classes?

I mean we did it on css props because someone could use that one but those classes are defined in an atomic css style I dont see why an app would define those.

It also get very long and kindof unpleasant to use IMO

fraincs: bad find and replace

packages/foundation/src/atoms/roles/colors.css
```diff
@@ -0,0 +1,8 @@
+html * {
+    /* LIGHT */
+    /* STATIC */
+    --o-ui-static-white-light: var(--o-ui-white);
+    /* DARK */
+    /* STATIC */
+    --o-ui-static-white-dark: var(--o-ui-white);
```
patricklafrance, line 7: Not sure I get this one.

You told me static is used to identify a color that doesn't change when switching between theme.

So why having a light & dark static white?

I would have expected a single prop called "o-ui-static-white"

packages/foundation/src/atoms/roles/icon.css
```diff
@@ -0,0 +1,24 @@
...
+    --o-ui-icon-positive-strong-light: var(--o-ui-green-900);
+    --o-ui-icon-warning-light: var(--o-ui-sunray-500);
+    --o-ui-icon-warning-strong-light: var(--o-ui-sunray-900);
+    --o-ui-icon-info-light: var(--o-ui-marine-500);
+    /* DARK */
+    --o-ui-icon-1-dark: var(--o-ui-static-white);
+    --o-ui-icon-2-dark: var(--o-ui-cloud-500);
+    --o-ui-icon-primary-dark: var(--o-ui-primary-700);
+    --o-ui-icon-danger-dark: var(--o-ui-alert-700);
+    --o-ui-icon-danger-strong-dark: var(--o-ui-alert-50);
```
patricklafrance, line 18: I am not sure about this "strong" and "pale" naming. It seem to go against our current naming strategy which is to use "1", "2", "3", etc....

I am not close to it but I do want to make sure our naming is consistent.

packages/foundation/src/atoms/roles/colors.css
fraincs: it was for consistency, but this is a valid point

packages/foundation/src/atoms/roles/icon.css
fraincs: at first I went light but light-light war a smell, pale felt wrong too, going with 1 - 2 is ok although the consistency wouldn't be there as not every colour has a "pale" version

PR #444, opened 2021-02-03
Added minimum needed to allow typescript compilation in react-components

alexasselin008:
- ESLint linting for typescript
- type checking in yarn lint + ci
- tsconfig in react-components
- typescript preset in  babel  transpilation


packages/react-components/package.json
```diff
@@ -6,16 +6,33 @@
...
+        "clean": "rimraf dist",
+        "type-check": "tsc --noEmit",
+        "type-check:watch": "yarn run type-check -- --watch"
+    },
+    "devDependencies": {
+        "@types/lodash": "4.14.168",
+        "@types/react-with-styles": "4.0.5",
+        "@types/react-dates": "21.8.1",
+        "@types/react-dom": "17.0.0",
+        "@types/react": "17.0.0"
```
patricklafrance, line 35: All dev dependencies must go at the root package.json

packages/react-components/package.json
```diff
@@ -6,16 +6,33 @@
...
-        "build": "babel src -d dist --config-file ./babel.config.js --copy-files",
+        "build": "yarn run build:js",
+        "build:types": "tsc --emitDeclarationOnly",
+        "build:js": "babel src -d dist --config-file ./babel.config.js --copy-files --extensions .js,.jsx,.ts,.tsx ",
         "prewatch": "yarn clean",
         "watch": "run-p \"build --watch\"",
-        "clean": "rimraf dist"
+        "clean": "rimraf dist",
+        "type-check": "tsc --noEmit",
+        "type-check:watch": "yarn run type-check -- --watch"
```
patricklafrance, line 28: you don't to specify "run" with yarn, it's implicit

packages/react-components/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "allowJs": true,
+    "esModuleInterop": true,
+
+    // Ensure that .d.ts files are created by tsc, but not .js files
+    "declaration": true,
+    // Ensure that Babel can safely transpile files in the TypeScript project
+    "isolatedModules": true
+  },
+  "include": ["src"],
+  "exclude": ["./src/*/tests/**", "./src/*/docs/**"]
```
patricklafrance, line 23: look weird that it's not the same path syntaxe for "include" and "exclude" ? why excluse need "./" and not "include" ?

packages/react-components/package.json
patricklafrance: what's the meaning of "type-check" and "type-check:watch" ? what's their use case? when would use use them?

PR #445, opened 2021-02-05
Convert folder Shared to TypeScript

alexasselin008:
Convert all the "shared" folder to typescript


packages/react-components/src/shared/src/augmentElement.tsx
```diff
@@ -0,0 +1,37 @@
+import { isString } from "lodash";
```
patricklafrance, line 1: Why rename to tsx? The code doesn't contains any JSX.

package.json
```diff
@@ -91,6 +91,8 @@
     "@types/react-dates": "21.8.1",
     "@types/react-dom": "17.0.0",
     "@types/react": "17.0.0",
+    "@types/react-is": "17.0.0",
+    "@types/resize-observer-browser": "0.1.5",
```
patricklafrance, line 95: How come we need those types? Pretty sure I remove all the polyfill code for resize-observe and only use the native api

packages/react-components/src/shared/src/assertion.ts
```diff
@@ -3,6 +3,6 @@
 
 import { isNil } from "lodash";
 
-export function isNilOrEmpty(value) {
+export function isNilOrEmpty(value: any): value is null | undefined | "" {
```
patricklafrance, line 6: Is it a return type? `value is null | undefined | ""`

packages/react-components/src/shared/src/checkable.ts
```diff
@@ -1,19 +1,24 @@
 import { createContext, useContext } from "react";
 import { isNil } from "lodash";
 
-export const CheckableContext = createContext();
+interface CheckableContextProps {
+    checkedValue?: boolean;
+    [x: string]: any;
+}
```
patricklafrance, line 7: Props check for Context, yessss!!!

packages/react-components/src/shared/src/checkable.ts
```diff
@@ -1,19 +1,24 @@
...
+
+export const CheckableContext = createContext<CheckableContextProps>(null);
 
 export function useCheckableContext() {
     const context = useContext(CheckableContext);
 
     if (!isNil(context)) {
-        return [context, true];
+        return [context, true] as const;
```
patricklafrance, line 15: I would naively think you don't to do `as const` and simply set the return type of the function as an array

packages/react-components/src/shared/src/checkable.ts
```diff
@@ -24,8 +29,8 @@ export function useCheckableProps({ value }) {
             ...rest
         };
 
-        return [props, true];
+        return [props, true] as const;
```
patricklafrance, line 32: Same here

packages/react-components/src/shared/src/focusableTreeWalker.ts
```diff
@@ -14,21 +14,25 @@ const FocusableElement = [
     "audio[controls]",
     "video[controls]",
     "[contenteditable]"
-];
+] as const;
```
patricklafrance, line 17: I am not sure why you have to do this. This code have been copied from Spectrum which is in TS and they don't have to use `as const`.

See https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/focus/src/FocusScope.tsx#L152

packages/react-components/src/shared/src/focusableTreeWalker.ts
```diff
@@ -14,21 +14,25 @@ const FocusableElement = [
...
 
 export const FocusableElementSelector = [...FocusableElement, "[tabindex]"].join(",");
 
 export const TabbableElementSelector = [...FocusableElement, "[tabindex]:not([tabindex=\"-1\"])"].join(":not([tabindex=\"-1\"]),");
 
-export function createFocusableTreeWalker(root, { tabbable } = {}) {
+interface FocusableTreeWalkerOptions {
```
patricklafrance, line 23: 👍🏻 👍🏻 👍🏻

packages/react-components/src/shared/src/match.ts
```diff
@@ -1,7 +1,7 @@
 import { isFunction, isNil } from "lodash";
 
 // Took from https://github.com/tailwindlabs/headlessui/blob/develop/packages/%40headlessui-react/src/utils/match.ts
-export function match(value, lookup, ...args) {
+export function match<TValue extends string | number = string, TReturnValue = unknown>(value: TValue, lookup: Record<TValue, TReturnValue | ((...args: any[]) => TReturnValue)>, ...args: any[]): TReturnValue {
```
patricklafrance, line 4: TS is less appealing for these use cases haha

packages/react-components/src/shared/src/mergeClasses.ts
```diff
@@ -0,0 +1,13 @@
+export function mergeClasses(...values: (string | false)[]) {
```
patricklafrance, line 1: Should you add a return type?

packages/react-components/src/shared/src/mergeProps.ts
```diff
@@ -3,19 +3,24 @@ import { isFunction, isNil, isUndefined } from "lodash";
 import { mergeClasses } from "./mergeClasses";
 import { mergeRefs } from "./useMergedRefs";
 
-// Usefull to compose a weak map key with multiple objects.
-export class CompositeKeyWeakMap {
-    _root = new WeakMap();
+interface Child<T> {
```
patricklafrance, line 6: I would call it Node or WeakMapNode

patricklafrance: I guess the T could be another type, something like Pair or WeakMapPair

packages/react-components/src/shared/src/mergeProps.ts
```diff
@@ -27,33 +32,34 @@ export class CompositeKeyWeakMap {
...
     }
 }
 
 ////////
 
-const cache = new CompositeKeyWeakMap();
+const cache = new CompositeKeyWeakMap<any>();
 
-function memoMerge(x, y, fct) {
+function memoMerge<TFirst, TSecond, TReturnValue>(x: TFirst, y: TSecond, fct: (y: TSecond, x: TFirst) => TReturnValue) {
```
patricklafrance, line 62: How come in "fct" x and y are reversed?

patricklafrance: TReturnValue => TReturn ?

packages/react-components/src/shared/src/mergeProps.ts
```diff
@@ -107,12 +113,16 @@ function merge(props, newProps) {
     return props;
 }
 
-export function mergeProps(...args) {
+// taken from: https://stackoverflow.com/questions/51603250/typescript-3-parameter-list-intersection-type/51604379#51604379
+type TupleTypes<T> = { [P in keyof T]: T[P] } extends { [key: number]: infer V } ? V : never;
+type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never;
```
patricklafrance, line 118: Ouch.

Should these types be extracted somewhere global?

packages/react-components/src/shared/src/mergeProps.ts
```diff
@@ -107,12 +113,16 @@ function merge(props, newProps) {
...
+
+export function mergeProps<T extends Record<string, any>[]>(...args: T) {
```
patricklafrance, line 120: Does the return type are implicits?

packages/react-components/src/shared/src/omitProps.ts
```diff
@@ -0,0 +1,10 @@
+export function omitProps<TObject extends Record<string, any>, TProps extends keyof TObject>(obj: TObject, props: TProps[]): Omit<TObject, TProps> {
+    return (Object.keys(obj) as TProps[]).
+        reduce((acc, x) => {
+            if (!props.includes(x)) {
+                acc[x] = obj[x];
+            }
+
+            return acc;
+        }, {} as TObject);
+}
```
patricklafrance, line 10: The complexity of this thing explode in TS.

Does "props" could be a Record<string, any> ? Would make this simpler.

Something like:

```
export function omitProps(obj: Record<string, any>, props: Record<string, any>) => Record<string, any> {
   ....
}
```

packages/react-components/src/shared/src/resolveChildren.ts
```diff
@@ -0,0 +1,19 @@
...
+// Support first level fragment: https://github.com/facebook/react/issues/11859.
+export function resolveFragment(children: ReactNode) {
+    if (!isNil(children) && (children as ReactElement).type === Fragment) {
+        return (children as ReactElement).props.children;
+    }
+
+    return children;
+}
+
+export function resolveChildren<T extends any[]>(children: ReactNode | ((args: T[]) => ReactNode), renderProps?: T): ReactNode {
```
patricklafrance, line 13: I think `renderProps` should be a `Record<string, any>`. It's basically an object literal.

Not sure about the `args: T[]`, There shouldn't be any case where the function can receive an array of renderProps

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -1,19 +1,31 @@
-import { Children, useMemo } from "react";
-import { augmentElement, resolveChildren } from "../../shared";
+import { augmentElement, resolveChildren } from "..";
 import { isNil, isString, isUndefined } from "lodash";
+import React, { Children, ComponentType, ReactElement, ReactNode, useMemo } from "react";
```
patricklafrance, line 3: I don't think you need to import the React object.

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -22,7 +34,8 @@ export function getSlots(children, { _ = {}, ...slots }) {
 
         Children.forEach(children, x => {
             if (!isNil(x)) {
-                const slot = (x.props && x.props["slot"]) ?? (x.type && x.type[SLOT_KEY]);
+                const reactComponent = children as ReactElement;
+                const slot = (reactComponent.props && reactComponent.props["slot"]) ?? (reactComponent.type && (reactComponent.type as ElementTypeWithSlot)[SLOT_KEY]);
```
patricklafrance, line 38: I am not sure about this code you are doing:

`reactComponents.props` is the result of `children as ReactElement` but in the original code, it's `x.props` so basically this is not using the same arg as the original code.

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -22,7 +34,8 @@ export function getSlots(children, { _ = {}, ...slots }) {
 
         Children.forEach(children, x => {
             if (!isNil(x)) {
-                const slot = (x.props && x.props["slot"]) ?? (x.type && x.type[SLOT_KEY]);
+                const reactComponent = children as ReactElement;
```
patricklafrance, line 37: Couldn't it be done with `children = resolveChildren(children);` instead?

```
children = resolveChildren(children) as ReactElement;
```

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -1,19 +1,31 @@
...
-import { augmentElement, resolveChildren } from "../../shared";
+import { augmentElement, resolveChildren } from "..";
 import { isNil, isString, isUndefined } from "lodash";
+import React, { Children, ComponentType, ReactElement, ReactNode, useMemo } from "react";
 
 const SLOT_KEY = "__slot__";
 
-function slotDecorator(slotName, ElementType) {
-    ElementType[SLOT_KEY] = slotName;
+interface ElementTypeWithSlot {
```
patricklafrance, line 7: ElementTypeWithSlot -> SlottedElement or ElementWithSlot ?

packages/react-components/src/shared/src/styles.ts
```diff
@@ -1,24 +1,24 @@
 import { createContext, useContext } from "react";
 import { isNil } from "lodash";
 
-export const StyleContext = createContext();
+export const StyleContext = createContext<Record<string, any>>(null);
```
patricklafrance, line 4: null is not the same as undefined which was the original value.

packages/react-components/src/shared/src/useChainedEventCallback.ts
```diff
@@ -2,16 +2,16 @@
...
 import { useEventCallback } from "./useEventCallback";
 
-export function createChainedFunction(...funcs) {
-    return (...args) => {
-        funcs
-            .filter(Boolean)
+export function createChainedFunction<T extends (...args: any[]) => void>(...funcs: (T | false)[]) {
+    return (...args: any[]) => {
+        (funcs
+            .filter(Boolean) as T[])
```
patricklafrance, line 8: This is weird, why the casting after filter? `filter` is a native function, shouldn't TS understand it returns an array?

packages/react-components/src/shared/src/useChainedEventCallback.ts
```diff
@@ -2,16 +2,16 @@
 
 import { useEventCallback } from "./useEventCallback";
 
-export function createChainedFunction(...funcs) {
-    return (...args) => {
-        funcs
-            .filter(Boolean)
+export function createChainedFunction<T extends (...args: any[]) => void>(...funcs: (T | false)[]) {
```
patricklafrance, line 5: Instead of the extends thing wouldn't it be easier to read if you introduced an interface, something like `ChainableFunction` ?

packages/react-components/src/shared/src/useDisposables.ts
```diff
@@ -3,24 +3,24 @@ import { useEffect, useState } from "react";
...
             api.add(() => cancelAnimationFrame(id));
         },
 
-        nextFrame(...args) {
+        nextFrame(...args: Parameters<typeof requestAnimationFrame>) {
             api.requestAnimationFrame(...args);
         },
 
-        setTimeout(...args) {
+        setTimeout(...args: Parameters<typeof setTimeout>) {
```
patricklafrance, line 18: Should we use Parameters everywhere we have a rest parameter?

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -2,11 +2,11 @@ import { FocusTarget } from "./focusTarget";
 import { isFunction, isNil } from "lodash";
 import { useMemo } from "react";
 
-class ElementIterator {
+class ElementIterator<T> {
```
patricklafrance, line 5: I don't think you need to make this generic. This iterator is strictly for HtmlElement

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -36,19 +36,36 @@ class ElementIterator {
...
+    keyProp?: string;
+}
+
+export interface FocusElementOptions {
+    onFocus?(element?: Element): void
+    onNotFound?(): void
+    canFocus?(element: Element): boolean;
+}
+
+export interface FocusScope {
```
patricklafrance, line 49: This interface should be exported from `useFocusScope`

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -36,19 +36,36 @@ class ElementIterator {
     }
 }
 
+export interface FocusManagerOptions {
+    keyProp?: string;
+}
+
+export interface FocusElementOptions {
```
patricklafrance, line 43: Would call it `FocusOptions` instead

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -36,19 +36,36 @@ class ElementIterator {
...
     _keyProp;
 
-    constructor(scope, { keyProp } = {}) {
+    constructor(scope: FocusScope, { keyProp }: FocusManagerOptions = {}) {
         this._scope = scope;
         this._keyProp = keyProp;
     }
 
-    _focusElement(element, { onFocus, onNotFound } = {}) {
+    _focusElement(element: Element | HTMLElement, { onFocus, onNotFound }: FocusElementOptions = {}) {
```
patricklafrance, line 65: I think `element: HTMLElement` should be fine.

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -36,19 +36,36 @@ class ElementIterator {
...
         if (!isNil(element)) {
-            if (isFunction(element.focus)) {
-                element.focus();
+            if (isFunction((element as HTMLElement).focus)) {
```
patricklafrance, line 67: Why do you need a cast here? The arg is already typed

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -36,19 +36,36 @@ class ElementIterator {
...
+                (element as HTMLElement).focus();
```
patricklafrance, line 68: Same here

packages/react-components/src/shared/src/assertion.ts
alexasselin008: it's a little bit different of a return type

image

instead of saying the return value is a boolean, we let typescript know that if the return value is true, the value is either null, undefined or an empty string

packages/react-components/src/shared/src/useFocusManager.ts
```diff
@@ -146,7 +163,7 @@ export class FocusManager {
 
             canFocus = !isNil(canFocus) ? canFocus : () => true;
 
-            const index = elements.indexOf(document.activeElement);
+            const index = elements.indexOf(document.activeElement as HTMLElement);
```
patricklafrance, line 166: really need this? It's all native API shouldn't typescript understand all this?

packages/react-components/src/shared/src/useFocusScope.ts
```diff
@@ -1,13 +1,14 @@
...
+import { createContext, RefObject, useCallback, useContext, useMemo } from "react";
 import { isNil } from "lodash";
 import { useRefState } from "./useRefState";
 import { walkFocusableElements } from "./focusableTreeWalker";
+import { FocusScope } from "./useFocusManager";
 
-class DomScope {
-    _scopeRef;
-    _handlersRef;
+class DomScope<T extends HTMLElement> {
```
patricklafrance, line 7: Don't need to be generic it's only meant to deal with HtmlElement

packages/react-components/src/shared/src/assertion.ts
alexasselin008: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

the return value of this function is a boolean. however, we can let typescript know that if the return value is true, then value is either null, undefined or an empty string

packages/react-components/src/shared/src/useFocusScope.ts
```diff
@@ -16,27 +17,27 @@ class DomScope {
...
+    isInScope(element: T) {
         return this.elements.some(x => x.contains(element));
     }
 }
 
 export function useFocusScope() {
-    const [scopeRef, setScope] = useRefState([]);
-    const [handlersRef] = useRefState([]);
+    const [scopeRef, setScope] = useRefState<HTMLElement[]>([]);
+    const [handlersRef] = useRefState<((elements: Element[], scope: Element[]) => void)[]>([]);
```
patricklafrance, line 37: Should extract a type for this since it's quite complicated and re-used in DomScope

packages/react-components/src/shared/src/useFocusScope.ts
```diff
@@ -73,16 +74,20 @@ export function useFocusScope() {
...
-    const scope = useMemo(() => new DomScope(scopeRef, handlersRef), [scopeRef, handlersRef]);
+    const scope: FocusScope = useMemo(() => new DomScope(scopeRef, handlersRef), [scopeRef, handlersRef]);
 
     const { scope: contextScope } = useFocusContext();
 
     return isNil(contextScope)
-        ? [scope, setRef]
-        : [contextScope, undefined];
+        ? [scope, setRef] as const
+        : [contextScope, undefined] as const;
```
patricklafrance, line 83: Still don't get this `as const` thing ;)

packages/react-components/src/shared/src/augmentElement.tsx
alexasselin008: ```return <Type {...props} />;```

This is jsx!

packages/react-components/src/shared/src/useForwardInputApi.ts
```diff
@@ -8,14 +10,14 @@ const API_METHODS = [
     "checkValidity",
     "reportValidity",
     "setCustomValidity"
-];
+] as const;
 
-export function useForwardInputApi(inputRef) {
-    return targetRef => {
+export function useForwardInputApi<T extends Record<string, any>>(inputRef: MutableRefObject<T | undefined>) {
```
patricklafrance, line 15: Does T is really a Record? The ref will always be bound to a ReactElement so it should always be an HtmlElement

packages/react-components/src/shared/src/useMergedRefs.ts
```diff
@@ -1,7 +1,7 @@
 import { isFunction, isNil } from "lodash";
-import { useCallback } from "react";
+import { ForwardedRef, useCallback } from "react";
```
patricklafrance, line 2: I am getting confused... So there are RefObject, MutabableRefObject and ForwardedRef ?

packages/react-components/src/shared/src/useMergedRefs.ts
```diff
@@ -1,7 +1,7 @@
...
 
-export function assignRef(ref, node) {
+export function assignRef<T>(ref: ForwardedRef<T>, node: T) {
```
patricklafrance, line 4: I am not sure why here it would a specialized ref type. It can be any ref.

packages/react-components/src/shared/src/useMergedRefs.ts
```diff
@@ -26,7 +27,7 @@ export function mergeRefs(...refs) {
     return mergedRef;
 }
 
-export function useMergedRefs(...refs) {
+export function useMergedRefs<T>(...refs: ForwardedRef<T>[]) {
```
patricklafrance, line 30: Same here

packages/react-components/src/shared/src/useMergedRefs.ts
```diff
@@ -11,9 +11,10 @@ export function assignRef(ref, node) {
     }
 }
 
-export function mergeRefs(...refs) {
-    const mergedRef = current => {
+export function mergeRefs<T>(...refs: ForwardedRef<T>[]) {
```
patricklafrance, line 14: Same here

packages/react-components/src/shared/src/useRovingFocus.ts
```diff
@@ -1,25 +1,26 @@
 import { isNil } from "lodash";
 import { useLayoutEffect } from "react";
+import type { FocusScope } from "./useFocusManager";
 
-export function useRovingFocus(scope) {
+export function useRovingFocus(scope: FocusScope) {
     useLayoutEffect(() => {
-        const handleFocus = event => {
+        const handleFocus = (event: Event) => {
```
patricklafrance, line 7: Isn't there something like a specialized FocusEvent ?

packages/react-components/src/shared/src/useRovingFocus.ts
```diff
@@ -1,25 +1,26 @@
...
             scope.elements.forEach(x => {
                 if (x.tabIndex === 0) {
                     x.tabIndex = -1;
                 }
             });
 
-            event.target.tabIndex = 0;
+            (event.target as HTMLElement).tabIndex = 0;
```
patricklafrance, line 14: Since the event is already typed, how come we need to cast?

packages/react-components/src/shared/src/checkable.ts
alexasselin008: This is used for most of the custom hooks you use. 

const [state, setState] = useState(boolean)


without the as const, useState would return an array of boolean and functions

(boolean | (value: boolean) => void)[]
so you wouldn't be able to use setState(false) since setState could be a boolean according to typescript

but in fact, we know its an array of 2 elements, the first one is a bool, and the second one is a function. 

this blog shows the 2 way of typing this. i prefer the 2nd option

https://fettblog.eu/typescript-react-typeing-custom-hooks/#option-2%3A-as-const

packages/react-components/src/shared/src/focusableTreeWalker.ts
alexasselin008: Your right, not needed

packages/react-components/src/shared/src/mergeClasses.ts
alexasselin008: Not needed, it is inferred by typescript since .join returns a string

packages/react-components/src/shared/src/mergeProps.ts
alexasselin008: Ya, since we cast it the last line as a type, typescript infers the return value of this method

packages/react-components/src/shared/src/omitProps.ts
alexasselin008: while simpler to read, this does not leverage typescript.

const testObject = { 
variable1: true,
variable2: false,
}



using my method : 
```
const result = omit(testObject, ["variable1"])
result.variable2 // typescript does not allow this
result.variable1 // this is a bool
```

using a generic Record<string, any> : 
```
const result = omit(testObject, ["variable1"])
result.variable2 // no warning here
result.variable1 // this is type any
```

packages/react-components/src/shared/src/slots.tsx
alexasselin008: <Wrapper>{children}</Wrapper>

this is jsx in this file

packages/react-components/src/shared/src/styles.ts
alexasselin008: good point. 

also, FYI : https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24509#issuecomment-382213106

undefined might be a weird default value

packages/react-components/src/shared/src/useChainedEventCallback.ts
alexasselin008: filter is a method that excludes some items from the list. it does not discriminate types for typescript sadly.

packages/react-components/src/shared/src/useDisposables.ts
alexasselin008: This synthax is when we forward parameters we don't care about to another function. 

I'm not sure wich rest parameter you refer to, but in those scenario it's a good idea

package.json
alexasselin008: The timing is a bit awkward. ResizeObserver type was and is not yet typed in typescript (in the lib.dom.d.ts file).

It's been added in the TypeScript Repo 27 days ago, but there hasn't been a release yet. Most likely, in Typescript 4.1.4 or 4.2, the types will be present and we won't need to import this package. 

TypeScript DOM Lib generator feature link : https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/847

packages/react-components/src/shared/src/omitProps.ts
alexasselin008: I've created a sandbox for this case

https://www.typescriptlang.org/play?#code/MYewdgzgLgBFCm0DyAjAVvYsC8MDeMAUAG4CGATgJakoA28AjAFxzkCu8ANCRdXfACYWAM1K0IXQgF9CheAA8ADiHKxhbMFkrgYIALaUoABVJQAFCHQsASphUATADzQqYAOacYpMAE8AfJ6K5CCKECwulO4A2gC6AJQ2duROEe6e3v74hDA5uTDk8FBs5GAwZqgYWAB0ANbwPhAW6HFxVdl5uQX2bMDwZmakwMCe8nEw2H5ZHdMwlMJlAIRBIRBVkcC0bPaIZqNjeO0zM4PAUfIx47roZzEA3IdHMDIPRwVFJV5D90dSnnhSXggMFsoGSzigrg8Xl8kzi92eCmUqhg6k0UG0pX0hgAgvR5I4ACoVTCwBQIMD2IEghzgyHpGGeAlGYKhGBk+AUoF1HwgeZE9AkvxNNAsfmVKCBFlhGBMqWxBIwJAGKCE4lYRnMlaTA55N7FUrlAXVbmNSxoMakIGylbytqveDdXr9E4jMYTKaPHJzRbLUJrTSbbaNPYez15E43S5mm7fMPPMP5Qr6z7AWPTX74AGWmVqqBw6SyUCQWAFCBsWjGUxR5UmcwIZBGiVRABEZCoNHoDGb8XupfLlagVTbfE7txgAHpx9CfJ4wCA2eRguRCH2K7Wh7wO4Ix5Pp7P5-BFypC+BoImyxXcQpqzi8WZ61Bc54W8Ot12eyvEP2r-IN+3+AwE5ToeS6fheUA-n+I6CEBcA+Io8AwCgIAgPQ3hAA

packages/react-components/src/shared/src/mergeProps.ts
alexasselin008: memoMerge was coded that way. It felt weird to me as well. 
if you look below, you can see 

    const mergeResult = fct(y, x);

packages/react-components/src/shared/src/resolveChildren.ts
alexasselin008: In this case, T means i don't really care what the type is. As long as the function passed as the children parameter accept whatever is passed as the 2nd argument.

packages/react-components/src/shared/src/slots.tsx
alexasselin008: 😱 that was bad

packages/react-components/src/shared/src/useFocusManager.ts
alexasselin008: if you insist i could. However, since this class work for any array of element, i don't see why we would limit ourselves with that

packages/react-components/src/shared/src/useFocusManager.ts
alexasselin008: document.activeElement is of type Element, not HTMLElement. I did not want to modify the code, but according to some forums, activeElement might be elements that are not HTMLElement.

i googled a couple of minutes, and found this answer on a stack overflow

```
.blur() is only guaranteed to exist on HTMLElements, not all Elements. I have frequently seen exceptions caused by unguarded document.activeElement.blur(), because document.activeElement was an SVG element rather than HTML.
```

packages/react-components/src/shared/src/useForwardInputApi.ts
alexasselin008: HTMLInputElement is perfect, thank you

packages/react-components/src/shared/src/augmentElement.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export function createOrAugmentElement<T extends string, P extends HTMLAttributes<T>>(element: ReactElement<P, T> | ElementType, props: P) {
+    if (isValidElementType(element) && !isString(element)) {
+        const Type = element;
+
+        return <Type {...props} />;
+    }
+
+    // The isString in the previous condition assume it can be possible that element is a string. however, if its the case, cloneElement does not support string, so it does not make sense
+    // Therefore, since we are only typing the code and i don't want to introduce regression bugs, i will disable the ts compiler for that line
```
patricklafrance, line 22: Yeah it happens because some components pass children directly to this function and children can be a string.

However, this function is deprecated it should be delete before the release.

packages/react-components/src/shared/src/resolveChildren.ts
```diff
@@ -0,0 +1,19 @@
...
```
patricklafrance, line 13: Could we make custom types for repetive React patterns? For example ReactChildren would make sense for `ReactNode | ((args: T[]) => ReactNode)`

patricklafrance: Might be hard though since you can't really know which args the function signature could have

patricklafrance: Not sure about the generic constraint T.

From what I understand the constraint is to ensure T is an array. How come then you still have to specify T as an array? `((args: T[])`

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -22,7 +34,8 @@ export function getSlots(children, { _ = {}, ...slots }) {
 
         Children.forEach(children, x => {
             if (!isNil(x)) {
-                const slot = (x.props && x.props["slot"]) ?? (x.type && x.type[SLOT_KEY]);
+                const reactComponent = x as ReactElement;
```
patricklafrance, line 37: instead of doing this, could you do:

`Children.forEach(children, x: ReactElement => {`

packages/react-components/src/shared/src/useChainedEventCallback.ts
```diff
@@ -0,0 +1,19 @@
+// Copied from https://github.com/react-bootstrap/react-bootstrap/blob/master/src/createChainedFunction.js.
+
+import { useEventCallback } from "./useEventCallback";
+
+type ChainableFunction = (...args: any[]) => void;
+
+export function createChainedFunction<T extends ChainableFunction>(...funcs: (T | false)[]) {
+    return (...args: any[]) => {
+        (funcs
+            .filter(Boolean) as T[])
```
patricklafrance, line 10: Not a big deal but I have seeing parantheses starting and finishing on a different line like this!

packages/react-components/src/shared/src/useDisposables.ts
```diff
@@ -1,31 +1,35 @@
 import { useEffect, useState } from "react";
 
+// we use function here since we copied the method from another repo and they use this
+// eslint-disable-next-line @typescript-eslint/ban-types
+type DisposableFunction = Function;
```
patricklafrance, line 5: Not sure I get it?

Seems useless to me

packages/react-components/src/shared/src/useEventCallback.ts
```diff
@@ -4,7 +4,7 @@ import { isNil } from "lodash";
 import { useCallback } from "react";
 import { useCommittedRef } from "./useCommittedRef";
 
-export function useEventCallback(callback) {
+export function useEventCallback<T extends (...args: any[]) => void>(callback: T) {
```
patricklafrance, line 7: I don't see why a generic constraint is usefull here.

PR #446, opened 2021-02-09
Explicit returns on typescript functions

alexasselin008:
"typescript-eslint/explicit-function-return-type": ["warn", { allowExpressions: true }]


packages/react-components/src/shared/src/useControllableState.ts
```diff
@@ -42,7 +49,7 @@ function useComputeInitialState<T>(controlledValue: T, initialValue: T, defaultV
     return result(state, isControlled, true);
 }
 
-function computeSubsequentState<T>(controlledValue: T, currentState: T, isControlled: boolean) {
+function computeSubsequentState<T>(controlledValue: T, currentState: T, isControlled: boolean): { newState: T; hasChanged: boolean; } {
```
alexasselin008, line 52: Are we fine with this synthax?

patricklafrance: Looks good, what are the alternatives?

alexasselin008: naming the type using an interface

patricklafrance: Looks fined, finding a name here would be quite hard.

packages/react-components/src/shared/src/useKeyboardNavigation.ts
```diff
@@ -14,7 +14,11 @@ interface KeyboardNavigationBindings {
     last: string[];
 }
 
-export function useKeyboardNavigation(focusManager: FocusManager, { previous = [], next = [], first = [], last = [] }: KeyboardNavigationBindings, { onSelect }: KeyboardNavigationOptions = {}) {
+interface KeyboardNavigationAPI {
```
patricklafrance, line 17: Usually for inputs (props) we usually suffix the interface name with "Props". For example, "TextInputProps".

Should we have a similar naming convention for return types?

packages/react-components/src/shared/src/useMergedRefs.ts
```diff
@@ -11,8 +11,8 @@ export function assignRef<T>(ref: ForwardedRef<T>, node: T) {
     }
 }
 
-export function mergeRefs<T>(...refs: ForwardedRef<T>[]) {
-    const mergedRef = (current: T) => {
+export function mergeRefs<T>(...refs: ForwardedRef<T>[]): (current: T) => void {
+    const mergedRef = (current: T): void => {
```
patricklafrance, line 15: Do you think we could somehow also exempt these local arrow functions from the rule?

packages/react-components/src/shared/src/useResizeObserver.ts
```diff
@@ -21,15 +21,15 @@ function getResizeObserver() {
...
+export const useResizeObserver = (element: Element, onResize: (entry: ResizeObserverEntry) => void, { box }: ResizeObserverOptions = {}): void => {
     useEffect(() => {
         if (!isNil(element)) {
             getResizeObserver().observe(element, { box });
 
             handlersMap.set(element, onResize);
         }
 
-        return () => {
+        return (): void => {
```
patricklafrance, line 32: :(

packages/react-components/src/shared/src/useMergedRefs.ts
patricklafrance: Not a big deal if we can't

patricklafrance (approved):
Looks good to me!

packages/react-components/src/shared/src/useKeyboardNavigation.ts
alexasselin008: I feel this might be hard finding an appropriate standard for this. Maybe one will emerge naturally

PR #449, opened 2021-02-16
box/layout/text components to typescript

alexasselin008:


packages/react-components/src/shared/src/component-types.ts
```diff
@@ -0,0 +1,28 @@
+import { ElementType, ForwardedRef, PropsWithChildren, ReactNode } from "react";
```
alexasselin008, line 1: Is the file name alright? 

The plan is to have this file be refered by all the components to import redundant types

packages/react-components/src/shared/src/component-types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface ForwardedRefAttribute<T = any> {
```
alexasselin008, line 3: Generics types won't be used, or they will only be used with the default value. However, adding them to the types describe the type better than 

```
export interface ForwardedRefAttribute {
    /**
     * @ignore
     */
    forwardedRef: ForwardedRef<any>
}
```

packages/react-components/src/shared/src/component-types.ts
```diff
@@ -0,0 +1,28 @@
...
+    * An HTML element type or a custom React element type to render as.
+    */
+    as?: As;
+}
+
+export interface ChildrenAttribute {
+    children?: ReactNode;
+}
+
+export type AsAttributeWithChildren<As extends ElementType = ElementType> = PropsWithChildren<AsAttribute<As>>;
```
alexasselin008, line 21: I ran small analysis on the code base. Those might not be 100% accurate, but they are close enough i think

86 components total (search on `.displayName = ` in .jsx files)
75 of them have the as property
52 of them as children property
45 have both children and as property
12 have slots

With those stats 

I could make a BaseAttributes interface with only forwarded ref
I could make a BaseWithChildrenAttributes interface with forwarded ref and children
I could make a BaseWithAsAndChildrenAttributes interface with forwarded ref, children and as

Components with as but no children would be Base + AsAttributes, no shared interface to group them
Components with children but no as would be Base + ChildrenAttributes, no shared interface to group them

SlotAttribute would be extended for the 15 components using slots, in addition to one of the 3 base types

alexasselin008: I need to find appropriate names for those interface if you have any ideas.

alexasselin008: Also, the Box components would then be 
```
export interface StrictBoxProps extends BaseWithAsAndChildrenAttributes, SlotAttribute
```

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,37 @@
+import { AsAttributeWithChildren, ForwardedRefAttribute, SlotAttribute, omitProps } from "../../shared";
+import { ReactElement, forwardRef } from "react";
+
+export interface StrictBoxProps extends AsAttributeWithChildren, SlotAttribute, ForwardedRefAttribute {
+}
+
+export interface BoxProps extends StrictBoxProps {
```
alexasselin008, line 7: Pattern taken from Semantic-UI. 

- React-spectrum : doesn't have as property
- Chakra : doesn't need because they specifiy the as property type explicitly. Smart, but requires more code than we would like to make it work
- React-bootstrap : extends React.HTMLAttributes, which in my opinion is misleading since we allow custom components, and that component might not expect those properties. I think i'd rather have an untyped property bag than misleading props

packages/react-components/src/shared/src/index.ts
```diff
@@ -48,3 +48,6 @@ export * from "./useRefState";
 export * from "./slots";
 export * from "./checkable";
 export * from "./styles";
+
+// Types
+export type { SlotAttribute, ChildrenAttribute, AsAttribute, ForwardedRefAttribute } from "./component-types";
```
alexasselin008, line 53: This is a little bit irritating, but since there is no `export type * from ...`, i tried `export * from ...` but it causes an error. I assume it's because `export * from` a file that contains nothing (after the babel transpilation) isn't working properly

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <Wrapper
+            {...rest}
+            ref={forwardedRef}
+        >
+            {children}
+        </Wrapper>
+    );
+}
+
+export const Box = forwardRef<any, PropsWithoutForwardedRef<BoxProps>>((props, ref) => (
```
alexasselin008, line 41: required in order to not autocomplete the "forwardedRef" property in the Box component

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -154,13 +121,13 @@ export function InnerFlex({
...
-}) {
+}: FlexProps): ReactElement {
     const isGapSupported = useIsGapSupported(noGap);
 
     // Normalize values until Chrome support `start` & `end`, https://developer.mozilla.org/en-US/docs/Web/CSS/align-items.
-    alignContent = alignContent && alignContent.replace("start", "flex-start").replace("end", "flex-end");
-    alignItems = alignItems && alignItems.replace("start", "flex-start").replace("end", "flex-end");
-    justifyContent = justifyContent && justifyContent.replace("start", "flex-start").replace("end", "flex-end");
+    const normalizedAlignContent = alignContent && alignContent.replace("start", "flex-start").replace("end", "flex-end");
+    const normalizedAlignItems = alignItems && alignItems.replace("start", "flex-start").replace("end", "flex-end");
```
alexasselin008, line 129: We do this in order to not mutate the component's props

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,69 @@
...
-        "stretch",
-        "baseline",
-        "first baseline",
-        "last baseline",
-        "safe center",
-        "unsafe center"
-    ]),
+    justifyContent?: "start" | "end" | "center" | "left" | "right" | "space-between" | "space-around" | "space-evenly" | "stretch" | "baseline" | "first baseline" | "last baseline" | "safe center" | "unsafe center"
+    /**
+     * TODO: document this property
```
alexasselin008, line 35: This props was not documented. Do we leave TODO here and come back at the end of the conversion?

BobbyWhiskey:
Doc de Text avant:
image

Apres:
image

alexasselin008:
Box before : 
before
Box After : 
after

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,48 @@
...
+     * React children
+    */
+    children?: ReactNode;
+    /**
+    * @ignore
+    */
+    forwardedRef: ForwardedRef<any>
+}
+
+export type BoxProps = InnerPropsToProps<InnerBoxProps>;
```
alexasselin008, line 23: InnerBoxProps is used by the inner component and the storybook documentation generation. It contains a forwardedRef, but no ref attributes.

BoxProps is the resulting props interface for the box component. It contains a ref attribute, and no forwardedRef

packages/react-components/src/layout/src/Inline.tsx
```diff
@@ -52,29 +57,29 @@ export function InnerInline({
     wrap,
     forwardedRef,
     ...rest
-}) {
+}: InnerInlineProps): ReactElement {
     const alignProps = useFlexAlignment("horizontal", align, verticalAlign);
 
+    const extraFlexProps: Partial<FlexProps> = {
```
alexasselin008, line 63: we can't type inline elements, so we extract it and type it before the render function

alexasselin008: The alternative for the inline is 

```
{
    gap: gap !== 0 ? gap : undefined,
    wrap: !isNil(wrap) ? "wrap" : undefined,
    ref: forwardedRef
} as Partial<FlexProps>
``` 

However, since it's a type casting and not a typing, there is no validation that the value passed are actually valid

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export function InnerBox(props: InnerBoxProps): ReactElement {
+    const {
+        as: Wrapper = "div",
+        children,
+        forwardedRef,
+        ...rest
+    } = omitProps(props, ["slot"]);
+
```
alexasselin008, line 32: this code looks weird. We accept slot as a property, however we never use it. 

Why do we accept it as a props then? Should i remove it from the interface?

packages/react-components/src/text/src/Text.tsx
```diff
@@ -61,9 +61,9 @@ export function InnerText(props: TextProps): ReactElement {
     );
 }
 
-export const Text = forwardRef<any, PropsWithoutForwardedRef<TextProps>>((props, ref) => (
+export const Text = slot("text", forwardRef<any, PropsWithoutForwardedRef<TextProps>>((props, ref) => (
```
patricklafrance: Is there a way to simplify all this? Not a big fan of `PropsWithoutForwardedRef` and `InnerPropsToProps`.

Would it be possible to do something like:

```
export interface TextProps {
    /**
     * A text can vary in size.
     */
    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
    /**
     * An HTML element type or a custom React element type to render as.
     */
    as?: ElementType;
    /**
     * [Slot](?path=/docs/getting-started-slots--page) to render into.
     */
    slot?: string;
    /**
     * React children
     */
    children?: ReactNode;
}

export interface InnerTextProps extends TextProps {
    /**
    * @ignore
    */
    forwardedRef: ForwardedRef<any>;
}

.....

export function InnerText(props: InnerTextProps): ReactElement {
}

.....

export const Text = slot("text", forwardRef<any, TextProps>((props, ref) => (
    <InnerText {...props} forwardedRef={ref} />
)));
```

Could even be:

```
export interface InnerTextProps {
    /**
     * A text can vary in size.
     */
    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
    /**
     * An HTML element type or a custom React element type to render as.
     */
    as?: ElementType;
    /**
     * [Slot](?path=/docs/getting-started-slots--page) to render into.
     */
    slot?: string;
    /**
     * React children
     */
    children?: ReactNode;
    /**
    * @ignore
    */
    forwardedRef: ForwardedRef<any>;
}

....

export const Text = slot("text", forwardRef<any, Omit<InnerTextProps, "forwardedRef">((props, ref) => (
    <InnerText {...props} forwardedRef={ref} />
)));
```

alexasselin008: There has been a couple iteration, but what i found the most effective is this : 

```
export interface InnerTextProps {
    /**
     * A text can vary in size.
     */
    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
    /**
     * An HTML element type or a custom React element type to render as.
     */
    as?: ElementType;
    /**
     * [Slot](?path=/docs/getting-started-slots--page) to render into.
     */
    slot?: string;
    /**
     * React children
     */
    children?: ReactNode;
    /**
    * @ignore
    */
    forwardedRef: ForwardedRef<any>;
}

export type TextProps = InnerPropsToProps<InnerTextProps>;



export function InnerText(props: InnerTextProps): ReactElement {
ééé
}


export const Text = slot("text", forwardRef<any, TextProps>((props, ref) => (
    <InnerText {...props} forwardedRef={ref} />
)));
```
The issue with the 2nd option you proposed, is that if you need the BoxProps somewhere else in the code, you can't access it since you never named it. 

for instance : 

```

        <Flex
            {...mergeProps<Partial<FlexProps>[]>(
                rest,
                alignProps,
                {
                    gap: gap !== 0 ? gap : undefined,
                    wrap: !isNil(wrap) ? "wrap" : undefined,
                    ref: forwardedRef
                }
            )}
        >
            {children}
        </Flex>
```

i need the FlexProps here, and i need it to have the "ref" property.

packages/react-components/src/layout/src/Flex.tsx
patricklafrance: ??

packages/react-components/src/box/src/Box.tsx
patricklafrance: As I wrote in another PR I would prefer if we could somehow do this without any custom code, only native TS. It's too abstract for me.

packages/react-components/src/box/src/Box.tsx
patricklafrance: This `slot` prop is basically the only reason the Box component exist.

We remove it to ensure it's not rendered on the DOM element but it's pickedup by parent Component who support dynamic slots with (useSlots).

packages/react-components/src/layout/src/Flex.tsx
patricklafrance: sure leave it there.

packages/react-components/src/layout/src/Stack.tsx
```diff
@@ -52,12 +57,12 @@ export function InnerStack({
...
     forwardedRef,
     ...rest
-}) {
+}: InnerStackProps): ReactElement {
     const alignProps = useFlexAlignment("vertical", align, verticalAlign);
 
     return (
         <Flex
-            {...mergeProps(
+            {...mergeProps<Partial<FlexProps>[]>(
```
patricklafrance, line 65: Do you really have to specify a type to mergeProps? Things will get ugly real quick if we must do this everywhere.

packages/react-components/src/text/src/Text.tsx
```diff
@@ -1,36 +1,39 @@
...
-    size: oneOf(["xs", "sm", "md", "lg", "xl", "2xl", "3xl", "4xl", "5xl", "inherit"]),
+    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
     /**
      * An HTML element type or a custom React element type to render as.
      */
-    as: oneOfType([string, elementType]),
+    as?: ElementType;
     /**
-     * Default slot override.
+     * [Slot](?path=/docs/getting-started-slots--page) to render into.
```
patricklafrance, line 21: This is not what is it, please rever to the original comment

packages/react-components/src/text/src/Text.tsx
```diff
@@ -1,36 +1,39 @@
...
      */
-    slot: string,
+    slot?: string;
     /**
-     * @ignore
+     * React children
```
patricklafrance, line 25: We only document children when it's required, otherwise it  shouldn't be in the docs.

packages/react-components/src/layout/src/adapters.ts
```diff
@@ -0,0 +1,26 @@
+type Orientation = "horizontal" | "vertical";
+type Direction = "row" | "column"
+type Alignment = "start" | "end" | "center";
+
+const DIRECTION: Record<Orientation, Direction> = {
```
patricklafrance, line 5: Seems like a good time to rename DIRECTION to Direction

packages/react-components/src/layout/src/Stack.tsx
```diff
@@ -1,48 +1,53 @@
...
      * An HTML element type or a custom React element type to render as.
      */
-    as: oneOfType([string, elementType]),
+    as?: ElementType;
     /**
      * React children.
      */
-    children: any.isRequired
-};
+    children?: ReactNode
```
patricklafrance, line 43: It's was required and now it's not

packages/react-components/src/layout/src/Inline.tsx
```diff
@@ -1,48 +1,53 @@
...
+    */
+    as?: ElementType;
     /**
-     * React children.
-     */
-    children: any.isRequired
-};
+     * React children
+    */
+    children?: ReactNode;
```
patricklafrance, line 43: It was required and now it's not

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,71 @@
...
      */
-    as: oneOfType([string, elementType]),
+    as?: ElementType;
     /**
-     * @ignore
+     * React children
      */
-    children: any.isRequired
-};
+    children?: ReactNode;
```
patricklafrance, line 61: It was required and now it's not

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,48 @@
...
+    */
+    as?: ElementType;
+    /**
+     * [Slot](?path=/docs/getting-started-slots--page) to render into.
+     */
+    slot?: string;
+    /**
+     * React children
+    */
+    children?: ReactNode;
```
patricklafrance, line 16: Don't need. We don't add every native React props only those that are required.

patricklafrance (changes requested):
View comments

packages/react-components/src/box/src/Box.tsx
alexasselin008: I'll remove it from the interface, and tweak the omitProps method the not only expect keyof props, but also any string.

packages/react-components/src/text/src/Text.tsx
patricklafrance: Sorry you are right it's better.

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -194,8 +158,8 @@ export function InnerFlex({
     );
 }
 
-InnerFlex.propTypes = propTypes;
-
-export const Flex = forwardRef((props, ref) => (
+export const Flex = forwardOrbitRef<InnerFlexProps>((props, ref) => (
```
alexasselin008, line 161: Le component à pas besoin de displayName comme les autres?

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,46 @@
...
+        >
+            {children}
+        </Wrapper>
+    );
+}
+export const Box = forwardRef<InnerBoxProps>((props, ref) => (
+    <InnerBox {...props} forwardedRef={ref} />
+));
+
+export type BoxProps = ComponentProps<typeof Box>
```
alexasselin008, line 44: I feel like this might be the only line in component codes you might dislike. 

The idea here is that some component might require accès to a component props, either to extend them or to type an object that will be passed as props.

There is two choice available, either we export the props like this all the time, or if someone needs the box component's properties, he uses `ComponentProps<typeof Box>` every time.

I'd rather export the type by default everywhere

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,46 @@
...
+            {...rest}
+            ref={forwardedRef}
+        >
+            {children}
+        </Wrapper>
+    );
+}
+export const Box = forwardRef<InnerBoxProps>((props, ref) => (
+    <InnerBox {...props} forwardedRef={ref} />
+));
```
alexasselin008, line 42: Using the forwardRef from Orbit, we get code that is really similar to regular react. 

Here are some exemple with the types generated

 **Generic use case : We don't specify a type**

```
export const Box = forwardRef<InnerBoxProps>((props, ref) => (
    <InnerBox {...props} forwardedRef={ref} />
));


export type BoxProps = ComponentProps<typeof Box>;

type OnClick = BoxProps["onClick"]; // (event: MouseEvent<Element, MouseEvent>) => void
type UnknownProperty = BoxProps["aaaa"]; // any, we allow extra properties, but we don't type them
type Slot = BoxProps["slot"]; //  string
type Ref = BoxProps["ref"]; //  ((instance: InnerBoxProps<Element>) => void) | RefObject<InnerBoxProps<Element>>

```
So we see that when we don't specifiy a type as the 2nd generic parameter in forward Ref, we extends every html properties, using with an Element type.


**Specific properties**
```
export const BoxAsInput = forwardRef<InnerBoxProps, "input">((props, ref) => {
    return <InnerBox {...props} forwardedRef={ref} />;
});

export type BoxProps = ComponentProps<typeof BoxInput>;

type OnClick = BoxProps["onClick"]; // (event: MouseEvent<HTMLInputElement, MouseEvent>) => void
type UnknownProperty = BoxProps["aaaa"]; // any, we allow extra properties, but we don't type them
type Slot = BoxProps["slot"]; //  string
type Ref = BoxProps["ref"]; //  ((instance: HTMLInputElement) => void) | RefObject<HTMLInputElement>
type OnChange = BoxProps["onChange"]; // (event: ChangeEvent<HTMLInputElement>) => void

```

**Custom Components**

```

const MyRouterLink = React.forwardRef<HTMLLinkElement, { allo: boolean }>((props, ref) => {
    return <>{props.allo}</>;
});

export const Box = forwardRef<InnerBoxProps, typeof MyRouterLink>((props, ref) => (
    <InnerBox {...props} forwardedRef={ref} />
));
export type BoxProps = ComponentProps<typeof Box >;

type OnClick = BoxProps["onClick"]; //any -> since we provided a component, we don't add all html props
type UnknownProperty = BoxProps["aaaa"]; // any, we allow extra properties, but we don't type them
type Slot = BoxProps["slot"]; //  string
type Ref = BoxProps["ref"]; // ((instance: HTMLLinkElement) => void) | RefObject<HTMLLinkElement>
type CustomProps = BoxProps["allo"]; //  boolean
```

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,70 @@
...
      */
-    wrapChildren: bool,
+    wrapChildren?: boolean;
     /**
      * An HTML element type or a custom React element type to render as.
      */
-    as: oneOfType([string, elementType]),
+    as?: ElementType;
     /**
-     * @ignore
```
alexasselin008, line 96: I would document children properties when they are required, what do you think?

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -179,9 +142,10 @@ export function InnerFlex({
...
                         flexDirection: direction && `${direction}${reverse ? "-reverse" : ""}`,
-                        alignContent: alignContent,
-                        alignItems: alignItems,
-                        justifyContent: justifyContent,
+                        // Normalize values until Chrome support `start` & `end`, https://developer.mozilla.org/en-US/docs/Web/CSS/align-items.
+                        alignContent: alignContent && alignContent.replace("start", "flex-start").replace("end", "flex-end"),
+                        alignItems: alignItems && alignItems.replace("start", "flex-start").replace("end", "flex-end"),
+                        justifyContent: justifyContent && justifyContent.replace("start", "flex-start").replace("end", "flex-end"),
                         flexWrap: !isNil(wrap) ? "wrap" : undefined,
                         "--o-ui-gap": gap && (isString(gap) ? gap : `var(${Spacing[(gap) - 1]})`)
```
alexasselin008, line 150: At the moment, there is an error on this line. 
```
Type '{ flexDirection: Property.FlexDirection; alignContent: string; alignItems: string; justifyContent: string; flexWrap: "wrap"; "--o-ui-gap": string; }' is not assignable to type 'CSSProperties'.
  Object literal may only specify known properties, and '"--o-ui-gap"' does not exist in type 'CSSProperties'.ts(2322)
```

This is due to the fact that --o-ui-gap is a custom property. After some research, there is multiple ways to handle this scenario : 

**Allow custom property for this object only :** 
Using type assertion, you can allow this property on this object. The downside is that we have to do this synthax everytime we have one of these scenarios : 
```
    ["--o-ui-gap" as any]: gap && (isString(gap) ? gap : `var(${Spacing[(gap) - 1]})`)
```
Source : https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors 

**The second option is Augment the CSSProperties to allow this property**
Using [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces), we can allow this property everywhere

This is painful to write, but we ensure type validation

So at the root of Flex.tsx, we add the following lines : 
```
declare module 'csstype' {
  interface Properties {
    // Add a CSS Custom Property
    '--o-ui-gap'?:  any; // or the right typing.
  }
}
```
Source : https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors

**The last solution, wich might be the one we want.** 
We use module augmentation to allow any extra properties.

somewhere in the code, we add this : 
```
declare module 'csstype' {
  interface Properties {
     // ...or allow any other property
    [index: string]: any;
  }
}
```
Allowing any extra parameters to be passed. The cons are that we don't type those properties, but we don't have to add extra synthax in the components

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -141,7 +109,7 @@ function useIsGapSupported(noGap) {
 }
 
 export function InnerFlex({
-    direction,
+    direction = "row",
```
alexasselin008, line 112: I added a default value, instead of doing the || "row" where we needed it. It also fixes an issue i had with the flexdirection property

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,46 @@
...
+    */
+    as?: ElementType;
+    /**
+     * [Slot](?path=/docs/getting-started-slots--page) to render into.
+     */
+    slot?: string;
+    /**
+     * React children
+    */
+    children?: ReactNode;
```
patricklafrance, line 16: Add @ignore since as discussed we only want to document required children for now.

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,70 @@
...
-        "space-around",
-        "space-evenly",
-        "stretch",
-        "baseline",
-        "first baseline",
-        "last baseline",
-        "safe center",
-        "unsafe center"
-    ]),
+    alignContent?: "start" | "end" | "center" | "space-between" | "space-around" | "space-evenly" | "stretch" | "baseline" | "first baseline" | "last baseline" | "safe center" | "unsafe center";
```
patricklafrance, line 25: Please split on multiple lines if possible

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,70 @@
...
-        "stretch",
-        "self-start",
-        "self-end",
-        "baseline",
-        "first baseline",
-        "last baseline",
-        "safe center",
-        "unsafe center"
-    ]),
+    alignItems?: "start" | "end" | "center" | "stretch" | "self-start" | "self-end" | "baseline" | "first baseline" | "last baseline" | "safe center" | "unsafe center";
```
patricklafrance, line 29: Same

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,70 @@
...
-        "space-around",
-        "space-evenly",
-        "stretch",
-        "baseline",
-        "first baseline",
-        "last baseline",
-        "safe center",
-        "unsafe center"
-    ]),
+    justifyContent?: "start" | "end" | "center" | "left" | "right" | "space-between" | "space-around" | "space-evenly" | "stretch" | "baseline" | "first baseline" | "last baseline" | "safe center" | "unsafe center"
```
patricklafrance, line 33: Same

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,102 +1,70 @@
...
     /**
      * Whether to wrap children in a `div` element.
      */
-    wrapChildren: bool,
+    wrapChildren?: boolean;
     /**
      * An HTML element type or a custom React element type to render as.
      */
-    as: oneOfType([string, elementType]),
+    as?: ElementType;
```
patricklafrance, line 57: can't be a "string" anymore? like "div"

packages/react-components/src/layout/src/Flex.tsx
patricklafrance: Of course!

packages/react-components/src/layout/src/Flex.tsx
patricklafrance: I am not sure but I might haven't set a default value on purpose to because when it's not provided I might have wanted to have undefined here:

`flexDirection: direction && `${direction}${reverse ? "-reverse" : ""}`,`

patricklafrance: There were some tricky cases with this code.

packages/react-components/src/layout/src/Inline.tsx
```diff
@@ -1,48 +1,51 @@
...
      */
-    fluid: bool,
+    fluid?: boolean;
     /**
-     * An HTML element type or a custom React element type to render as.
-     */
-    as: oneOfType([string, elementType]),
+    * An HTML element type or a custom React element type to render as.
+    */
+    as?: ElementType;
```
patricklafrance, line 39: No string?

packages/react-components/src/box/src/Box.tsx
```diff
@@ -0,0 +1,46 @@
+import { ComponentProps, ElementType, ForwardedRef, ReactElement, ReactNode } from "react";
+import { forwardRef, omitProps } from "../../shared";
+
+export interface InnerBoxProps {
```
patricklafrance, line 4: I thought all our props interface were supposed to inherit HtmlElement or something more specific like HTMLLinkElement for special cases?

packages/react-components/src/box/src/Box.tsx
patricklafrance: So the following would also work with full typing available? (giving we do exposed BoxProps as you suggested)

```
export const ParagraphBox = forwardRef<BoxProps, "p">((props, ref) => (
    <Box {...props} ref={ref} />
));
```

The best way to do this would be the following but both syntax would be supported right?

```
<Box as="p">Blabla</Box>
```

patricklafrance:
Pretty sure we discussed about introducing an `OrbitComponent` interface? Even is it wouldn't work in the docs for now.

packages/react-components/src/text/src/Text.tsx
```diff
@@ -1,36 +1,37 @@
...
 
-////////
-
-const propTypes = {
+export interface InnerTextProps {
     /**
      * A text can vary in size.
      */
-    size: oneOf(["xs", "sm", "md", "lg", "xl", "2xl", "3xl", "4xl", "5xl", "inherit"]),
+    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
```
patricklafrance, line 15: "2xs" have been added here, it's not supported by a Text component

PR #454, opened 2021-03-02
Conversion of the Icon folder to typescript

alexasselin008:


packages/react-components/src/icons/src/EmbeddedIcon.tsx
```diff
@@ -0,0 +1,38 @@
+import { Children, ReactElement, ReactNode } from "react";
+import { augmentElement, createSizeAdapter } from "../../shared";
+
+export interface EmbeddedIconProps {
+    /**
```
alexasselin008, line 5: I added some documentation here, even if it's not used by storybook. It's now available through intellisense

packages/react-components/src/icons/src/EmbeddedIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+     */
+    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "inherit";
+    /**
+     * React children.
+     */
+    children: ReactNode;
+    /**
+     * Allow any extra props
+     */
+    [key: string]: any
```
alexasselin008, line 16: first time we have to specify this, since it's the only component so far that is not using the forwardRef property

alexasselin008: It's all good with me, but i wanted to show you

packages/react-components/src/icons/src/index.ts
```diff
@@ -6,14 +7,14 @@ export * from "./MultiVariantIcon";
...
-function createOrbitIcon(source, displayName) {
+function createOrbitIcon(source: ElementType, displayName: string): ReturnType<typeof Icon.create> {
     const Component = Icon.create(source);
     Component.displayName = displayName;
 
     return Component;
 }
 
-function createOrbitMultiVariantIcon(source24, source32, displayName) {
+function createOrbitMultiVariantIcon(source24: ElementType, source32: ElementType, displayName: string): ReturnType<typeof MultiVariantIcon.create> {
```
alexasselin008, line 17: i prefer the `ReturnType<typeof MultiVariantIcon.create>` synthax over `OrbitComponent<"svg", CreatedMultiVariantIconProps>`. What do you think?

packages/react-components/src/icons/src/MultiVariantIcon.tsx
```diff
@@ -0,0 +1,81 @@
...
+    );
+}) as InnerMultiVariantIconComponentWithFactory;
+
+export type MultiVariantIconComponentWithFactory = OrbitComponent<"svg", InnerMultiVariantIconProps> & {
+    create: MultiVariantIconFactory
+}
+
+export const MultiVariantIcon = slot("icon", forwardRef<InnerMultiVariantIconProps, "svg">((props, ref) => (
+    <InnerMultiVariantIcon {...props} forwardedRef={ref} />
+))) as MultiVariantIconComponentWithFactory;
```
alexasselin008, line 60: We have to cast since we don't feed the "create" property right away.

typings/named-asset-import.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module '*.svg' {
```
alexasselin008, line 1: copied from CRA

packages/react-components/src/icons/src/EmbeddedIcon.tsx
```diff
@@ -0,0 +1,38 @@
+import { Children, ReactElement, ReactNode } from "react";
+import { augmentElement, createSizeAdapter } from "../../shared";
+
+export interface EmbeddedIconProps {
+    /**
+     * An icon can vary in size.
+     */
+    size?: "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "inherit";
```
patricklafrance, line 8: I might be confused but I think you told me you need to use the () syntax for the enum to render correctly in SB

packages/react-components/src/icons/src/EmbeddedIcon.tsx
```diff
@@ -0,0 +1,38 @@
...
+     */
+    children: ReactNode;
+    /**
+     * Allow any extra props
+     */
+    [key: string]: any
+}
+
+export const embeddedIconSize = createSizeAdapter({
+    "2xs": "2xs",
```
patricklafrance, line 20: now that we export a `Size` we should use it here

packages/react-components/src/icons/src/Icon.tsx
```diff
@@ -60,27 +73,31 @@ export function InnerIcon(props) {
...
 
 Icon.displayName = "Icon";
 
+export type IconProps = ComponentProps<typeof Icon>;
+
 ////////
 
-function createIconFactory(type) {
-    return slot("icon", forwardRef((props, ref) =>
+const createIconFactory: IconFactory = type => {
```
patricklafrance, line 92: just wondering, is it a must to convert the function declaration to an arrow function here to use TS?

packages/react-components/src/icons/src/Icon.tsx
```diff
@@ -49,7 +62,7 @@ export function InnerIcon(props) {
                         "o-ui-icon",
                         disabled && "disabled",
                         normalizeSize(size),
-                        color ? `color-${color}`: ""
+                        color ? `color-${color}` : ""
```
patricklafrance, line 65: I don't think this is used. I think you added this in the dark mode branch @fraincs can you confirm?

packages/react-components/src/icons/src/Icon.tsx
```diff
@@ -1,30 +1,43 @@
...
      */
-    slot: string
-};
-export function InnerIcon(props) {
+    slot?: string;
+}
+
+export interface InnerIconProps extends CreatedIconProps {
+    /**
+     * An icon as a React component.
```
patricklafrance, line 25: I would document this as "the concrete icon type" something like this.

```
import { Icon } from "@orbit-ui/react-components";
import { ReactComponent as Azure32 } from "./icon-azure-32.svg";

<Icon type={Azure32} />
```

packages/react-components/src/icons/src/Icon.tsx
```diff
@@ -60,27 +73,31 @@ export function InnerIcon(props) {
             )}
         />
     );
-}
+}) as InnerIconComponentWithFactory;
 
-InnerIcon.propTypes = propTypes;
+type IconComponentWithFactory = OrbitComponent<"svg", InnerIconProps> & {
+    create: IconFactory
+}
```
patricklafrance, line 80: I would replace all this `Icon.create` thing by a `createIcon` function, what do you think?

Would also requires to update the docs pages "?path=/docs/icon--default-story"
- update the usage
- update the imports to add `createIcon`

patricklafrance: Oh we would also need a `createMultivariantIcon`

packages/react-components/src/icons/src/IconList.tsx
```diff
@@ -1,22 +1,29 @@
...
-import { any, elementType, oneOfType, string } from "prop-types";
-import { slot } from "../../shared";
+import { forwardRef, slot } from "../../shared";
 
-const propTypes = {
+interface InnerIconListProps {
+    /**
+     * Whether or not the IconList is disabled.
+     */
+    disabled?: boolean;
```
patricklafrance, line 9: ignore this in the docs since it's a standard html prop

packages/react-components/src/icons/src/MultiVariantIcon.tsx
```diff
@@ -0,0 +1,81 @@
...
+export type MultiVariantIconFactory = (type24: ElementType, type32: ElementType) => OrbitComponent<"svg", CreatedMultiVariantIconProps>;
+
+export type InnerMultiVariantIconComponentWithFactory = ((props: InnerMultiVariantIconProps) => ReactElement) & {
+    create: MultiVariantIconFactory
+}
+
+export const InnerMultiVariantIcon = (({ type24: Component24, type32: Component32, size, forwardedRef, ...rest }: InnerMultiVariantIconProps): ReactElement => {
+    let type = Component24;
+
+    if (size === "lg" || size === "xl") {
```
patricklafrance, line 40: We can use `Size` here

packages/react-components/src/icons/src/MultiVariantIcon.tsx
```diff
@@ -0,0 +1,81 @@
+import { ComponentProps, ElementType, ForwardedRef, ReactElement } from "react";
+import { Icon } from "./Icon";
+import { OrbitComponent, forwardRef, slot } from "../../shared";
+
+export interface CreatedMultiVariantIconProps {
```
patricklafrance, line 5: An instance of a React type is called a React element, therefore we could instead call this `MultiVariantIconElement` ?

packages/react-components/src/icons/src/Icon.tsx
patricklafrance: In fact, since we use "as" everywhere, should this be renamed to "as" ?

packages/react-components/src/icons/src/index.ts
```diff
@@ -6,14 +7,14 @@ export * from "./MultiVariantIcon";
 export * from "./IconList";
 export * from "./EmbeddedIcon";
 
-function createOrbitIcon(source, displayName) {
+function createOrbitIcon(source: ElementType, displayName: string): ReturnType<typeof Icon.create> {
```
patricklafrance, line 10: Maybe rename to "as" if we do it in Icon.create?

typings/named-asset-import.d.ts
patricklafrance: Please rename the file to use camelcase as we do for other files that are not React components.

patricklafrance: Is there a standard naming convention for TS typings?

patricklafrance (approved):

typings/named-asset-import.d.ts
alexasselin008: The only standard i could find is that typescript declaration files are prefixed by lib.
```
TypeScript names these declaration files with the pattern lib.[something].d.ts. If you navigate into a file with that name, you can know that you're dealing with some built-in part of the platform, not user code.
```
Reference : https://microsoft.github.io/TypeScript-New-Handbook/chapters/type-declarations/

React seems to type their declaration files in lowercase with -
```
jsx-dev-runtime.d.ts
jsx-runtime.d.ts
```

packages/react-components/src/icons/src/EmbeddedIcon.tsx
alexasselin008: I was mistaken

packages/react-components/src/icons/src/EmbeddedIcon.tsx
alexasselin008: Size is a type union, not an enum. So if we provide a string that is not allowed in Size, typescript will put it in error.

packages/react-components/src/icons/src/Icon.tsx
alexasselin008: Wouldn't work with MultiVariantIcons

packages/react-components/src/icons/src/MultiVariantIcon.tsx
alexasselin008: This is not a react type, nor a react element. It's an interface for the MultiVariantIconElement's props

alexasselin008: i renamed it to MultiVariantIconElementProps

patricklafrance (approved):

typings/namedAssetImport.d.ts
```diff
@@ -0,0 +1,10 @@
+declare module '*.svg' {
```
patricklafrance, line 1: can we use double quote here? or the single quote are mandatory?

packages/react-components/src/icons/src/Icon.tsx
fraincs: This can be removed

PR #455, opened 2021-03-02
Form,Image,Lozenge,Paragraph,ThemeProvider,Toolbar,Transition converted to TypeScript

alexasselin008:


packages/react-components/src/form/src/Form.tsx
```diff
@@ -1,27 +1,34 @@
...
+export interface InnerFormProps {
     /**
      * Whether or not the form take up the width of its container.
      */
-    fluid: bool,
+    fluid?: boolean;
+    /**
+     * Whether or not the Form is disabled.
+     */
+    disabled?: boolean;
```
patricklafrance, line 16: ignore since tit's a standard html prop

packages/react-components/src/form/src/FormContext.ts
```diff
@@ -1,9 +1,14 @@
 import { createContext, useContext } from "react";
 import { isNil } from "lodash";
 
-export const FormContext = createContext(null);
+export interface FormContextProps {
```
patricklafrance, line 4: If you use this name for the actual Context instance how would you call an interface for `useFormContext` if it eventually receive args?

patricklafrance: Well I guess you could call it "UseFormProps".

That being said, suffixing a type describing an actual context with "props" seems wrong.

I would have call it "FormContext" but we obviously can't.

packages/react-components/src/lozenge/src/Lozenge.tsx
```diff
@@ -1,30 +1,33 @@
...
-    size: oneOf(["sm", "md"]),
+    size?: "sm" | "md";
     /**
      * An HTML element type or a custom React element type to render as.
      */
-    as: oneOfType([string, elementType]),
+    as?: ElementType;
     /**
-     * @ignore
+    * @ignore
```
patricklafrance, line 23: required children shouldb't be ignored, was a mistake in original code.

patricklafrance (approved):

packages/react-components/src/form/src/FormContext.ts
alexasselin008: i'll do this in the next PR

PR #456, opened 2021-03-02
Heading, Placeholder,Link,Divider and Counter converted to TypeScript

alexasselin008:


packages/react-components/src/link/src/useLink.ts
```diff
@@ -1,6 +1,32 @@
-import { cssModule, mergeClasses, normalizeSize, useAutoFocus, useMergedRefs } from "../../shared";
+import { ForwardedRef } from "react";
+import { MergedRef, Size, cssModule, mergeClasses, normalizeSize, useAutoFocus, useMergedRefs } from "../../shared";
 import { isNumber } from "lodash";
 
+export interface UseLinkProps {
+    cssModule?: string;
```
alexasselin008, line 6: can't know at the moment if they are required props here. Should i let them all optional except forwardRef?

packages/react-components/src/link/src/useLink.ts
```diff
@@ -1,6 +1,32 @@
-import { cssModule, mergeClasses, normalizeSize, useAutoFocus, useMergedRefs } from "../../shared";
+import { ForwardedRef } from "react";
+import { MergedRef, Size, cssModule, mergeClasses, normalizeSize, useAutoFocus, useMergedRefs } from "../../shared";
 import { isNumber } from "lodash";
 
+export interface UseLinkProps {
```
alexasselin008, line 5: Should this be the LinkProps?

packages/react-components/src/link/src/TextLink.tsx
```diff
@@ -1,61 +1,64 @@
...
     /**
      * A link can vary in size.
      */
-    size: oneOf(["sm", "md", "inherit"]),
+    size?: "sm" | "md" | "inherit";
     /**
      * Whether or not the link is disabled.
      */
-    disabled: bool,
+    disabled?: boolean;
```
patricklafrance, line 46: ignore disabled, it was a mistake in original code

packages/react-components/src/link/src/useLink.ts
patricklafrance: yeah leave everything optionnal and i'll update them sometime

packages/react-components/src/link/src/useLink.ts
patricklafrance: Not exactly, It will of course be a subset of the props but it could differ at some point

packages/react-components/src/link/src/useLink.ts
```diff
@@ -1,6 +1,32 @@
...
+    active?: boolean;
+    focus?: boolean;
+    hover?: boolean;
+    visited?: boolean;
+    target?: string;
+    rel?: string;
+    forwardedRef: ForwardedRef<HTMLElement>;
+}
+
+export interface UseLinkApi {
```
patricklafrance, line 23: If we need these, I suggest we use Chakra "Return" naming convention. Could also use "Output" instead of "Return"

packages/react-components/src/shared/src/useMergedRefs.ts
```diff
@@ -11,7 +11,7 @@ export function assignRef<T>(ref: ForwardedRef<T>, node: T): void {
     }
 }
 
-type MergedRef<T> = ((instance: T | null) => void) & MutableRefObject<T | null>;
+export type MergedRef<T> = ((instance: T | null) => void) & MutableRefObject<T | null>;
```
patricklafrance, line 14: Can use the RefCallback type instead? The idea of mergeRefs is really to mimick a ref callback behavior the same signature

PR #458, opened 2021-03-03
Remove all explicit type anotation and renamed all context interface with suffix Type

alexasselin008:


packages/react-components/src/form/src/useFormButton.ts
```diff
@@ -1,7 +1,7 @@
-import { FormContextProps, useFormContext } from "./FormContext";
+import { FormContextType, useFormContext } from "./FormContext";
 import { omitProps } from "../../shared";
 
-export type FormButtonProps = Omit<FormContextProps, "fluid">;
+export type FormButtonProps = Omit<FormContextType, "fluid">;
```
patricklafrance, line 4: Could rename for UseFormButtonReturn please

patricklafrance (approved):

PR #459, opened 2021-03-05
Alert, Badge, Button and Field converted to typescript

alexasselin008:


packages/react-components/src/shared/src/size.ts
```diff
@@ -1,15 +1,15 @@
 export type Size = "2xs" | "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "3xl" | "4xl" | "5xl" | "inherit";
 
-export function normalizeSize(size?: Size) {
+export function normalizeSize<T extends Size>(size?: T) {
```
alexasselin008, line 3: doing this creates better adapters.

for this example
```
const adapter = createSizeAdapter({
"sm": "md",
"md": "lg"
})

const res = adapter("md");
```

before : 
res was of type Size

now: 
res is "md" | "lg"

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -52,15 +52,18 @@ export function useRawSlots(children: ReactNode, slots: string[]) {
...
     _: {
         defaultWrapper?: ComponentType;
         required?: string[];
-    },
-    [x: string]: any;
+    }
 }
 
-export function getSlots(children: ReactNode, { _ = {}, ...slots }: Slots) {
+type SlotElements<T extends RequiredDefaultSlot> = {
```
alexasselin008, line 62: given
```
    const slots} = useSlots(children, useMemo(() => ({
        _: {
            defaultWrapper: Content
        },
        icon: {
            className: "o-ui-alert-icon"
        },
        content: {
            className: "o-ui-alert-content",
            as: AlertContent
        },
        button: {
            variant: "ghost",
            color: "inherit",
            condensed: true,
            size: "sm",
            className: "o-ui-alert-action"
        }
    }), []));
```

before : 
slot was type with [key:string]: any

now : 
slot is typed with 
```
{
    icon?:ReactElement;
    content?:ReactElement;
    button?: ReactElement;
   [key:string]: ReactElement
}
```
Wich means we give suggestion for autocomplete  while destructuring + allow any other slot that would be extracted (to support the defaultWrapper case)

packages/react-components/src/shared/src/forwardRef.ts
```diff
@@ -1,12 +1,17 @@
 // Idea and naming based off Chakra UI. https://github.com/chakra-ui/chakra-ui/blob/main/packages/system/src/system.types.tsx
 import { ComponentProps, ElementRef, ElementType, ForwardRefExoticComponent, ForwardRefRenderFunction, HTMLProps, RefAttributes, WeakValidationMap, forwardRef as reactForwardRef } from "react";
 
-type UnwrapElementType<T> = T extends ElementType ? ElementRef<T> : T;
+type AsRef<T> = T extends ElementType ? ElementRef<T> : T;
 
-type PropsOf<T> =
-    T extends ElementType ? HTMLProps<UnwrapElementType<T>> & ComponentProps<T> & RefAttributes<UnwrapElementType<T>> :
+type PropsWithoutChildren<P> =
```
alexasselin008, line 6: In some cases, we didn't want to allow the children props, since we don't allow it, but the as type was bringing it back

packages/react-components/src/alert/src/Alert.tsx
```diff
@@ -157,15 +164,15 @@ const variations = [
     { tone: "success", icon: <CheckIcon /> },
     { tone: "warning", icon: <WarningIcon /> },
     { tone: "critical", icon: <InfoIcon /> }
-];
+] as const;
```
patricklafrance, line 167: Souldn't it be a Record<string, Icon> ?

packages/react-components/src/alert/src/Alert.tsx
```diff
@@ -191,42 +198,45 @@ const [
...
-    // eslint-disable-next-line react/no-unused-prop-types
-    children: any.isRequired
-};
+    children: ReactNode
+}
+
+// Dummy component for documentation purpose.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function AlertTemplate(_props: AlertTemplateProps) {
+    // When returning null, react-docgen doesn't ignore the component.
```
patricklafrance, line 225: oops made a mistake in my comment..

should be: // When returning null, react-docgen ignore the component.

packages/react-components/src/button/src/Button.tsx
```diff
@@ -1,82 +1,97 @@
...
-     * Called when the button is click.
-     * @param {SyntheticEvent} event - React's original SyntheticEvent.
-     * @returns {void}
-     */
-    onClick: func,
+    * Called when the button is click.
+    * @param {SyntheticEvent} event - React's original SyntheticEvent.
+    * @returns {void}
+    */
+    onClick?: (event: SyntheticEvent) => void
```
patricklafrance, line 58: Do we want to be more specific for handlers like a button `onClick` ? We could use `MouseEvent` here.

packages/react-components/src/button/src/ButtonGroup.tsx
```diff
@@ -43,9 +46,9 @@ const Gap = {
         "sm": 2,
         "md": 3
     }
-};
+} as const;
```
patricklafrance, line 49: I guess we are using const here instead of a nesting of Records because it's easier and no one cares if we go with something more precise?

packages/react-components/src/button/src/IconButton.tsx
```diff
@@ -1,75 +1,98 @@
...
+     */
+    fluid?: boolean;
+    /**
+     * @ignore
+     */
+    active?: boolean;
+    /**
+     * @ignore
+     */
+    focus?: boolean;
```
patricklafrance, line 76: Since `active`, `focus`, `hover` will never be documented, we could make a base interface for them?

Something like `InteractionStatesProps`

packages/react-components/src/button/src/useButton.ts
```diff
@@ -1,6 +1,23 @@
-import { cssModule, mergeClasses, normalizeSize, useAutoFocus, useMergedRefs } from "../../shared";
+import { ForwardedRef } from "react";
+import { Size, cssModule, mergeClasses, normalizeSize, useAutoFocus, useMergedRefs } from "../../shared";
 import { isNumber } from "lodash";
 
+export interface UseButtonProps {
```
patricklafrance, line 5: Hallelujah!

packages/react-components/src/button/src/useToggleButton.ts
```diff
@@ -1,7 +1,22 @@
...
+    checked?: boolean;
+    defaultChecked?: boolean;
+    value?: string | number;
+    onChange?(event: SyntheticEvent, isChecked: boolean): void;
+    onCheck?(event: SyntheticEvent, value: string | number): void;
+    forwardedRef?: ForwardedRef<any>;
+}
+
+export function useToggleButton<
+    Shape extends "pill" | "rounded" | "circular"
```
patricklafrance, line 18: How come you have to do this instead of:

```
shape?: "pill" | "rounded" | "circular";
```

packages/react-components/src/field/src/Field.tsx
```diff
@@ -2,45 +2,48 @@ import "./Field.css";
...
     /**
      * A field id.
      */
-    id: string,
+    id?: string,
     /**
      * Whether the field should display as "valid" or "invalid".
      */
-    validationState: oneOf(["valid", "invalid"]),
+    validationState?: "valid" | "invalid";
```
patricklafrance, line 19: God, I hope react-docgen support remote types soon.

packages/react-components/src/field/src/FieldContext.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+interface ClearFieldContextProps {
+    children?: ReactNode;
+}
+
+export function ClearFieldContext({ children }: ClearFieldContextProps) {
+    return (
+        <FieldContext.Provider value={null} >
+            { children}
```
patricklafrance, line 157: Remove white space after {

packages/react-components/src/field/src/FieldContext.tsx
```diff
@@ -0,0 +1,160 @@
...
+        size,
+        className: inputClassName,
+        "aria-labelledby": !isNil(labelId) ? labelId : undefined,
+        "aria-describedby": !isNil(messageId) ? messageId : undefined
+    };
+
+    return [props || {}, isInField];
+}
+
+export type UseFieldMessageProps = [{
```
patricklafrance, line 120: Should be `UseFieldMessageReturn` I guess? Otherwise if we add props to useFieldMessageProps
 how would we name the interface?

packages/react-components/src/field/src/FieldContext.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    return [{}, false];
+}
+
+export interface UseFieldLabelProps {
+    as?: ElementType;
+}
+
+export type UseFieldLabelReturn = [{
```
patricklafrance, line 52: UseFieldLabelPropsReturn

packages/react-components/src/field/src/FieldContext.tsx
```diff
@@ -0,0 +1,160 @@
+import { ElementType, ReactNode, createContext, useContext } from "react";
+import { cssModule, normalizeSize } from "../../shared";
+import { isNil } from "lodash";
+
+export interface CommonFieldContextType {
```
patricklafrance, line 5: Do you have to export a base interface like this one?

packages/react-components/src/field/src/FieldContext.tsx
```diff
@@ -0,0 +1,160 @@
...
+    isGroupField?: boolean;
+}
+
+export const FieldContext = createContext<FieldContextType>(null);
+
+export function useFieldContext(): [UseFieldContextType, boolean] {
+    const context = useContext(FieldContext);
+
+    if (!isNil(context)) {
+        const { isGroup: isGroupField = false, ...rest } = context;
```
patricklafrance, line 35: Oh well my renaming was cute without TS. Now it seems stupid haha. Maybe we should normalize this to `isGroup` ?

packages/react-components/src/shared/src/forwardRef.ts
patricklafrance: Looks like a good comment to add in the code instead of the review ;)

packages/react-components/src/shared/src/omitProps.ts
```diff
@@ -1,4 +1,4 @@
-export function omitProps<TObject extends Record<string, any>, TKey extends keyof TObject>(obj: TObject, omit: TKey[]) {
+export function omitProps<TObject extends Record<string, any>, TKey extends string>(obj: TObject, omit: TKey[]) {
```
patricklafrance, line 1: Could you instead do: 

```
export function omitProps<TObject extends Record<string, any>>(obj: TObject, omit: string[]) {

```

packages/react-components/src/shared/src/forwardRef.ts
patricklafrance: Is it really without? Seems to me more like `PropsWithOptionalChildren`

packages/react-components/src/shared/src/checkable.ts
```diff
@@ -1,9 +1,9 @@
-import { createContext, useContext } from "react";
+import { SyntheticEvent, createContext, useContext } from "react";
 import { isNil } from "lodash";
 
 interface CheckableContextType {
-    checkedValue?: boolean;
-    [x: string]: any;
+    checkedValue?: boolean | string | number;
+    onCheck?(event: SyntheticEvent, newValue: boolean | string | number): void;
```
patricklafrance, line 6: Should it be a `ChangeEvent` ?

packages/react-components/src/shared/src/augmentElement.tsx
```diff
@@ -8,8 +8,8 @@ export function augmentElement(element: ReactElement & RefAttributes<any>, newPr
     return cloneElement(element, augmentedProps);
 }
 
-export function createEmbeddableAdapter(sizeAdapter: SizeAdapter) {
-    return <Props extends Record<string, any>>(element: ReactElement<Props, any>, { size, ...props }: Props) => {
+export function createEmbeddableAdapter<T extends Size>(sizeAdapter: SizeAdapter<T>) {
+    return <Props extends { size: Size } & Record<string, any>>(element: ReactElement<Props, any>, { size, ...props }: Props) => {
```
patricklafrance, line 12: Maybe extract this interface?

packages/react-components/src/shared/src/slots.tsx
```diff
@@ -52,15 +52,18 @@ export function useRawSlots(children: ReactNode, slots: string[]) {
     return useMemo(() => getRawSlots(children, slots), [children, slots]);
 }
 
-interface Slots {
+interface RequiredDefaultSlot {
```
patricklafrance, line 55: Hmm why this rename?

Don't like my names to be an Concatenation of the type props.. If I add a "toto" props will I rename this to RequiredDefaultTotoSlot?

patricklafrance: Could be named maybe SlotsDefinition

patricklafrance: Maybe `SlotOptions` ?

patricklafrance (approved):

packages/react-components/src/button/src/ButtonGroup.tsx
alexasselin008: Exactly, since its a local const, i could do a Record<"horizontal" | "vertical", { sm: number, md: number}> i prefer the const synthax in this case

packages/react-components/src/button/src/useToggleButton.ts
alexasselin008: The idea here is to preserve the type of shape in the output variable. Meaning that in some cases, InnerToggleIconButton for instance, the allowed shape properties are circular or rounded, never pill. So if we typed the shape allowing  "pill" | "rounded" | "circular", the output props of the toggle button would also be  "pill" | "rounded" | "circular", even if we specified that the shape can only be rounded or circular

packages/react-components/src/field/src/FieldContext.tsx
alexasselin008: we seem to be using it in the useField typings

packages/react-components/src/shared/src/omitProps.ts
alexasselin008: i can't since i'm using TKey to determine the return value

packages/react-components/src/button/src/IconButton.tsx
alexasselin008: Where do you think would be the best place to put that interface? 

`shared/interactiveStatesProps.ts`?
new folder `sharedTypes/interactiveStates.ts` 

other choice would be to make a ButtonInteractionStatesProps and put it in Button.tsx, with the other buttons referencing it

packages/react-components/src/alert/src/Alert.tsx
alexasselin008: It may be a personal opinion, but i prefer not creating types i really don't care about. In this case, either i do "as const" at the end of the array, or i create a temporary type like this

```
const variations: { tone: keyof typeof Role, icon: ReactElement }[] = [
    { tone: "info", icon: <NotificationIcon /> },
    { tone: "success", icon: <CheckIcon /> },
    { tone: "warning", icon: <WarningIcon /> },
    { tone: "critical", icon: <InfoIcon /> }
];
```
image

Also, the as const is a little bit more precise, since in theory we know it's not a random array of tone and icons, its an array where the first tone is info, the second is success ... 

You can see the typing generated with the as const 
image

packages/react-components/src/field/src/FieldContext.tsx
alexasselin008: Haha agreed, i renamed everything and it simplified the typings a lot

packages/react-components/src/field/src/FieldContext.tsx
alexasselin008: Removed that interface, not an issue anymore

packages/react-components/src/button/src/IconButton.tsx
alexasselin008: shared/types/

packages/react-components/src/button/src/useToggleButton.ts
alexasselin008: TODO: TS commentaire

PR #461, opened 2021-03-12
Perfo

fraincs:
## Issue:

CSS was partly causing some performance issues, the root cause was due to the use of `html *` and some `[classˆ="o-ui"]` 

It is not recommended to use `:root` since it has a higher specificity as `html`.

The Dot component was not very permissive color wise, one could not pass it an hexadecimal color nor an rgb/hsl color code, on top of that it was not able to differentiate a global from an aliased color.

## What I did

- Scoped variables to more specific classes.
- Renamed variables in order to easily know if they are global or an alias (heavily inspired by Spectrum)
- Added logic to the Dot component in order to accomodate many situations that were present in our apps


docs/getting-started/Tachyons.stories.mdx
```diff
@@ -165,7 +165,7 @@ Here's a few common use cases to get you started.
 ```jsx
 <span className="flex items-center">
     Check
-    <CheckIcon className="ml1 fill-success-500" />
+    <CheckIcon className="ml1 fill-botanic-500" />
```
patricklafrance, line 168: botanic? is it an existing color in Sharegate specs? Otherwise I don't see how it is an improvement over "success" ?

packages/foundation/src/atoms/core/colors/primary.css
```diff
@@ -1,4 +1,4 @@
-:root {
+/* [class^="o-ui"] {
```
patricklafrance, line 1: code in comment here

patricklafrance:
Would really help to have some kind of documentation about your interpretation of "global" and "aliases" because I would personally not have expected to see any "alias" in "/foundation"  and "/tachyons".

Can't review this PR without having access to more documentation.

packages/react-components/src/dot/src/Dot.tsx
```diff
@@ -65,6 +65,30 @@ export function InnerDot(props: InnerDotProps) {
...
+            } else {
+                dotCSSDeclaration = {
+                    "--o-ui-dot-color": `var(--o-ui-global-${color})`
+                };
+            }
+        }
+
+        return dotCSSDeclaration;
+    }, [color]);
+}
```
patricklafrance, line 90: I would personally rename this to `useColor` and return the actual color value instead of a style object. The consumer can take care of building the style object if this is the intended use.

This way, it is more generic and less specific to a given context. If needed, this code could then be used elsewhere.

packages/react-components/src/dot/src/Dot.tsx
```diff
@@ -65,6 +65,30 @@ export function InnerDot(props: InnerDotProps) {
...
 
+function useColorStyle(color: string) {
+    return useMemo(() => {
+        let dotCSSDeclaration = {};
+
+        if (color) {
+            if (color.startsWith("rgb") || color.startsWith("hsl") || color.startsWith("#")) {
+                dotCSSDeclaration = {
+                    "--o-ui-dot-color": color
+                };
```
patricklafrance, line 76: No need for a var for this, you can use multiple "return"

docs/getting-started/Tachyons.stories.mdx
fraincs: It is now, since success was role based, and could have been used for other reasons, we, me and Emie, decided that success was not a good color name, as it could also be used for a positive message. It is now on par with the sunray / marine naming convention.

fraincs:
> Would really help to have some kind of documentation about your interpretation of "global" and "aliases" because I would personally not have expected to see any "alias" in "/foundation" and "/tachyons".
> 
> Can't review this PR without having access to more documentation.

This now lives in the RFC folder.

packages/react-components/src/dot/src/Dot.tsx
fraincs: If need it twice this would be moved in an utility component.

packages/react-components/src/dot/src/Dot.tsx
```diff
@@ -65,6 +66,22 @@ export function InnerDot(props: InnerDotProps) {
     );
 }
 
+function useColor(color: string) {
+    return useMemo(() => {
+        if (color) {
```
patricklafrance, line 71: if (!isNil(color))

packages/react-components/src/dot/src/Dot.tsx
```diff
@@ -65,6 +66,22 @@ export function InnerDot(props: InnerDotProps) {
...
+            if (color.startsWith("rgb") || color.startsWith("hsl") || color.startsWith("#")) {
+                return color;
+            } else if (color.startsWith("--")) {
+                return `var(${color})`;
+            } else {
+                return `var(--o-ui-global-${color})`;
+            }
+        } else {
+            return;
+        }
```
patricklafrance, line 81: you don't need an `else` to specify a default return.

```
return useMemo(() => {
     if (color) {
            if (color.startsWith("rgb") || color.startsWith("hsl") || color.startsWith("#")) {
                return color;
            } else if (color.startsWith("--")) {
                return `var(${color})`;
            } else {
                return `var(--o-ui-global-${color})`;
            }
    }
    
    return undefined;
}
```

patricklafrance: In fact you could probably even omit `return undefined;` since returning nothing if the same but I think eslint or TS might complain if you do omit the default return.

rfcs/variables-naming.md
```diff
@@ -0,0 +1,13 @@
+# Orbit Variables naming convention
```
patricklafrance, line 1: The name is too vague, Orbit have a lot of type of variables. JS variables, CSS variables, environment variables, ...

This RFC should

patricklafrance: One thing that would be really usefull is to understand the scope behind all this.

What goes in foundation? What goes in a theme?

For example, does `--o-ui-alias-primary-50` goes in foundation? It's not a token.

docs/getting-started/foundation/Foundation.stories.mdx
```diff
@@ -69,25 +69,25 @@ Once installed the default values of the CSS foundation can be customized to you
...
 ```css
 :root {
-    --o-ui-primary-50: #c63a0c;
-    --o-ui-apricot-50: #c63a0c;
-    --o-ui-overcast-50: #c63a0c;
-    --o-ui-desktop-50: #c63a0c;
+    --o-ui-alias-primary-50: #c63a0c;
+    --o-ui-global-apricot-50: #c63a0c;
+    --o-ui-global-overcast-50: #c63a0c;
+    --o-ui-global-desktop-50: #c63a0c;
```
patricklafrance, line 75: Is it still true that can redefine vars in `:root` ?

Should also remove "overcast"

docs/getting-started/foundation/Foundation.stories.mdx
```diff
@@ -45,22 +45,22 @@ A foundation CSS variable can be accessed in a CSS class:
 
 ```css
 .button {
-    color: var(--o-ui-primary-50);
+    color: var(--o-ui-alias-primary-50);
```
patricklafrance, line 48: Does "primary" still belong in foundation? It seem's theme related to me. It's not a token.

packages/foundation/src/atoms/core/colors/brand.css
```diff
@@ -1,38 +1,38 @@
-html * {
+.o-ui-theme {
```
patricklafrance, line 1: Since foundation is all about token I am not sure about naming the scope "o-ui-theme"

packages/foundation/src/atoms/core/shadows.css
```diff
@@ -1,4 +1,9 @@
...
-    --o-ui-shadow-100: 0 4px 12px 0 hsla(223, 13%, var(--o-ui-shadow-lightness), var(--o-ui-shadow-opacity));
-    --o-ui-shadow-200: 0 8px 24px 0 hsla(223, 13%, var(--o-ui-shadow-lightness), var(--o-ui-shadow-opacity));
+.o-ui-dark-color-scheme {
+    --o-ui-global-shadow-100: 0 4px 12px 0 hsla(223, 13%, 3%, 1);
+    --o-ui-global-shadow-200: 0 8px 24px 0 hsla(223, 13%, 3%, 1);
+}
+
+.o-ui-light-color-scheme {
+    --o-ui-global-shadow-100: 0 4px 12px 0 hsla(223, 13%, 56%, .32);
+    --o-ui-global-shadow-200: 0 8px 24px 0 hsla(223, 13%, 56%, .32);
```
patricklafrance, line 8: I believe our boundaries are wrong here. It would really help is clarified in the RFC.

I doubt foundation should switch variable based on the current color scheme. It should only offer tokens (or const if you prefer).

Doesn't make sense to me that a global is color scheme sensitive, it seems to break the whole concept.

patricklafrance: It doesn make sense based on what you stated in your current RFC:

> Variables containing `-alias` in their name are directly affected by themes and are using global variables(see them as constant) who will not change based on the app theme.

But Spectrum also do it?

image

They seem to use "static" for tokens? for color tokens at least:

image

patricklafrance: Maybe your interpretation of "alias" vs "globals" need to be further improved? It might be alright, I am trying to make sense of all this.

packages/foundation/src/atoms/roles/background.css
```diff
@@ -1,70 +1,70 @@
...
-    --o-ui-background-warning-2-light: var(--o-ui-sunray-50);
-    --o-ui-background-positive-1-light: var(--o-ui-green-500);
-    --o-ui-background-positive-1-hover-light: var(--o-ui-green-700);
-    --o-ui-background-positive-1-active-light: var(--o-ui-green-900);
-    --o-ui-background-positive-2-light: var(--o-ui-green-50);
-    --o-ui-background-info-1-light: var(--o-ui-marine-50);
-    --o-ui-background-info-1-hover-light: var(--o-ui-marine-100);
-    --o-ui-background-info-active-light: var(--o-ui-marine-300);
-    --o-ui-background-input-selection-light: var(--o-ui-primary-100);
+    --o-ui-alias-background-light: var(--o-ui-global-white);
```
patricklafrance, line 3: I would not have expected to find "alias" in foundation and furthermore "alias" that are not related to a theme or a color scheme.

packages/react-components/src/dot/docs/Dot.stories.mdx
```diff
@@ -41,11 +41,14 @@ A dot can have a label.
 
 A dot can have a color.
 
+You can use a color name from our foundations(e.g. `sunray-50`), an hexadecimal color code `#c5cac8`, an `hsl`, an `hsla`, a `rgb` and a `rgba` color code. Alias are also accepted as a full var declarations (e.g. `--o-ui-alias-primary-500`).
```
patricklafrance, line 44: or a `rgba` color code

packages/foundation/src/atoms/core/shadows.css
fraincs: Alias are meant to be abstractions on top of our foundations, aka text-secondary, border-invalid, etc. They are often affected by a theme, but could not change. 

What is important to remember is that a global variable is static, it is used by aliases/semantic variables but they are a constant.

packages/foundation/src/atoms/roles/background.css
fraincs: image

I am not sure if this will help explain the light/dark naming of aliases.

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,164 @@
+Theming Strategy in Orbit
+
+## Summary
+
+Orbit has been, in it's infancy, a monolitich piece of CSS, the only theming options were to modify the primary colors in order to adapt to each products in the Sharegate ecosystem(Apricot, Overcast and Desktop). With Orbit 2.0 we aim to support, among things, a dark theme, while opening the door for future theming possibilitiess(high Contrast, density themes). In this RFC we propose a new architecture for CSS Custom Properties and a way to organize them.
+
+## Motivation
+
+Being able to quickly add a design token, and semantic variables, without questionning where to do so, is a main goal of ours. This is why we came up with a strategy, and are moving many custom properties. In order to provide a consistent theme, that behaves in a predictable manne, we are introducing the concept of semantic colors. Semantic colors are a layer over design tokens representing a meaning, e.g. `border-danger` has a semantic meaning while `border-red-500` has none and could be used on many situations. This make our intefaces more cohesive and gives intention to our design choices. As Adobe Spectrum sums it up:
```
alexasselin008, line 9: in a predictable manner

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,164 @@
...
+
+On top of design tokens lives *semantic variables*, these have an assigned meaning and expedite design decisions. One example of a semantic variable would be a `border-danger` variable, this variable would use the `beetle-500` color behind the scene. When in need of a border that represent a negative state one designer wouldn't need to reinvent the wheel and should use this abstraction : `border-danger`.  These are prefixed with `o-ui-alias`.
+
+Here is an example on how semantic variables work at a core level: 
+
+`--o-ui-alias-text-primary` is a semantic variable who uses `--o-ui-global-marine-900` when in a light theme and `--o-ui-global-marine-500` when in a dark theme.
+
+### Supported Themes
+
+Orbit supports two themes, light and dark as well as two color schemes, Apricot and Desktop. Every semantic variables need to support both themes, light and dark. Some variables can be affected by a color scheme, e.g. a primary background is different in Apricot than in Desktop. Color schemes colors can also be affected by a theme.
```
alexasselin008, line 25: I think this paragraph could be written in a way where we don't enumerate all our current themes and color schemes. This would make this section just as good, but would make it be accurate longer. We wouldn't need to update it if we added a theme for instance

rfcs/theming-color-scheme-strategy.md
fraincs: Where would you add this: > in a predictable manner

alexasselin008: There is a typo, `that behaves in a predictable manne` should be `manner`

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
+Theming Strategy in Orbit
+
+## Summary
+
+Orbit has been, in it's infancy, a monolitich piece of CSS, the only theming options were to modify the primary colors in order to adapt to each products in the Sharegate ecosystem. With Orbit 2.0 we aim to support, among things, a dark theme, while opening the door for future theming possibilitiess. In this RFC we propose a new architecture for CSS Custom Properties and a way to organize them.
```
patricklafrance, line 5: possibilitiess -> possibilities (remove an s)

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Motivation
+
+Being able to quickly add design tokens, and semantic variables, without questionning where to do so, is a main goal of ours. This is why we came up with a strategy, and are moving many custom properties. In order to provide a consistent theme, we are introducing the concept of semantic colors. Semantic colors are a layer over design tokens representing a meaning, e.g. `border-danger` has a semantic meaning while `border-red-500` has none and could be used in many situations. This make our intefaces more cohesive and gives intention to our design choices. As Adobe Spectrum sums it up:
```
patricklafrance, line 9: 100% but I would also add that semantic colors also make the codebase easier to maintains, which is very important.

If we choose that the color of our dabger border have to change color we can update one var instead of find & replace every `border-red-500`.

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+> Too much color can create cognitive overload, affecting users’ ability to efficiently interact with products.
+
+### Design Tokens
```
patricklafrance, line 13: Not sure I would call this section "Design Tokens" we kind of emulate the concept of design tokens with our global var but it's not "design tokens" we can it like this because it's easier for us to understand.

I would call this sections "CSS Variables"

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+Design tokens are the values needed to construct and maintain our design system, from spacing, color, typograghy, to shadows. These values are reflected in the form of CSS Custom Properties(variables). These are prefixed with `o-ui-global`. Design tokens are not affected by a theme. or color scheme, they are what one could call a constant.
```
patricklafrance, line 15: I would change every occurence of "Design tokens" for "CSS variables" and add something along the way to think of global variables as our "design tokens" or "constants"

rfcs/theming-color-scheme-strategy.md
patricklafrance: To match with your next example you should use `border-beetle-500` instead of `border-red-500` here.

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+On top of design tokens lives *semantic variables*, these have an assigned meaning and expedite design decisions. One example of a semantic variable would be a `border-danger` variable, this variable would use the `beetle-500` color behind the scene. When in need of a border that represent a negative state one designer wouldn't need to reinvent the wheel and should use this abstraction : `border-danger`.  These are prefixed with `o-ui-alias`.
```
patricklafrance, line 17: It get confusing when you mix "danger" and "negative" you should pick one.

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+Here is an example on how semantic variables work at a core level: 
+
+`--o-ui-alias-text-primary` is a semantic variable who uses `--o-ui-global-marine-900` when in a *light* theme and `--o-ui-global-marine-500` when in a *dark* theme.
```
patricklafrance, line 21: You should continue with the danger / beetle example instead of introducing a new one here, it would be easier to understand.

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Supported Themes
+
+Orbit supports themes as well as color schemes. Themes can ben defined with adjectives, dark, light, dense, etc while color schemes are a bunch of colors reorganized. Semantic variables need to support themes. Some variables can also  be affected by a color scheme. Color schemes colors can also be affected by a theme.
+
+| ![theme-vs-color-scheme.jpg](https://raw.githubusercontent.com/gsoft-inc/sg-orbit/master/rfcs/assets/theme-vs-color-scheme.jpg?raw=true) |
```
patricklafrance, line 27: You inverted color scheme and theme in the picture

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+On top of design tokens lives *semantic variables*, these have an assigned meaning and expedite design decisions. One example of a semantic variable would be a `border-danger` variable, this variable would use the `beetle-500` color behind the scene. When in need of a border that represent a negative state one designer wouldn't need to reinvent the wheel and should use this abstraction : `border-danger`.  These are prefixed with `o-ui-alias`.
+
+Here is an example on how semantic variables work at a core level: 
+
+`--o-ui-alias-text-primary` is a semantic variable who uses `--o-ui-global-marine-900` when in a *light* theme and `--o-ui-global-marine-500` when in a *dark* theme.
+
+### Supported Themes
+
+Orbit supports themes as well as color schemes. Themes can ben defined with adjectives, dark, light, dense, etc while color schemes are a bunch of colors reorganized. Semantic variables need to support themes. Some variables can also  be affected by a color scheme. Color schemes colors can also be affected by a theme.
```
patricklafrance, line 25: color schemes are "light", "dark", etc.. I dont need the previous paragraph is accurate.

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+#### Alias variables
+
+While developing or modifying an Orbit component, it is important to use alias variables whenever possible. Remember that aliases are theme sensitive, which is what we aim for in order to support a dark theme as well as keeping apps concise.
+
+Here instead of using `color: var(--o-ui-global-marine-900);` we refer to the defined primary text color.
+
+``` css
+// Label.css
+.o-ui-field-label {
+    color: var(--o-ui-alias-text-1);
```
patricklafrance, line 42: Since you are refering to a primary text color in the text following the exemple I believe you should a primary text alias here

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Supported Themes
+
+Orbit supports themes as well as color schemes. Themes can ben defined with adjectives, dark, light, dense, etc while color schemes are a bunch of colors reorganized. Semantic variables need to support themes. Some variables can also  be affected by a color scheme. Color schemes colors can also be affected by a theme.
+
+| ![theme-vs-color-scheme.jpg](https://raw.githubusercontent.com/gsoft-inc/sg-orbit/master/rfcs/assets/theme-vs-color-scheme.jpg?raw=true) |
+| :----------------------------------------------------------: |
+| *Themes and Color Schemes while similar are not affecting colors the same way* |
+
+### Components usage
```
patricklafrance, line 31: I would move all the content of this section under the section currently called "Design tokens". What is described here is not specific to "components".

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+- colors : all of Orbit's colors
+- background : aliases for our backgrounds
+- text: aliases for our text colors
+- lines: aliases for our borders
+- icons: aliases for our icon fills
+- shadows: shadow definitions as well as their aliases
+- spacing / type scale
+
+### Implementation in apps
```
patricklafrance, line 138: This section should be renamed to "Usage in Apps"

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+// ...
+
+.circle {
+    border-radius: var(--o-ui-circular-border-radius);
+}
+```
+
+### Implementation in Orbit
+
+The *Theme Provider* componen is responsible of adding the 3 needed classes: `-o-ui o-ui-{theme} -o-ui-{color-scheme}) in order to benefit from the complete Orbit UI experience.
```
patricklafrance, line 97: You should move the previous paragraph to the apps section. We don't really care about this when developping in Orbit Apps do care.

I would move it somewhere after this block:

>- `o-ui` gives you access to design tokens via CSS Custom Properties
>- `o-ui-dark` `o-ui-light` gives you access to semantic tokens via CSS Custom Properties
>- `o-ui-apricot` `o-ui-desktop` gives you access to brand related semantic tokens, e.g. `--o-ui-primary-50`

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+Orbit's utility variables are declared at the root CSS file `react-components/src/index.css` as well as generic CSS declarations like the app default font-color.
```
patricklafrance, line 99: Should mention that utility variables shouldn't have any prefix (if this is what we choose to do)

rfcs/theming-color-scheme-strategy.md
```diff
@@ -0,0 +1,168 @@
...
+
+In order to be able to benefits from themes, an app need some setup. Apart from having Orbit installed, see Orbit documentation, 3 classes are needed at the root element of your app, Orbit ThemeProvider takes care of this. Here is the classes provided along with what they do
+
+- `o-ui` gives you access to design tokens via CSS Custom Properties
+- `o-ui-dark` `o-ui-light` gives you access to semantic tokens via CSS Custom Properties
+- `o-ui-apricot` `o-ui-desktop` gives you access to brand related semantic tokens, e.g. `--o-ui-primary-50`
+
+Once setup you can refer to these variables in your CSS, as well as use their utility classes counterpart.
+
+Using a semantic token in your CSS
```
patricklafrance, line 148: We should probably have some kind of ThemeProvider example? Basically, for an app, the ThemeProvider is the root of all this.

PR #464, opened 2021-03-18
Fixed the documentation for the item and section elements

alexasselin008:
Issue:

## How to test


packages/react-components/src/placeholders/src/Item.tsx
```diff
@@ -20,7 +20,8 @@ export interface InnerItemProps {
     forwardedRef: ForwardedRef<any>
 }
 
-export function InnerItem() {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function InnerItem(_props: InnerItemProps) {
     // When returning null, react-docgen doesn't ignore the component.
     return <></>;
```
patricklafrance, line 26: Do you still need to return the Fragment then or you could return `null` ? If you can return `null` it would be better!

alexasselin008: it works!

patricklafrance (approved):

PR #465, opened 2021-03-23
RFC OrbitComponent and the As property

alexasselin008:


rfcs/as-prop-and-custom-component.md
```diff
@@ -0,0 +1,154 @@
...
+
+The `OrbitComponent` is a type that is used to mark specific components as Orbit components rather than using ForwardRefExoticComponent.
+This is because an `OrbitComponent` comes with the specific props of the component and adds properties extracted from the as property type. It also removes the forwardedRef property declared in the Inner[X]Props interface, which is a standard pattern in Orbit.
+
+`OrbitComponent` takes 2 type generic, the props of the component, and the specific type of the as property. 
+
+The type property could be one of the following : 
+- an element type the element type (like "div", "button", etc), 
+- a React component Type
+- Element, or any of its subtype.
```
patricklafrance, line 100: Honestly I don't think I would understand this whole OrbitComponent section if is was not previously aware of what you did ;)

patricklafrance: Would also be nice to have an example oh how `OrbitComponent` can be used without `forwardRef`. Maybe a similar exemple for `ForwardRefExoticComponent` ?

PR #467, opened 2021-03-24
Convert Tag, Popover, Menu and Tabs to TypeScript

alexasselin008:


alexasselin008:
This PR can not be merge for now,  we must validate a couple of things that causes error in this pr. 

1 - TabList passes selectedIndex to useKeyedRovingFocus and useAutoFocusChild -> target. those types don't seem valid. This might be an error?

2- useTabsItems returns an array of Tab and Panels, and we expect the property disabled, but this value does not exist. should we add it to the interface or remove the code in Tabs.tsx

3- Tabs.tsx has a property autofocus wich is a boolean  | number, but we never do the conversion to boolean from number

packages/react-components/src/collection/useCollection.ts
```diff
@@ -17,16 +17,18 @@ export interface CollectionItem {
...
 
 export interface CollectionSection extends CollectionItem {
-    items: CollectionItem[]
+    items?: CollectionItem[]
 }
 
 export type CollectionDivider = CollectionItem
 
+export interface CollectionOptionTooltip {
+    props: Record<string, any>;
```
alexasselin008, line 26: weak type

packages/react-components/src/menu/src/MenuItem.tsx
```diff
@@ -112,7 +120,7 @@ export function InnerMenuItem({
                         hover && "hover"
                     ),
                     role: "menuitem",
-                    tabIndex: "-1",
```
alexasselin008, line 115: ok to do this? tabIndex takes a number, not a string

packages/react-components/src/menu/src/MenuTrigger.tsx
```diff
@@ -121,11 +129,11 @@ export function InnerMenuTrigger({
     // Open the menu on up & down arrow keydown.
     const handleTriggerKeyDown = useEventCallback(event => {
         switch (event.key) {
-            case Keys.down:
```
alexasselin008, line 124: Let's go TypeScript!!

packages/react-components/src/menu/src/MenuTrigger.tsx
```diff
@@ -137,7 +145,7 @@ export function InnerMenuTrigger({
             onSelect(event, key);
         }
 
-        close();
```
alexasselin008, line 140: close takes an event, is it ok?

packages/react-components/src/popover/src/PopoverTrigger.tsx
```diff
@@ -87,14 +97,14 @@ export function InnerPopoverTrigger({
...
     ...rest
-}) {
+}: InnerPopoverTriggerProps) {
     const { isOpen, setIsOpen, triggerProps, overlayProps, arrowProps } = usePopup("dialog", {
         id,
         open,
         defaultOpen,
         onOpenChange,
         hideOnEscape: true,
-        hideOLeave: true,
```
alexasselin008, line 97: Let's go TypeScript !!

packages/react-components/src/menu/src/Menu.tsx
```diff
@@ -20,66 +24,79 @@ import {
...
     /**
      * The type of selection that is allowed.
      */
-    selectionMode: oneOf(["none", "single", "multiple"]),
+    selectionMode?: SelectionMode;
     /**
      * Items to render.
      */
-    items: arrayOf(object),
+    items?: CollectionItem[],
```
alexasselin008, line 59: might put any in the documentation

packages/react-components/src/popover/src/PopoverTrigger.tsx
```diff
@@ -92,14 +102,14 @@ export function InnerPopoverTrigger({
...
     ...rest
-}) {
+}: InnerPopoverTriggerProps) {
     const { isOpen, setIsOpen, triggerProps, overlayProps, arrowProps } = usePopup("dialog", {
         id,
         open,
         defaultOpen,
         onOpenChange,
         hideOnEscape: !persistent,
-        hideOLeave: !persistent,
```
alexasselin008, line 102: Let's go TypeScript!

packages/react-components/src/menu/src/MenuItem.tsx
```diff
@@ -1,31 +1,44 @@
...
+export interface InnerMenuItemProps extends InteractionStatesProps {
+    /**
+     * @ignore
+     */
+    id?: string;
     /**
      * Matching collection item.
      */
-    item: object.isRequired,
+    item: {
```
alexasselin008, line 18: probably won't look good in the documentation

alexasselin008: the type is `{ key: string; tooltip?: { props: Record<string, any>; content: ReactElement<any, string | JSXElementConstructor<any>>; }; }`

is this acceptable for this case? or we prefer to remove the type validation and go with "object"

packages/react-components/src/collection/src/useCollection.ts
```diff
@@ -14,14 +14,14 @@ export interface CollectionItem {
     content: ElementType | ReactElement[];
     props: Record<string, any>,
```
patricklafrance, line 15: "," -> ";"

packages/react-components/src/collection/src/useCollection.ts
```diff
@@ -14,14 +14,14 @@ export interface CollectionItem {
...
-        content: ReactElement
+        props: Record<string, any>;
+        content: ReactElement;
     },
 }
 
 export interface CollectionSection extends CollectionItem {
     type: NodeType.section;
-    items: CollectionItem[]
+    items?: CollectionItem[]
```
patricklafrance, line 24: add a ";" ?

Why do we use a ";" instead of a "," ? I personally keep making the mistake of using "," and it seems accepted since no compiler errors and we don't have any linting rules on this.

- Spectrum is using ","
- Chakra doesn't use any character

I would personally go for using a "," because an interface syntax is similar to an object literal so it feels natural.

packages/react-components/src/menu/src/Menu.tsx
```diff
@@ -20,66 +23,79 @@ import {
 import { MenuContext } from "./MenuContext";
 import { MenuItem } from "./MenuItem";
 import { MenuSection } from "./MenuSection";
-import { NodeType, useCollection } from "../../collection";
 import { SelectionMode } from "./selectionMode";
```
patricklafrance, line 26: I think SelectionMode could be move back to Menu.tsx. I extracted the enum to prevent circular dependency since Menu import MenuItem and MenuItem was importing SelectionMode from Menu but with `ìmport type` it should be fine.

packages/react-components/src/menu/src/Menu.tsx
```diff
@@ -20,66 +23,79 @@ import {
...
      * @param {String[]} keys - The keys of the selected items..
      * @returns {void}
      */
-    onSelectionChange: func,
+    onSelectionChange?(event: SyntheticEvent, keys: string[]): void
     /**
      * The type of selection that is allowed.
      */
-    selectionMode: oneOf(["none", "single", "multiple"]),
+    selectionMode?: SelectionMode;
```
patricklafrance, line 54: Does it render well in Storybook props table?

packages/react-components/src/menu/src/MenuItem.tsx
patricklafrance: For Listbox you used `CollectionItem` how does it look?

packages/react-components/src/menu/src/MenuItem.tsx
```diff
@@ -74,7 +87,7 @@ export function InnerMenuItem({
         _: {
             defaultWrapper: Text
         },
-        icon: (element, allElements) => {
+        icon: (_element: ReactElement, allElements: Record<string, any>) => {
```
patricklafrance, line 90: Why the "_" ?

Could use the same name suggestion as I suggested in Listbox.

packages/react-components/src/menu/src/MenuTrigger.tsx
```diff
@@ -71,10 +78,10 @@ export function InnerMenuTrigger({
     children,
     forwardedRef,
     ...rest
-}) {
+}: InnerMenuTriggerProps) {
     const [focusTargetRef, setFocusTarget] = useRefState(null);
 
-    const handleOpenChange = useChainedEventCallback(onOpenChange, (event, isVisible) => {
+    const handleOpenChange = useChainedEventCallback(onOpenChange, (_event, isVisible) => {
```
patricklafrance, line 84: Why the "_" ?

packages/react-components/src/popover/src/PopoverTrigger.tsx
```diff
@@ -1,79 +1,89 @@
...
-import { augmentElement, mergeProps, resolveChildren } from "../../shared";
+import { augmentElement, forwardRef, mergeProps, resolveChildren } from "../../shared";
 import { isNil } from "lodash";
 
-const propTypes = {
+export interface InnerPopoverTriggerProps {
+    /**
+     * @ignore
+     */
+    id?: string;
```
patricklafrance, line 11: A DomProps interface could always be useful?

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/shared/src/dom.d.ts#L47

packages/react-components/src/popover/src/PopoverTrigger.tsx
patricklafrance: haha, oh well we need to test the behavior lol

This is all great that TS find this but it really showcase the lack of interaction tests and the huge debt which comes with this release :(((((

packages/react-components/src/tabs/src/TabList.tsx
```diff
@@ -21,11 +22,17 @@ const NavigationKeyBinding = {
     }
 };
 
+
+export interface TabListProps extends Omit<BoxProps, "autofocus"> {
```
patricklafrance, line 26: Not sure I understand what you are trying to do with this type definition by using `BoxProps` ?

packages/react-components/src/tabs/src/TabsContext.ts
```diff
@@ -0,0 +1,14 @@
+import { SyntheticEvent, createContext, useContext } from "react";
+
+export interface TabsContextType {
+    selectedKey?: string;
+    onSelect?(event: SyntheticEvent, newKey: string): void;
```
patricklafrance, line 5: newKey -> key

packages/react-components/src/collection/src/useCollection.ts
alexasselin008: I found a eslint rule that ensures this. I'll add it after the PRs

packages/react-components/src/collection/src/useCollection.ts
alexasselin008: I found a eslint rule that ensures this. I'll add it after the PRs.

Typescript historically only allowed semi colon. Since TS 1.6, they support a colon or nothing as the delimiter.

TS still uses semicolon in all their examples, and its the most common delimiter i've seen. 
It's also the default in the eslint rule, specifying its the preferred synthax
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-delimiter-style.md

packages/react-components/src/menu/src/Menu.tsx
alexasselin008: Since SelectionMode is both used as a type (`selectionMode?: SelectionMode;`) and as a value (`selectionMode = SelectionMode.none`), it can not be imported using import type, since it would be removed at runtime.

packages/react-components/src/menu/src/MenuTrigger.tsx
alexasselin008: In our TSConfig, we added the noUnusedParameter: true. 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals

Parameters declaration with names starting with _ are exempt from the unused parameter checking. e.g.:


This rules helps maintain a cleaner codebase, notifying you when you can remove a parameter from a function. In this case, we have to keep it since we need it to match a onChange event handler. but on custom function, it's easier to notify and remove them

packages/react-components/src/tabs/src/TabList.tsx
alexasselin008: TabList does not have an as property. It also only uses 2 props, autoFocus and Tabs.

Any extra props passed will be forwarded to the Box component. This extends represents that. Any props that would be allowed on a box is allowed in this component. We omit the autoFocus, since the default autofocus only accepts a boolean, and we specify boolean | null

packages/react-components/src/tabs/src/TabsContext.ts
alexasselin008: done

packages/react-components/src/menu/src/MenuItem.tsx
alexasselin008: In our TSConfig, we added the noUnusedParameter: true.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals

Parameters declaration with names starting with _ are exempt from the unused parameter checking. e.g.:

This rules helps maintain a cleaner codebase, notifying you when you can remove a parameter from a function. In this case, we have to keep it since we need it to match a onChange event handler. but on custom function, it's easier to notify and remove them

alexasselin008: done for the variable rename

packages/react-components/src/popover/src/PopoverTrigger.tsx
alexasselin008: outdated

packages/react-components/src/menu/src/MenuItem.tsx
alexasselin008: image

Do we prefer to use a type for documentation like object, but not having the correct type validation, or do we prefer to keep CollectionItem and have any in the documentation?

packages/react-components/src/menu/src/Menu.tsx
alexasselin008: image

packages/react-components/src/collection/src/useCollection.ts
patricklafrance: All good then!

packages/react-components/src/menu/src/Menu.tsx
```diff
@@ -174,55 +187,55 @@ export function InnerMenu({
...
         index,
-        elementType: ElementType = MenuItem,
+        elementType: As = MenuItem,
         ref,
         content,
         props,
         tooltip
-    }) => (
-        <ElementType
+    }: Omit<CollectionItem, "type">) => (
```
patricklafrance, line 195: Wondering why in this case you choose to use "Omit" instead of "Partial" ?

packages/react-components/src/tabs/src/TabList.tsx
patricklafrance: Ok but this is the case for all our components, we basically always endup forwarding the props to a `<Box>` component and the other components interface aren't define like this.

packages/react-components/src/menu/src/Menu.tsx
```diff
@@ -20,55 +24,64 @@ import {
...
     /**
      * React children.
      */
-    children: oneOfType([any, func])
-};
+    children: ReactNode;
+    /**
+     * @ignore
+     */
+    "aria-label"?: string;
```
patricklafrance, line 75: Maybe it's in another PR but I wonder why a `DomProps` interface has been introduce but not an `AriaLabelingProps ` interface? IMO it's the coolest one ;)

packages/react-components/src/menu/src/Menu.tsx
alexasselin008: done

packages/react-components/src/menu/src/Menu.tsx
alexasselin008: other PR

packages/react-components/src/tabs/src/TabList.tsx
alexasselin008: TabList now have a ref

patricklafrance (approved):

PR #468, opened 2021-03-24
Input, Checkbox, Switch and Radio converted to TypeScript

alexasselin008:


packages/react-components/src/radio/src/RadioGroup.tsx
```diff
@@ -120,6 +124,9 @@ export function InnerRadioGroup(props) {
         disabled,
         children,
         forwardedRef,
+        // we don't want size passed down to Group, since
```
alexasselin008, line 127: size used to be in the rest property, but nobody down the line supports this props, we pass it down to a Box, whos passes it down to a div, but its not supported there.

packages/react-components/src/radio/src/RadioGroup.tsx
```diff
@@ -158,12 +165,12 @@ export function InnerRadioGroup(props) {
     const { groupProps, itemProps } = useGroupInput({
         cssModule: "o-ui-radio-group",
         role: "radio-group",
-        keyProp: KeyProp,
-        value,
-        defaultValue,
+        // keyProp: KeyProp, //TODO: why is this passed but not used?
```
alexasselin008, line 168: useGroupInput doesn't use those property, what do we do?

packages/react-components/src/radio/src/RadioGroup.tsx
```diff
@@ -158,12 +165,12 @@ export function InnerRadioGroup(props) {
     const { groupProps, itemProps } = useGroupInput({
         cssModule: "o-ui-radio-group",
         role: "radiogroup",
-        keyProp: RadioKeyProp,
-        value,
-        defaultValue,
+        // keyProp: RadioKeyProp, //TODO: why is this passed but not used?
```
alexasselin008, line 168: why are we forwarding useless props? GroupInput doesnt use them

packages/react-components/src/checkbox/src/Checkbox.tsx
```diff
@@ -1,79 +1,90 @@
...
     /**
      * The initial value of `indeterminate`.
      */
-    defaultIndeterminate: bool,
+    defaultIndeterminate?: boolean;
     /**
      * The value to associate with when in a group.
      */
-    value: oneOfType([string, number]),
+    value?: string | number;
```
patricklafrance, line 42: I forgot why a checkbox accept a "number". 

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox) specify that a checkbox `value` prop is a DOMString so it doesn't make sense.

I think we should only accept a "string" as value.

Is this something you would to take care of?

packages/react-components/src/input/src/wrappedInputPropsAdapter.ts
```diff
@@ -0,0 +1,29 @@
...
+        ...fieldProps,
+        wrapperProps: {
+            className
+        }
+    };
+
+    return [props, isInField];
+}
+
+export function wrappedInputPropsAdapter({ className, ...rest }: Record<string, any>) {
```
patricklafrance, line 22: I believe it should have the same return type as useFieldWrappedInput (minus the array of course)

packages/react-components/src/input/src/wrappedInputPropsAdapter.ts
```diff
@@ -0,0 +1,29 @@
+import { UseFieldInputPropsReturn, useFieldInputProps } from "../../field";
+
+export interface FieldWrappedInputProps extends Omit<UseFieldInputPropsReturn[0], "className"> {
```
patricklafrance, line 3: This array index thing is weird.

Looking at the code in ``FieldContext.tsx` I would cheat a little bit and do:

```
export type UseFieldInputPropsReturn = {
    validationState?: "valid" | "invalid";
    id?: string;
    required?: boolean;
    disabled?: boolean;
    fluid?: boolean;
    size?: "sm" | "md";
    className?: string;
    "aria-labelledby"?: string;
    "aria-describedby"?: string;
};

export function useFieldInputProps(): [UseFieldInputPropsReturn, boolean] {
```

packages/react-components/src/radio/src/Radio.tsx
```diff
@@ -17,51 +17,68 @@ import {
...
     /**
      * The initial value of `checked` when uncontrolled.
      */
-    defaultChecked: bool,
+    defaultChecked?: boolean;
     /**
      * The value to associate with when in a group.
      */
-    value: oneOfType([string, number]).isRequired,
+    value: string | number
```
patricklafrance, line 36: Same as checkbox, as per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) the value should only be a string.

(Would also requires a few changes like Checkbox, let me know if you want to take care of it)

packages/react-components/src/radio/src/RadioGroup.tsx
patricklafrance: Bad refactor I guess, remove them. I think useGroupInput used to "spread" the rest.

packages/react-components/src/radio/src/RadioGroup.tsx
```diff
@@ -133,7 +140,7 @@ export function InnerRadioGroup(props) {
 
     const groupRef = useMergedRefs(setFocusRef, forwardedRef);
 
-    const handleArrowSelect = useEventCallback((event, element) => {
+    const handleArrowSelect = useEventCallback((_event, element) => {
```
patricklafrance, line 143: Why the "_" ?

packages/react-components/src/radio/src/RadioGroup.tsx
patricklafrance: Why size would be in the "rest" prop? Pretty sure a RadioGroup or Radio doesn't support `size`.

packages/react-components/src/shared/src/checkable.ts
```diff
@@ -1,9 +1,11 @@
 import { ChangeEvent, createContext, useContext } from "react";
 import { isNil } from "lodash";
 
+type CheckableValueType = boolean | string | number | CheckableValueType[]
```
patricklafrance, line 4: CheckableValueType > CheckableValue ?

patricklafrance: Was looking at

```
type CheckableProps = Omit<CheckableContextType, "checkedValue"> & {
    checked?: boolean;
}

export function useCheckableProps({ value }: { value?: boolean | string | number }): [CheckableProps, boolean] {
```

To be consistant, CheckableProps  -> UseCheckablePropsReturn ? Not sure about our naming conventions here we seem to diverge.

packages/react-components/src/shared/src/useRovingFocus.ts
```diff
@@ -57,7 +57,7 @@ export function useRovingFocus(scope: DomScope, { isDisabled = false } = {}) {
 /*
 IMPORTANT: Keyed roving focus doesn't handle disabled elements. This is the responsability of the calling component to ensure that the `currentKey` doesn't match a disabled element.
 */
-export function useKeyedRovingFocus(scope: DomScope, currentKey: string, { keyProp = "value", isDisabled = false } = {}) {
+export function useKeyedRovingFocus(scope: DomScope, currentKey: string | number, { keyProp = "value", isDisabled = false } = {}) {
```
patricklafrance, line 60: I guess the "string | number" would change once Checkbox and Radio are fixed

packages/react-components/src/checkbox/src/Checkbox.tsx
alexasselin008: i don't mind doing the change in this PR. However, according to the doc you linked, any value other than DOMString will be stringified, so it's not exactly true that it only accept a string. According to Typescript's typing of an input, the allowed values are : `value?: string | ReadonlyArray<string> | number;`

packages/react-components/src/input/src/wrappedInputPropsAdapter.ts
alexasselin008: done

packages/react-components/src/input/src/wrappedInputPropsAdapter.ts
alexasselin008: should
```

export function useFieldWrappedInput(): [FieldWrappedInputProps, boolean] {
    const [{ className, ...fieldProps }, isInField] = useFieldInputProps();

    const props = {
        ...fieldProps,
        wrapperProps: {
            className
        }
    };

    return [props, isInField];
}
```

become 
```

export function useFieldWrappedInput(): [FieldWrappedInputProps, boolean] {
    const [fieldProps, isInField] = useFieldInputProps();

    const props =wrappedInputPropsAdapter(fieldProps);

    return [props, isInField];
}
```

packages/react-components/src/radio/src/RadioGroup.tsx
alexasselin008: My fix was a bad one. 

the correct fix would be 
```
const {
        value,
        defaultValue,
        required,
        validationState,
        name,
        onChange,
        autoFocus,
        orientation = "vertical",
        gap,
        wrap,
        reverse,
        disabled,
        children,
        forwardedRef,
        ...rest
    } = mergeProps(
        props,
        toolbarProps,
        omitProps(fieldProps, ["fluid", "size"])  <---- fieldProps contains a size, but size is not supported by <Group
    );
```

patricklafrance: Sounds good

packages/react-components/src/input/src/wrappedInputPropsAdapter.ts
patricklafrance: Sounds good

packages/react-components/src/radio/src/RadioGroup.tsx
alexasselin008: done

packages/react-components/src/shared/src/checkable.ts
alexasselin008: done x2

packages/react-components/src/input/src/wrappedInputPropsAdapter.ts
alexasselin008: done

packages/react-components/src/radio/src/Radio.tsx
alexasselin008: in other pr

patricklafrance (approved):

PR #471, opened 2021-03-25
Update TagsPicker content when the items property changes

tommysanterre:
Issue:
When the items property of TagsPicker changes the content of the dropdown is not updated

## What I did
Added items to autoControlledProps

## How to test

Create a TagsPicker and change it's items prop content with a list containing different values

- Is this testable with Jest or Chromatic screenshots? Maybe


tommysanterre:
Abandoned, we will use the key prop and change it`s value based on the content of the items prop. This way the component will be unmounted and the new items prop value will be used.

PR #478, opened 2021-04-08
Multi browser sanity check

fraincs:
- Added an active state for checkboxes
- Added an hover state for dateRangeInput
- DateInput has invalid chromatic tests now
- Changed the disabled dot test to a more likely use case
- Form Chromatic test now uses uses DateInput
- Input disabled icon now shows a not allowed cursor
- List Component has more explicit default spacing
- Popover and Tooltip are now more distant from their trigger
- Constrained Radio label now shows an ellipsis if missing space
- Loading Tab content is not using SUI anymore
- Fixed an invalid Textarea chromatic test
- Themes CSS were not loaded, this is fixed
- Accordion states were too obvious, they are now less obvious


packages/react-components/src/dot/tests/chromatic/Dot.chroma.jsx
```diff
@@ -24,7 +24,7 @@ stories()
     .add("disabled", () =>
         <Stack>
             <Dot disabled></Dot>
-            <Dot disabled>Habitable</Dot>
+            <Dot color="hsl(25, 69%, 41%)" disabled>Habitable</Dot>
```
patricklafrance, line 27: Dot chromatic "color" tests should be updated to have a color value with every supported color format.

- CSS class
- hexa
- hsl
- etc..

packages/react-components/src/form/tests/chromatic/Form.chroma.jsx
```diff
@@ -130,7 +131,7 @@ stories()
             <div>
                 <Field>
                     <Label>Departure date</Label>
-                    <TextInput placeholder="AAAA/MM/DD" />
+                    <TextInput placeholder="dd/mm/yyyy" />
```
patricklafrance, line 134: Maybe you should also replace this one by a DateInput?

patricklafrance: Maybe change one of these for a DateRangeInput?

packages/react-components/src/menu/src/Menu.css
```diff
@@ -14,6 +14,7 @@
     min-width: 200px;
     max-width: 325px;
     box-shadow: var(--o-ui-alias-elevation-lifted);
+    overflow-y: auto;
```
patricklafrance, line 17: We should have a chromatic test for this

packages/react-components/src/tabs/docs/loader.css
```diff
@@ -0,0 +1,22 @@
+.loading {
```
patricklafrance, line 1: Prefix these classes with "o-ui-sb"

packages/react-components/src/text-input/src/TextInput.jsx
```diff
@@ -152,6 +152,7 @@ export function InnerTextInput(props) {
                     className: cssModule(
                         "o-ui-input",
                         iconMarkup && "has-icon",
+                        disabled && "disabled",
```
patricklafrance, line 155: You don't need to add this class, the CSS selector can target the native "disabled" attribute.

patricklafrance:
Since you changed the spacing between the trigger and the popover + tooltip, should you also do it for Menu? Maybe Select & Autocomplete too?

packages/react-components/src/text-input/src/TextInput.jsx
fraincs: on the input element yes but I need a way to be aware that the whole input is disabled in order to add a not allowed cursor to child like an icon, this is not possible by targeting the input element only.

packages/react-components/src/dot/tests/chromatic/Dot.chroma.jsx
fraincs: > Since you changed the spacing between the trigger and the popover + tooltip, should you also do it for Menu? Maybe Select & Autocomplete too?

this was due to the arrow being too close to the trigger, menu/select and autocomplete do not have those arrows.

packages/react-components/src/form/tests/chromatic/Form.chroma.jsx
fraincs: Maybe you should also replace this one by a DateInput? I was doing it ;)

Good idea. Will update.

PR #483, opened 2021-04-13
convert all remaining files to TS

alexasselin008:
Issue:

## How to test


packages/react-components/src/tag/src/TagList.tsx
```diff
@@ -59,7 +69,7 @@ function TagItem({
             {...mergeProps(
                 rest,
                 {
-                    variant: "outline",
+                    variant: "outline" as const,
```
patricklafrance, line 72: weird? I don't get the "as const" for this one

packages/react-components/src/tag/src/TagList.tsx
```diff
@@ -81,8 +90,9 @@ export function InnerTagList({
     forwardedRef,
     children,
     ...rest
-}) {
-    const nodes = useCollection(children, { items });
+}: InnerTagListProps) {
+    const nodes = useCollection(children);
+    const items = useOnlyCollectionItems(nodes);
```
patricklafrance, line 95: why useCollectionItems? Filtering is useless in this case since it will never be anything else than items. Maybe do an hard cast instead?

Seems weird to call an extra function only for types

alexasselin008: It used to be items passed as a prop down to useCollection. Its not a valid usecase anymore. I copied the way it was done in other places, useCollection + useOnlyCollectionItems. i might be wrong on how to do it

packages/react-components/src/tag/src/TagList.tsx
alexasselin008: it was a mistake

patricklafrance (approved):

PR #484, opened 2021-04-13
Remove the accept all properties from forward ref

alexasselin008:


packages/react-components/src/autocomplete/src/Autocomplete.tsx
```diff
@@ -452,7 +452,7 @@ export function InnerAutocomplete(props: InnerAutocompleteProps) {
                         disabled,
                         validationState,
                         fluid,
-                        active,
+                        active, //TODO: TS active doesnt exist on search input
```
alexasselin008, line 455: active doesnt exist on search input, however we pass it in the auto complete. Should i remove the active property from the AutoComplete?

alexasselin008: @patricklafrance

patricklafrance (approved):

PR #485, opened 2021-04-13
Activate type generation when building the react-component packages

alexasselin008:


packages/react-components/tsconfig.types.json
```diff
@@ -0,0 +1,15 @@
+{
```
alexasselin008, line 1: having multiple tsconfig is generally a good pratice, one for building, one for testing, one for linting.

i copied the naming `tsconfig.types.json` from react-spectrum. its a good name since this tsconfig is only used to generate type. The tsconfig at the root is used for linting

alexasselin008: Unfortunately, the include and outDir options can’t be hoisted in the root config because they are resolved relatively to the config they’re in.

patricklafrance (approved):

Issue #541, opened 2021-04-22
🐛 ThemeProvider's typings should be improved


alexasselin008:
## Describe the bug

setColorScheme accepts a ColorScheme, it should accept ColorSchemeOrSystem. Also, all the property should not be optional. A  better defaultValue should be provided

Before : 
```
export interface ThemeContextType {
    theme?: Theme;
    colorScheme?: ColorScheme;
    setColorScheme?: (newColorScheme: ColorScheme) => void;
}
```

After: 
```
export interface ThemeContextType {
    theme: Theme;
    colorScheme: ColorScheme;
    setColorScheme: (newColorScheme: ColorSchemeOrSystem) => void;
}
```

patricklafrance:
I don't agress with removing the optional theme and colorScheme, default values should be provided instead.

theme -> "apricot"
colorScheme -> "light"

alexasselin008:
@patricklafrance we do pass default values to the ThemeProvider itself 
```
<ThemeProvider colorScheme="light" defaultColorScheme="light" theme="apricot">
```

However, when we consume the 
```
    const { colorScheme, setColorScheme } = useThemeContext();
```

it says setColorScheme might be undefined, wich does not make sense

alexasselin008:
If you meant to provide default values to the context, i've added a comment here if you want to talk about it : 
https://github.com/gsoft-inc/sg-orbit/pull/542/files#r618750017

patricklafrance:
Confused the Context with the Component sorry!

PR #542, opened 2021-04-22
provide better typings to ThemeProvider

alexasselin008:
PR For issue : https://github.com/gsoft-inc/sg-orbit/issues/541


packages/react-components/src/theme-provider/src/ThemeContext.ts
```diff
@@ -1,13 +1,13 @@
 import { createContext, useContext } from "react";
-import type { ColorScheme, Theme } from "./ThemeProvider";
+import type { ColorScheme, ColorSchemeOrSystem, Theme } from "./ThemeProvider";
 
 export interface ThemeContextType {
-    theme?: Theme;
-    colorScheme?: ColorScheme;
-    setColorScheme?: (newColorScheme: ColorScheme) => void;
+    theme: Theme;
```
patricklafrance, line 5: As mentionned in the issue, I think `theme` and `colorScheme` should remain optional

packages/react-components/src/theme-provider/src/ThemeContext.ts
```diff
@@ -1,13 +1,13 @@
...
+    colorScheme: ColorScheme;
+    setColorScheme: (newColorScheme: ColorSchemeOrSystem) => void;
 }
 
-export const ThemeContext = createContext<ThemeContextType>({});
+export const ThemeContext = createContext<ThemeContextType>({} as ThemeContextType);
```
patricklafrance, line 10: I don't get why we need to cast here since we don't have to for other similar cases like 

```
export const DialogTriggerContext = createContext<DialogTriggerContextType>({});

export function useDialogTriggerContext() {
    return useContext(DialogTriggerContext);
}
```

alexasselin008, line 10: did you mean specifying default value here? if so, will it look awkward for setColorScheme? but i'm ok with giving initial values here

packages/react-components/src/theme-provider/src/ThemeProvider.tsx
```diff
@@ -38,11 +38,11 @@ export function ThemeProvider({
     as: TriggerType = "div",
     ...rest
 }: ThemeProviderProps) {
-    const [remoteColorScheme, setRemoteColorScheme] = useState();
+    const [remoteColorScheme, setRemoteColorScheme] = useState<ColorSchemeOrSystem>();
```
patricklafrance, line 41: Should be `ColorScheme` instead of `ColorSchemeOrSystem`.

Basically this is intended to work with some kind a dark mode toggle, doesn't make sense to accept "system" here.

patricklafrance (changes requested):

packages/react-components/src/theme-provider/src/ThemeContext.ts
patricklafrance: Sorry, I thought it was the component interface, closing.

packages/react-components/src/theme-provider/src/ThemeContext.ts
patricklafrance: No, sorry I though it was the component. Still the `<ThemeProvider>` component should define a define value for `colorScheme`.

patricklafrance: I was confused, my bad.

packages/react-components/src/theme-provider/src/ThemeProvider.tsx
alexasselin008: Wouldn't it make sense that the consumer app lets the user choose to "use system settings". Or should i put system as the default value and overwrite when the user changes the toggle the first time? and never go back to system. In our app at the moment, we have 3 buttons, dark/light/system. I can't enforce the system value at the moment
```
<ThemeProvider colorScheme="system" defaultColorScheme="light" theme="apricot">
```

packages/react-components/src/theme-provider/src/ThemeContext.ts
alexasselin008: It means that all the property in DialogTriggerContextType are optional. 

I beleive that the object passed here is used in the case where someone would call useThemeContext without being in a ThemeContext.Provider. it would fallback on those value. 

Since we don't support this case, i forced the type in this situation

patricklafrance: So the context values should remain optional then? If the hook can be called without having a parent ThemeContext.

packages/react-components/src/theme-provider/src/ThemeProvider.tsx
patricklafrance: I dont really see any use cases where the user would go back to "system". Will be glad to offer it if we have the use case.

packages/react-components/src/theme-provider/src/ThemeContext.ts
patricklafrance: Basically the empty object literal is to ensure it could be destructured without having to check against an undefined value.

alexasselin008: Chakra does the same for their color mode provider https://github.com/chakra-ui/chakra-ui/blob/326813ea261ddb6c9f0106e7e63f49f51e111fa2/packages/color-mode/src/color-mode-provider.tsx#L25

packages/react-components/src/theme-provider/src/ThemeProvider.tsx
alexasselin008: Perfect, ill revert those changes

packages/react-components/src/theme-provider/src/ThemeContext.ts
patricklafrance: It's a pattern that have been generalized and was working fine without TS.

With TS, I believe those default empty object literals should be replaced by undefined (well no default value) and it should be the responsability of the consumer to ensure it's not undefined before destructuring?

Therefore the context props could be required.

That being said, I would prefer to do it accross the whole codebase not only for this usecase. I can create an issue for this.

I doubt this typing really block your migration?

alexasselin008: This is not blocking, so i don't mind looking at a solution that would be good for the whole codebase

Issue #567, opened 2021-04-27
🐛 Properly type every SyntheticEvent


alexasselin008:
## Describe the bug

I tought the SyntheticEvent would be ok in most cases. Turns ou it's not ideal. I would change them to their most accurate type

## Steps to reproduce

In a typescript codebase, try to code the following : 
![image](https://user-images.githubusercontent.com/38871812/116305977-7ec6cc00-a772-11eb-9c68-9063168a7562.png)

If we change it on our side : 
![image](https://user-images.githubusercontent.com/38871812/116305943-740c3700-a772-11eb-8a4e-74a9392814fa.png)

## Expected results

works properly

PR #573, opened 2021-04-28
Type every event with the appropriate type

alexasselin008:
PR for #567


packages/react-components/src/alert/src/Alert.tsx
```diff
@@ -60,7 +60,7 @@ export interface InnerAlertProps {
      * @param {SyntheticEvent} event - React's original SyntheticEvent.
      * @returns {void}
      */
-    onDismiss?: (event: SyntheticEvent) => void;
```
alexasselin008, line 63: not very useful, could still be synthetic event

patricklafrance: What do you mean by not useful?

packages/react-components/src/menu/src/MenuTrigger.tsx
```diff
@@ -107,18 +107,18 @@ export function InnerMenuTrigger({
         throw new Error("A menu trigger must have exactly 2 children.");
     }
 
-    const open = useCallback((event, focusTarget) => {
+    const open = useCallback((event: SyntheticEvent, focusTarget: string) => {
```
patricklafrance, line 110: For Menu, I wonder if a union type FocusTarget should be added?

I am not 100% sure because a focusTarget could be "first", "last" or any string key.

Une item, a union type should be added for FocusTarget with values being

patricklafrance: If we do add a union type, it could be used here

patricklafrance (approved):

packages/react-components/src/alert/src/Alert.tsx
alexasselin008: This was meant as a note to myself, but the idea would be that even if onDismiss is passed to a button onClick handler, onDismiss is not necessarely tied to a MouseEvent in our "public contract" which is the interface.

 For instance if we had an alert that would be dismissed after x seconds, the event would not be tied to a MouseEvent, but onDismiss could still be called

Issue #599, opened 2021-05-07
🚀 TextLink should be accessible


fraincs:
## Description

TextLink are indiscernible from standard text when surrounded by text. This issue aims to fix this, by setting a new sensible default.

A decision has to be made as to what is the default (underlined or dotted) as well as a name for a no underline textLink, quiet could be a great name.

## Checklist

- [x] Chromatic tests
- [x] Documentation
- [ ] Figma

fraincs:
Here is what I suggest

- isQuiet : bool when a link has no underline in it's default state
- add an OverBackground prop that would be sort of an inverted link (white) for cases where the links are over a background.
- default to an underline link, dotted would be supported via the underline prop with a value of dotted

patricklafrance:
I don't think a TextLink should be that much configurable (adding 2 props like this). Doing so, is a patch instead if taking time to establishing clear guidelines about how we should use links in our apps.

Furthermore, `isQuiet` in Spectrum is a not a link prop, it's ThemeProvider mode which affect all component when set, I don't it should be added to a specific component.

fraincs:
> I don't think a `<TextLink>` should be that much configurable (adding 2 props like this). Doing so, is a patch instead of taking time to establish clear guidelines about how we should use links in our apps.
> 
> Furthermore, `isQuiet` in Spectrum is a not a prop of their Link component, it's a ThemeProvider variant which affect all components when set. I don't think it should be added to a specific component.

Good point concerning the isQuiet, this make sense then.

I agree that we should choose between dotted and underlined and provide a accented link that could be not underlined.

fraincs:
> I suggest having two types of Text links:
> The first one would be basic and classic, one line below the text without any other particularity.
> 
> The second one would also be very basic with a line below the text. The whole would be of a colour like Apricot500. The goal is to have a catchy text link option.
> 
> Finally, I suggest to slowly move towards microanimation with this practical "easy" topic by proposing this https://codepen.io/aaroniker/pen/VwjexVy

This micro animation seems more like a macro animation. This could work in a marketing website but would make our app look childish. I agree with two textLinks one standard and one with the app main color.

GateNk:
If we go by the logic that a link within a body of text must have an underline, and a floating link mustn't, I think it works. Both should still show an underline when hovered at the very least. 

And I think dotted underlines are another pattern entirely as they allow the presentation of UI elements without leaving the page (e.g. tooltips). I wouldn't use them for general links. Imo!

[...]

PR #614, opened 2021-05-13
underline solid is now the default behaviour of a TextLink

fraincs:
Issue: 

This PR fixes the issue https://github.com/gsoft-inc/sg-orbit/issues/599

## Summary

TextLinks were not accessible since there was no way to tell a link was a link visually before hovering it.

## What I did

the underline prop now defaults to solid, none is now an accepted value for the underline prop. 

## How to test

/?path=/story/chromatic-textlink--default

### DIFF
--- a/packages/react-components/src/link/src/Link.css
+++ b/packages/react-components/src/link/src/Link.css
@@ -5,6 +5,7 @@
     padding: 0;
     position: relative;
     color: inherit;
+    /* direct text is responsible of setting the text decoration style */
     text-decoration: none;
 }
 
@@ -136,6 +137,11 @@
 }
 
 /* TEXT | VARIATIONS */
+/* TEXT | VARIATIONS | UNDERLINE | NONE */
+.o-ui-link-underline-none {
+    text-decoration: none;
+}
+
 /* TEXT | VARIATIONS | UNDERLINE | DOTTED */
 .o-ui-link-dotted .o-ui-link-text {
     text-decoration: underline;
--- a/packages/react-components/src/link/src/Link.css
+++ b/packages/react-components/src/link/src/Link.css
@@ -138,7 +138,7 @@
 
 /* TEXT | VARIATIONS */
 /* TEXT | VARIATIONS | UNDERLINE | NONE */
-.o-ui-link-underline-none {
+.o-ui-link-no-underline {
     text-decoration: none;
 }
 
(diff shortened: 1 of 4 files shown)


fraincs:
@patricklafrance do you think o-ui-link-none should be o-ui-link-underline-none instead?

packages/react-components/src/link/tests/chromatic/createTextLinkTestSuite.jsx
```diff
@@ -120,17 +120,16 @@ export function createTextLinkTestSuite(element, stories) {
                 </div>
             </Stack>
         )
-        .add("solid", () =>
-            <Inline verticalAlign="end">
-                <TextLink underline="solid" size="sm" href="#" element={element}>Flight details</TextLink>
-                <TextLink underline="solid" href="#" element={element}>Flight details</TextLink>
-            </Inline>
-        )
```
patricklafrance, line 128: I would keep this test, if we switch our default to something else, we would lose the solid test

packages/react-components/src/link/src/Link.css
```diff
@@ -118,13 +118,13 @@
...
 .o-ui-link[disabled].o-ui-link-hover .o-ui-link-text {
-    text-decoration: none;
+    text-decoration: underline;
 }
 
 /* TEXT | STATE | DISABLED | FOCUS */
 .o-ui-link[disabled]:focus .o-ui-link-text,
 .o-ui-link[disabled].o-ui-link-focus .o-ui-link-text {
-    text-decoration: none;
+    text-decoration: underline;
```
patricklafrance, line 127: I think it would be better if you don't want assume in CSS what is the default underline style.

I would prefer if it's all controlled by our default value in the TextLink component. 

Could we do like we do for variants and render a css class for each underline values?

Something like:

- solid -> solid
- dotted -> dotted
- none -> no-underline

Names are only a suggestion you might have a better proposition

patricklafrance (changes requested):

patricklafrance:
> @patricklafrance do you think o-ui-link-none should be o-ui-link-underline-none instead?

Yes, something like, see my suggestion https://github.com/gsoft-inc/sg-orbit/pull/614#discussion_r631839377

packages/react-components/src/link/tests/chromatic/createTextLinkTestSuite.jsx
fraincs: missed that will add it now

patricklafrance (approved):

Issue #609, opened 2021-05-12
🐛 The Header does not accept a size property (It works, but the property is not available through TypeScript)


alexasselin008:
## Describe the bug

In this example, it shows how to specify a size to the accordion header.
https://orbit.sharegate.design/?path=/docs/accordion--default-story#size

However, the Header component does not accept a size in it's typescript interface. Should the component used here be AccordionHeader?

## Steps to reproduce

```
<Stack>
    <Accordion>
        <Item key="mars">
            <Header size="sm" as="h3"> {/* <---------- */}
                Mars
            </Header>
            <Content>
                Mars is the fourth planet from the Sun and the second-smallest planet.
            </Content>
        </Item>
        <Item key="jupiter">
            <Header size="sm" as="h3">
                Jupiter
            </Header>
            <Content>
                Jupiter is the fifth planet from the Sun and the largest in the Solar System.
            </Content>
        </Item>
        <Item key="venus">
            <Header size="sm" as="h3">
                Venus
            </Header>
            <Content>
                Venus is the second planet from the Sun. It is named after the Roman goddess of love
                and beauty.
            </Content>
        </Item>
    </Accordion>
    <Accordion>
        <Item key="mars">
            <Header as="h3">Mars</Header>
            <Content>
                Mars is the fourth planet from the Sun and the second-smallest planet.
            </Content>
        </Item>
        <Item key="jupiter">
            <Header as="h3">Jupiter</Header>
            <Content>
                Jupiter is the fifth planet from the Sun and the largest in the Solar System.
            </Content>
        </Item>
        <Item key="venus">
            <Header as="h3">Venus</Header>
            <Content>
                Venus is the second planet from the Sun. It is named after the Roman goddess of love
                and beauty.
            </Content>
        </Item>
    </Accordion>
</Stack>
```

## Expected results

```
<Stack>
    <Accordion>
        <Item key="mars">
            <AccordionHeader size="sm" as="h3"> {/* <---------- */}
                Mars
            </Header>
            <Content>
                Mars is the fourth planet from the Sun and the second-smallest planet.
            </Content>
        </Item>
        <Item key="jupiter">
            <AccordionHeader size="sm" as="h3">
                Jupiter
            </Header>
            <Content>
                Jupiter is the fifth planet from the Sun and the largest in the Solar System.
            </Content>
        </Item>
        <Item key="venus">
            <AccordionHeader size="sm" as="h3">
                Venus
            </Header>
            <Content>
[...]

PR #622, opened 2021-05-18
Rework accordion

fraincs:
Issue: 

https://github.com/gsoft-inc/sg-orbit/issues/609

## Summary

Accordion was not useful in its current state, two sizes were not necessary. The UI was marketing aimed and not covering real use cases.

## What I did

Changed the Accordion component in order to only accept 1 size and modified its UI so it matches our product needs.

## How to test

/?path=/story/chromatic-accordion--icon


patricklafrance:
Looks good but I think it miss an hover style?

packages/react-components/src/accordion/src/Accordion.tsx
```diff
@@ -119,7 +124,7 @@ export function InnerAccordion({
             {...mergeProps(
                 rest,
                 {
-                    className: "o-ui-accordion",
+                    className: `o-ui-accordion o-ui-accordion-${variant}`,
```
patricklafrance, line 127: Please use a syntax similaire to this:

```
                    className: cssModule(
                        "o-ui-button",
                        icon && "has-start-icon",
                        endIcon && "has-end-icon"
                    ),
```

patricklafrance (approved):

Issue #632, opened 2021-05-25
🐛 A Select without a value arrow should be the same color as the placeholder


fraincs:
## Describe the bug

![image](https://user-images.githubusercontent.com/361632/119507887-3c76b780-bd3d-11eb-9c65-a30572525b81.png)

When a select has not a set value the right arrow should be the same colour as the placeholder text.

## Steps to reproduce

/?path=/docs/select--default-story#default

## Expected results

The right arrow should be the same colour as the placeholder text

PR #633, opened 2021-05-25
added a placeholder class to the select trigger in order to style the…

fraincs:
… arrow accordingly

Issue: 

#632 

## Summary

The Select trigger arrow was not the same color as the placeholder text when a Select had no value.

## What I did

Added a class to the Select Trigger in order to stylize the arrow accordingly

## How to test

/?path=/docs/select--default-story#default

### DIFF
--- a/packages/react-components/src/select/src/Select.tsx
+++ b/packages/react-components/src/select/src/Select.tsx
@@ -238,7 +238,7 @@ export function InnerSelect(props: InnerSelectProps) {
                             active && "active",
                             focus && "focus",
                             hover && "hover",
-                            isNil(selectedItem) && "placeholder"
+                            isNil(selectedItem) && "has-placeholder"
                         )
                     },
                     triggerProps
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -17000,7 +17000,7 @@ react-popper-tooltip@^3.1.1:
     "@popperjs/core" "^2.5.4"
     react-popper "^2.2.4"
 
-react-popper@2.2.5, react-popper@^2.2.4:
+react-popper@^2.2.4:
```
patricklafrance, line 17003: Hmm, no? ;)

packages/react-components/src/select/src/Select.tsx
```diff
@@ -237,7 +237,8 @@ export function InnerSelect(props: InnerSelectProps) {
                             fluid && "fluid",
                             active && "active",
                             focus && "focus",
-                            hover && "hover"
+                            hover && "hover",
+                            isNil(selectedItem) && "placeholder"
```
patricklafrance, line 241: You should use "has-placeholder" instead which is more in line with what we use elsewhere for this.

yarn.lock
fraincs: Yarn lock acting out here.

patricklafrance (approved):

Issue #532, opened 2021-04-21
🚀  Swap Calibre for TT Interphase in all Components


fraincs:
## Description

Phasing Out Calibre means adapting all components for the smaller, more standard, size of TT Interphase. Every components need to be validated, especially Typographic related components(Paragraph, Heading, Text).

## Checklist

Every items might not be applicable to your task. When an item is not applicable, please ✔️ the item.

- [ ] Chromatic tests
- [ ] Documentation

patricklafrance:
Numbers looks quite different:

Now:
![image](https://user-images.githubusercontent.com/794579/117189927-c241bb80-adac-11eb-90aa-e246fa03b721.png)

Before:
![image](https://user-images.githubusercontent.com/794579/117189963-cd94e700-adac-11eb-82ae-4a080f10b1d8.png)

GateNk:
Ouais c'est normal que les O et 0s dans TT Interphases soient légèrement plus bombés!
![image](https://user-images.githubusercontent.com/41477329/117190148-1056bf00-adad-11eb-8d8f-d472aa2e1cb7.png)

Moi je trouve ça clean comme transition!

patricklafrance:
It seems like everything if more compact & lighter? It's harder to visualize, for example:

Now (https://6092dd1c0217f100ca5b7304--sg-storybook.netlify.app/?path=/story/field--input):
![image](https://user-images.githubusercontent.com/794579/117190271-35e3c880-adad-11eb-95e6-21b307ec4595.png)

Before(https://sg-storybook.netlify.app/?path=/story/field--input):
![image](https://user-images.githubusercontent.com/794579/117190325-485e0200-adad-11eb-8c87-ed27c23c5b58.png)

GateNk:
Indeed, TT seems slightly wider and shorter, and lighter too! But that may be our eyes playing tricks on us :P The mapping was also set the closest equivalents to what we had before, but Calibre was such an oddity that almost everything had to be lowered by 2 points. 

![image](https://user-images.githubusercontent.com/41477329/117193172-8d376800-adb0-11eb-81a1-c9093ab8d215.png)

fraincs:
> I do feel like everything is more compact & lighter? It's harder to visualize, for example:
> 
> Now (https://6092dd1c0217f100ca5b7304--sg-storybook.netlify.app/?path=/story/field--input):
[...]

PR #642, opened 2021-05-31
Issue/532

fraincs:
Issue: 

#532 

## Summary

Orbit was still using Sharegate's old font, Calibre. Calibre is now gone from Orbit in favour of Interphases.

## What I did

- Adapted the doc to reflect the font change.
- Added line height variables
- Replaced the font in all components


packages/react-components/src/text/src/Text.css
```diff
@@ -1,47 +1,47 @@
...
+    line-height: var(--o-ui-global-line-height-4);
 }
 
 .o-ui-text-2xl {
     font-size: var(--o-ui-global-type-scale-4);
-    line-height: 1.3333333;
+    line-height: var(--o-ui-global-line-height-5);
 }
-
+/* todo */
```
alexasselin008, line 31: todo

packages/tachyons/src/core/_font-weight.css
```diff
@@ -7,39 +7,48 @@
 }
 
 .fw1 {
-    font-weight: 100;
+    font-variation-settings: "wght" 100;
```
alexasselin008, line 10: In the release documentation, we should specify that everywhere that we used font-weight: 600 in a css file, we should also specify the equivalent font-variation-settings: "wght" 600;

fraincs: This issue is specific to Firefox Mac, that's why I am not sure if it should be documented since FFox Mac will probably fix this down the line and we will be able to go back to font weights eventually.

alexasselin008 (approved):

Issue #687, opened 2021-06-23
🐛 Listbox Section Title and Menu Description do not the share the same font-weight


fraincs:
## Describe the bug

There is a difference between a description font weight in a ListBox and a Menu.

## Steps to reproduce

Open a Select with a description then a Menu, the font weight is not the same.

## Expected results

The font weight renders the same way.

Issue #686, opened 2021-06-23
🚀 Motion examples could be easier to use


fraincs:
## Description

There is no way in Motion examples in our documentation to test easing and duration at the same time.  A Preview section with two select(easing and timing) as well as a way to start an animation in order to preview it would be a nice improvement.

## Checklist

- [ ] Documentation

PR #688, opened 2021-06-23
Bundle 21-06

fraincs:
Issue: 

#687 / #686


docs/materials/motion/MotionPreview.css
```diff
@@ -0,0 +1,17 @@
+.motion-preview {
+    padding: 1rem 0;
+    margin-top: 2rem;
+}
+
+.object {
```
patricklafrance, line 6: Try to always scope your classes with a module even if it's only for docs, i've been using `o-ui-sb-`.

packages/react-components/src/menu/src/Menu.css
```diff
@@ -150,7 +150,8 @@ IMPORTANT: The Menu component hardcoded a few CSS values, do not remove or renam
 .o-ui-menu-section-title {
     cursor: default;
     display: flex;
-    font-weight: 500;
+    font-weight: 400;
+    font-variation-settings: "wght" 500;
```
patricklafrance, line 154: I haven't confirm but my feeling was the even the font-size wasn't the same between Listbox and Menu sections

docs/materials/motion/MotionPreview.css
```diff
@@ -0,0 +1,17 @@
...
+    --duration: var(--o-ui-easing-duration-2);
+    --easing: var(--o-ui-easing-focus);
```
patricklafrance, line 8: Should the vars also be scoped?

patricklafrance (approved):

packages/react-components/src/menu/src/Menu.css
fraincs: they both share a `font-size: var(--o-ui-global-type-scale-8);` which is fine since section titles are all caps.

PR #719, opened 2021-08-03
Add Accessibility blurb to Alert Story in Storybook

LaurentDario:
Issue: 

## Summary

The goal of this PR is to add a short blurb about the Accessibility implications of the Alert component in Orbit.

## What I did

- Added documentations around the `alertdialog` role
- Added the logic to support `aria-describedby` in Dialog
- Added tests to cover the added logic.

## How to test

Added documentation:
image

Case with only Content:
image

Case with `aria-describedby` configured:
image

### DIFF
--- a/packages/react-components/src/alert/docs/Alert.stories.mdx
+++ b/packages/react-components/src/alert/docs/Alert.stories.mdx
@@ -26,7 +26,7 @@ import { Heading, InnerHeading } from "@react-components/typography";
 
 An alert comes with the `alertdialog` role and should only be used when an alert, error, or warning occurs.
 
-- This role requires an accessible name (through `aria-labelledby` provided by the `<Heading>` element), and an accessible description (using `aria-discribedby` provided by the `<Content>` element).
+- This role requires an accessible name (through `aria-labelledby` provided by the `<Heading>` element), and an accessible description (using `aria-describedby` provided by the `<Content>` element).
 - Alert dialog must have at least one focusable control and focus must be moved to that control when the alert dialog appears. Generally alert dialogs have at least a Confirmation, Close or Cancel button that can be used to move focus to.
 - The tab order inside the alert dialog must wrap.
 
--- a/packages/react-components/src/alert/docs/Alert.stories.mdx
+++ b/packages/react-components/src/alert/docs/Alert.stories.mdx
@@ -32,9 +32,12 @@ It requires:
 
 Label and description can be overwritten with the use of `aria-label` and `aria-describedby` props on the alert.
 
+### Focus
+
 If the alert calls for an action, make sure to set the focus correctly to avoid any unwanted action.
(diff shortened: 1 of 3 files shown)


packages/react-components/src/alert/docs/Alert.stories.mdx
```diff
@@ -20,6 +20,17 @@ import { Heading, InnerHeading } from "@react-components/typography";
     githubPath="/packages/react-components/src/alert/src"
 />
 
+## Accessibility
+
+An alert comes with the `alertdialog` role and should be used to display a confirmation message, an alert, error, or warning.
```
patricklafrance, line 25: "or warning" -> "or a warning"

packages/react-components/src/alert/docs/Alert.stories.mdx
```diff
@@ -20,6 +20,17 @@ import { Heading, InnerHeading } from "@react-components/typography";
...
+
+It requires:
+- The focus to be trapped inside the Alert.
```
patricklafrance, line 28: "the Alert" -> "the alert"

For consistency, everywhere we don't capitalize the components name

packages/react-components/src/alert/docs/Alert.stories.mdx
```diff
@@ -20,6 +20,17 @@ import { Heading, InnerHeading } from "@react-components/typography";
...
+- A label, provided by the `<Heading>`.
+- Optionally, A description, provided by the `<Content>`.
```
patricklafrance, line 30: The Content element isn't optionnal.

packages/react-components/src/alert/docs/Alert.stories.mdx
```diff
@@ -20,6 +20,17 @@ import { Heading, InnerHeading } from "@react-components/typography";
...
+
+Label and Description can be overwritten with the use of `aria-label` and `aria-describedby` on the Alert component.
```
patricklafrance, line 32: "Description" -> "description"
"Alert" -> "alert"

patricklafrance (commented):
I would like to see something like the following since it's an action which must me taken by the devs:

> If any of the buttons in the AlertDialog's footer should be focused when it opens, specify which one via the autoFocusButton prop.

Source: https://react-spectrum.adobe.com/react-spectrum/AlertDialog.html#accessibility

patricklafrance:
Is there any rules related to the content which would make it more accessible? E.g. : "An alert content shouldn't have more than 100 characters".

LaurentDario:
I updated the copy in the documentation!
As far as I'm aware, and from the official documentation I found, there doesn't seem to be any limit on content. It is still important to avoid forcing a user through a very lengthy text, especially when using assistive technology, so I'd recommend always keeping it short and sweet. But that's a case by case situation, not sure we can impose a limit.

patricklafrance (approved):

Issue #557, opened 2021-04-26
🚀 Inference in mergeProps parameters is not optimal


alexasselin008:
## Description

The idea behind `mergeProps` is to allow the consumer to pass multiple properties from multiple sources, and ensure that every props is merged and returned in a resulting object.

For example, if we had a component that accepts `props`, receive props from `useFieldInputProps` and we want to add a couple of extra props on top of that, we would have the following example in JavaScript.

```
export function InnerButtonGroup(props: InnerButtonGroupProps) {
    const [fieldProps, isInField] = useFieldInputProps();

    const {
        orientation = "horizontal",
        align,
        size,
        fluid,
        disabled,
        children,
        forwardedRef,
        ...rest
    } = mergeProps(
        props,
        omitProps(fieldProps, ["fluid"]),
        {
            orientation,
            align,
            verticalAlign: orientation === "horizontal" ? "center" : undefined,
            fluid,
            gap: Gap[orientation][normalizeSize(size)],
            className: cssModule(
                "o-ui-button-group",
                isInField && "as-field"
            ),
            role: !isInField ? "group" : undefined,
            ref: forwardedRef
        }
    );
```

This synthax centralize everything related to props, and is simple to use. However, in typescript, the returning type cannot be properly inferred since the anonymous litteral passed to the method is not typed. Because of [Typescript's Type widening](https://github.com/microsoft/TypeScript-New-Handbook/blob/master/reference/Widening-and-Narrowing.md#literal-type-widening-relationships) verticalAlign is typed with string | undefined instead of "center" | undefined.

### What about React-Spectrum?
We took the mergeProps method from React-Spectrum. The reason why they don't have this issue, is because they never have an untyped object passed as a parameter to this function. 

They take an approach that would look more like this
```
<Group
        {...mergeProps(props, omitProps(fieldProps, ["fluid"])}

        orientation={orientation}
        align={align}
        verticalAlign={orientation === "horizontal" ? "center" : undefined}
        fluid={fluid}
        gap={Gap[orientation][normalizeSize(size)]}
        className={cssModule(
            "o-ui-button-group",
            isInField && "as-field"
        )}
        role={!isInField ? "group" : undefined}
        ref={forwardedRef}
>
```

Since the props are directly infered from the group object, not an anonymous object, everything goes smoothly with TypeScript.

### Here is a list of our options

**Option 1** : The one we use now :  Specifying mergeProps' return type

We don't like that approach since it requires an annoying synthax

```
[...]

PR #727, opened 2021-08-10
Fix mergeProps with as const

alexasselin008:
#557


packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -87,7 +87,7 @@ export interface InnerFlexProps {
     /**
      * @ignore
      */
```
alexasselin008, line 89: error found by properly typing everything

patricklafrance (approved):

PR #730, opened 2021-08-19
Simplify as usage

alexasselin008:


packages/react-components/src/date-input/src/DateInput.tsx
```diff
@@ -43,7 +43,7 @@ export interface DatePreset {
     date: Date;
 }
 
-export interface InnerDateInputProps extends InteractionStatesProps, AriaLabelingProps {
+export interface InnerDateInputProps extends InteractionStatesProps, AriaLabelingProps, Omit<ComponentProps<"input">, "autoFocus" | "defaultValue" | "max" | "min" | "value"> {
```
patricklafrance, line 46: I thought replacing the "autoFocusDelay" prop for an hybrid "autoFocus" prop accepting a bool or a delay was smart... now I understand how bad of an idea it is ;)

packages/react-components/src/date-input/src/DateRangeInput.tsx
```diff
@@ -137,7 +139,7 @@ export interface InnerDateRangeInputProps extends InteractionStatesProps {
     forwardedRef: ForwardedRef<any>;
 }
 
-const DateInput = forwardRef<any, "input">(({
+const DateInput = forwardRef<HTMLInputElement, any>(({
```
patricklafrance, line 142: What's the problem with "input" ? Is it because we do accept additional props like "validationState" and "onDateChange" ?

packages/react-components/src/dialog/src/DialogTrigger.tsx
```diff
@@ -131,7 +133,7 @@ export function InnerDialogTrigger({
     );
 }
 
-export const DialogTrigger = forwardRef<InnerDialogTriggerProps>((props, ref) => (
+export const DialogTrigger = forwardRef<any, Omit<InnerDialogTriggerProps, "forwardedRef">>((props, ref) => (
```
patricklafrance, line 136: Is there a type similar to ComponentProps but for the a DOM type? Something which would allow us to "DomProps<DefaultElement>"

packages/react-components/src/disclosure/src/Disclosure.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { DisclosureContext } from "./DisclosureContext";
-import { DomProps, Keys, augmentElement, cssModule, forwardRef, isNil, mergeProps, resolveChildren, useControllableState, useEventCallback, useId, useMergedRefs } from "../../shared";
+import { DomProps, Keys, augmentElement, cssModule, isNil, mergeProps, resolveChildren, useControllableState, useEventCallback, useId, useMergedRefs } from "../../shared";
 import { Text } from "../../typography";
 import { useSlidingTransition } from "./useSlidingTransition";
 
-export interface InnerDisclosureProps extends DomProps {
+const defaultElement = "div";
+
+export interface InnerDisclosureProps extends DomProps, Omit<ComponentProps<typeof defaultElement>, "color"> {
```
patricklafrance, line 11: Do we still need the "DomProps" interface now that the inner props extends ComponentProps? DomProps only contains an optionnal "id" prop

packages/react-components/src/group/src/Group.tsx
```diff
@@ -1,8 +1,8 @@
-import { ComponentProps, ElementType, ForwardedRef, ReactNode } from "react";
-import { Flex, useFlexAlignment, useFlexDirection } from "../../layout";
-import { forwardRef, isNil, mergeProps } from "../../shared";
+import { ComponentProps, ElementType, ForwardedRef, ReactNode, forwardRef } from "react";
+import { Flex, FlexProps, useFlexAlignment, useFlexDirection } from "../../layout";
+import { isNil, mergeProps } from "../../shared";
 
-export interface InnerGroupProps {
+export interface InnerGroupProps extends Omit<FlexProps, "wrap">{
```
patricklafrance, line 5: I wouldn't extend Flex, I think it's best to duplicate the props as they could diverge in the future. Also, I always viewed Group as a simplified version of Flex. Flex will most likely change a lot in the "CSS in JS" PR.

packages/react-components/src/layout/src/Flex.tsx
```diff
@@ -1,8 +1,8 @@
-import { Box } from "../../box";
-import { CSSProperties, ComponentProps, ElementType, ForwardedRef, ReactNode } from "react";
-import { forwardRef, isNil, isNilOrEmpty, isString, mergeProps } from "../../shared";
+import { Box, BoxProps } from "../../box";
+import { CSSProperties, ComponentProps, ElementType, ForwardedRef, ReactNode, forwardRef } from "react";
+import { isNil, isNilOrEmpty, isString, mergeProps } from "../../shared";
 
-export interface InnerFlexProps {
+export interface InnerFlexProps extends BoxProps{
```
patricklafrance, line 5: Space before "{". Really weird that eslint didn't flag this.

packages/react-components/src/layout/src/Inline.tsx
```diff
@@ -1,9 +1,9 @@
-import { ComponentProps, ElementType, ForwardedRef, ReactNode } from "react";
-import { Flex } from "./Flex";
-import { forwardRef, isNil, mergeProps } from "../../shared";
+import { ComponentProps, ElementType, ForwardedRef, ReactNode, forwardRef } from "react";
+import { Flex, FlexProps } from "./Flex";
+import { isNil, mergeProps } from "../../shared";
 import { useFlexAlignment } from "./adapters";
 
-export interface InnerInlineProps {
+export interface InnerInlineProps extends Omit<FlexProps, "wrap"> {
```
patricklafrance, line 6: Wouldn't extends FlexProps

packages/react-components/src/shared/src/as.tsx
```diff
@@ -0,0 +1,12 @@
+import {
+    ComponentProps,
+    ElementType,
+    JSXElementConstructor
+} from "react";
```
patricklafrance, line 5: Could fit on a single line

packages/react-components/src/shared/src/as.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { Merge } from "type-fest";
+
+export function as<A extends JSXElementConstructor<{ as?: ElementType } & any>, B extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>>(component: A, asProp: B) {
+    const Component = component as JSXElementConstructor<any>;
+
+    return (props: Merge<ComponentProps<A>, ComponentProps<B>>) => <Component as={asProp} {...props} />;
```
patricklafrance, line 11: ❤️ ❤️ ❤️

packages/react-components/src/shared/src/mergeProps.ts
```diff
@@ -115,7 +116,6 @@ function merge(props: Record<string, any>, newProps: Record<string, any>) {
 
 // Taken from: https://stackoverflow.com/questions/51603250/typescript-3-parameter-list-intersection-type/51604379#51604379.
 type TupleTypes<T> = { [P in keyof T]: T[P] } extends { [key: number]: infer V } ? V : never;
-type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never;
```
patricklafrance, line 118: ❤️

packages/react-components/src/shared/src/mergeProps.ts
```diff
@@ -115,7 +116,6 @@ function merge(props: Record<string, any>, newProps: Record<string, any>) {
 
 // Taken from: https://stackoverflow.com/questions/51603250/typescript-3-parameter-list-intersection-type/51604379#51604379.
 type TupleTypes<T> = { [P in keyof T]: T[P] } extends { [key: number]: infer V } ? V : never;
```
patricklafrance, line 118: Would love to see a few tests for this type, it would also make it easier to understand for newcomes (and well established devs like me who keep forgetting how it works haha)

packages/react-components/src/shared/tests/typescript/as.tsx
```diff
@@ -0,0 +1,18 @@
...
+const RouterLink = (props: RouterLinkProps) => {
+    console.log(props.a);
+    console.log(props.b);
+
+    return <div></div>;
+};
+
+const RouterButton = as(Button, RouterLink);
+
+<RouterButton onClick={() => console.log("test")} a="a" b={false} >aa</RouterButton>;
```
patricklafrance, line 18: Why the console.log ?

patricklafrance (changes requested):

patricklafrance:
A few things:

- I think you forgot to update the Accordion component to use an Heading instead of an Header.

- The doc will need to be updated to document the usage of the "as" utility function. It could be a new page called "as" under the "Getting Started" section. I would also update the [following Link doc](https://deploy-preview-730--sg-orbit.netlify.app/?path=/docs/link--default-story#react-router) to reflect the new syntax.

- Yes I think we should export aliased types for a button as a link and vice versa. To prevent confusion, I would however called them ButtonAsLink and LinkAsButton.

- We still need to figure out how we handle our remaining existing "as" use cases like Heading and Tag. For Heading we do expose aliased component like H1, H2, etc.. so we might be fine. For Tag, I don't mind asking the use to "as" utility function. For custom component using Box, we might also want to force the "as" utility function pattern, we could show an example in "as" doc page.

- We also need to figure out if we internally continue to use the Box "as" prop or we switch to the "as" utility function. I think we might be fine with the "as" utility function but as always, I must forget a few use cases.

```
export function MyComponent() {
    const As = as(Box, "input");
    
    return (
         <As value={value} />
    )
}
```

packages/react-components/src/menu/src/Menu.tsx
```diff
@@ -86,6 +86,10 @@ export interface InnerMenuProps extends DomProps, AriaLabelingProps, ComponentPr
      * React children.
      */
     children: ReactNode;
+    /**
+     * @ignore
+     */
+    disabled?: boolean;
```
patricklafrance, line 92: I don't think we still need to define attributes like "disabled" in our interface since they should be included in ComponentProps

PR #731, opened 2021-08-23
Move Typescript configs and utils into a TypeScript folder + add some core type testing

alexasselin008:
Issues encountered :

- Did not code forwardRef.tsx tests, since the as PR will take care of this.
- Instead of adding a third party that specialize in testing types (tsd), we coded utils method to test directly the code in the src folder. However, that comes with some limitations : 
---- only expectAssignable is useable. tsd can use expectNotAssignable, because their CLI is taking care of the analysis, not typescript itself
---- we have to put @ts-expect error before any line using expectAssignable in order to simulate expectNotAssignable.


typescript/tests/helpers.ts
```diff
@@ -0,0 +1,3 @@
+export const expectAssignable = <T>(value: T) => {
```
patricklafrance, line 1: Instead of an helpers.ts file I would create one file per helper function

packages/react-components/src/shared/tests/typescript/arrayify.ts
```diff
@@ -0,0 +1,16 @@
+import { arrayify } from "../../src";
+import { expectAssignable } from "../../../../../../typescript/tests/helpers";
```
patricklafrance, line 2: Could you add an alias for this? maybe a "@typescript" alias and then import from "@typescript/tests"

packages/react-components/src/shared/tests/typescript/arrayify.ts
```diff
@@ -0,0 +1,16 @@
...
+
+expectAssignable<number[]>(arrayify(1));
+expectAssignable<number[]>(arrayify([1]));
+
+// @ts-expect-error
+expectAssignable<number[]>(arrayify(["1"]));
```
patricklafrance, line 8: Since there is a limitation for the "expectNotAssignable" tests, do we really want to write any?

I ended up writing only assignable tests in the "CSS in JS" PR and I do have great coverage.

packages/react-components/src/shared/tests/typescript/slots.tsx
```diff
@@ -0,0 +1,24 @@
...
+        size: "inherit",
+        className: "o-ui-accordion-title"
+    }
+});
+
+expectAssignable<keyof typeof result>("icon");
+expectAssignable<keyof typeof result>("text");
+
+// @ts-expect-error
+expectAssignable<keyof typeof result>("autre");
```
patricklafrance, line 24: What does "autre" mean? Kidding but please replace by an english word :)

patricklafrance (approved):

patricklafrance:
Could you exclude the typescript test files from the build? We could exclude by conventions anything matching the following "**/tests/typescript/*"

alexasselin008:
> Could you exclude the typescript test files from the output? We could exclude by convention anything matching the following "**/tests/typescript/*"

test/typescript is already ignored in the final build

PR #732, opened 2021-08-23
Convert Jest Tests to TypeScript

alexasselin008:
Convert Jest Tests to TypeScript

Issues encountered : 
- jest utils files at the root use module.export instead of normal import / export. Not sure why
- added 2 alias in the tsconfig. Should i make a tsconfig only for the tests, or is that overkill?


jest/utils/errorHandling.tsx
```diff
@@ -1,8 +1,8 @@
 const React = require("react");
```
alexasselin008, line 1: Need to validate why the export / import was made that way, and how to adapt the ts config to allow this

packages/react-components/src/avatar/tests/jest/Avatar.test.tsx
```diff
@@ -1,5 +1,5 @@
 import { Avatar } from "@react-components/avatar";
-import { createRef } from "react";
+import { Ref, createRef } from "react";
```
alexasselin008, line 2: not necessary

packages/react-components/src/shared/src/forwardRef.ts
```diff
@@ -40,6 +40,7 @@ type MergeWithAs<T, P> = RightJoinProps<PropsOf<T>, OmitCommonProps<P, "slot"> &
      * Default slot override. Added to every orbit component
      */
     slot?: string;
+    "data-testid"?: string;
```
alexasselin008, line 43: This will conflict with the Merge with As pr. 

`data-` are not automatically accepted by components. With typescript 4.4, there could be a workaround : 
https://github.com/microsoft/TypeScript/issues/28960#issuecomment-903519759

This is currently a limitation of typescript. We're waiting for microsoft/TypeScript#28960.
For more context see #18902.
Possible workarounds:
@ts-ignore:
<Checkbox
  inputProps={{
    // @ts-ignore
    'data-testid': 'test-checkbox',
  }}
/>;
use ByRole instead of ByTestid. We generally don't recommend data-testid. Most of the (interactions) tests can be implemented using ByRole

tsconfig.json
```diff
@@ -25,7 +25,12 @@
     "isolatedModules": true,
 
     // we don't create declaration files when linting
-    "noEmit": true
+    "noEmit": true,
+    "baseUrl": ".",
+    "paths": {
```
alexasselin008, line 30: These alias are only for tests. Should we create a special config only for testing?

patricklafrance:
> jest utils files at the root use module.export instead of normal import / export. Not sure why

Sorry about that, I was lazy and didn't fix the eslint config. I think you can fix it by adding the following to the eslint config.

```
{
    "parserOptions": {
        "sourceType": "module"
    }
}
```

image

tsconfig.json
patricklafrance: Yes I think that would be better, I guess you could inherit from the "common" configuration and extend it with the jest specific paths without duplicating any config?

How would you then tell Jest to use it's custom tsconfig?

patricklafrance (approved):

PR #739, opened 2021-08-27
Add accessibility blurb on Accordion

LaurentDario:
Issue: 

## Summary

Adding an Accessibility blurb for Accordion

## What I did

Adding informations about `aria-hidden` and keyboard navigation on Accordion

### DIFF
--- a/packages/react-components/src/accordion/docs/Accordion.stories.mdx
+++ b/packages/react-components/src/accordion/docs/Accordion.stories.mdx
@@ -1,9 +1,11 @@
 import { Accordion, InnerAccordion, InnerAccordionHeader, InnerAccordionPanel } from "@react-components/accordion";
 import { ArgsTable, Meta, Story } from "@storybook/addon-docs/blocks";
+import { Button } from "@react-components/button";
 import { ComponentInfo, Preview, Tagline } from "@stories/components";
 import { Content, Header, InnerContent, InnerHeader } from "@react-components/placeholders";
 import { HelpIcon, InfoIcon, NotificationIcon } from "@react-components/icons";
 import { InnerItem, Item } from "@react-components/collection";
+import { Paragraph } from "@react-components/typography";
 import { Text } from "@react-components/typography";
 
 <Meta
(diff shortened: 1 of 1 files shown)


packages/react-components/src/accordion/docs/Accordion.stories.mdx
```diff
@@ -23,6 +25,28 @@ import { Text } from "@react-components/typography";
...
+Orbit is taking care of handling the [`aria-hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute) value required to let the user know if the content is visible.
+
+<Preview>
+    <Story name="accessibility">
+        <Accordion>
+            <Item key="mars">
+                <Header as="h3">Mars</Header>
+                <Content>
+                    <Paragraph>Mars is the fourth planet from the Sun and the second-smallest planet in the Solar System.</Paragraph>
+                    <Button>Cutoff</Button>
```
patricklafrance, line 39: I am not sure why you add a example for the accessibility? This example seem similar to the default one.

patricklafrance (changes requested):

packages/react-components/src/accordion/docs/Accordion.stories.mdx
LaurentDario: I wasn't sure when adding it, not a fan that this is the first example that you come across on the page. But the point was to give an example of a focusable item inside of the tab, not being focusable when collapsed. But I can remove that example.

patricklafrance: I think I would remove the example.

packages/react-components/src/accordion/docs/Accordion.stories.mdx
```diff
@@ -23,6 +25,28 @@ import { Text } from "@react-components/typography";
 
 ## Accessibility
 
+The content in the Accordion is collapsable. For that reason, it is important to ensure the navigation is not focusing on content that is hidden.
```
patricklafrance, line 28: I am not convince we should mention this. I am not sure what the user should do about this? At least not worded like this.

LaurentDario: The point here is to ask the user to make sure the keyboard navigation works and makes sense. That if content is collapsed, it shouldn't be focusable and, the other way around, that focusable elements should be focusable when the accordion content is expanded.
I tried on Storybook, with Button it works, but with TextLink for example, it requires to add a `tabindex="0"` not sure if it's by design or an issue with TextLink in the accordion.

Could be:
> Make sure the keyboard navigation behaves accordingly: If the content is collapsed, only the accordion headers should be focusable. Once the content is expanded, focusable elements within those should be part of the keyboard navigation.

patricklafrance: Could you please provide a short gif illustrating the use case? I have a hard time understanding it.

Is it related to "autoFocus" ?

LaurentDario: I tried again and I think I tested first with a `TextLink` with no `href` set, which doesn't really make sense. With the `href`, the focus works fine:
accordion-focus
That was the point of that comment but being that Links and Buttons are working fine in those examples, I think I can get rid of that sentence unless we want to keep a mention of the keyboard navigation?

patricklafrance: I think I would get rid of the sentence. As we discussed earlier, I think the goal of those accessibility blurbs are not to mention all the accessibility concerns which are implementend by the component but rather to mention which actions the user should take or which decisions he might have to make.

LaurentDario: Then maybe the original copy is enough
> The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).

> That heading element should have a level that is appropriate for the information architecture of the page. You can do this by providing an `as` prop to the `<Header>` component.

The added content is things already covered by the component

patricklafrance: Yeah maybe

LaurentDario:
I'll close this one and consider Accordion covered

Issue #752, opened 2021-10-28
🐛  Autocomplete List Shows visual glitches


fraincs:
## Describe the bug

When typing in an Autocomplete Component there is a white empty cell followed by a 

## Steps to reproduce

Type Ma in an Autocomplete, there is a white empty cell.

![image](https://user-images.githubusercontent.com/361632/139317687-a851e8d0-da34-4a52-b9b1-37d1efb51559.png)

And there is a scroll. 

![image](https://user-images.githubusercontent.com/361632/139319161-1c4bed92-a53a-4b6e-921e-b3bac88d7310.png)

## Expected results

There is no overflow.

PR #756, opened 2021-11-02
Fix Autocomplete spacing glitch

LaurentDario:
Issue: #752 

## Summary

Autocomplete leaves an empty space or adds a scrollbar depending on the content entered.

## What I did

Adding the number of results returned by the autocomplete (as node) as a dependency of useLayoutEffect for useScrollableCollection makes sure we update the size of the content of the autocomplete Listbox.

### DIFF
--- a/packages/components/src/listbox/src/Listbox.tsx
+++ b/packages/components/src/listbox/src/Listbox.tsx
@@ -356,7 +356,7 @@ export function InnerListbox({
         target: selectionManager.selectedKeys[0] ?? defaultFocusTarget
     });
 
-    const scrollableProps = useScrollableCollection(containerRef, {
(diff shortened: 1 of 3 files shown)


packages/components/src/listbox/src/Listbox.tsx
```diff
@@ -356,7 +356,7 @@ export function InnerListbox({
         target: selectionManager.selectedKeys[0] ?? defaultFocusTarget
     });
 
-    const scrollableProps = useScrollableCollection(containerRef, {
+    const scrollableProps = useScrollableCollection(containerRef, nodesProp, {
```
alexasselin008, line 359: passe les nodes, pas nodeprops

fraincs (approved):
Good job

Issue #755, opened 2021-11-02
🐛 Textarea expands on each keystrokes


fraincs:
## Describe the bug

At every character typed in Textarea, the components height expands.

## Steps to reproduce

Fill a TextArea.

## Expected results

The TextArea expands only when full.

## Reproducible demo

/?path=/docs/textarea--default-story

PR #760, opened 2021-11-04
Fix Textarea expanding exponentially by revisiting the default line-h…

LaurentDario:
Issue: 

## Summary

As reported in #755.
The issue happens on the TextArea component. When entering content into the textarea, the number of rows assigned is increasing exponentially as the use types.
This was due to the fact that the maths used to calculate the number of rows was taking a 1:1 ratio for line height set to normal when in reality, depending on the font and the browser that ratio can be 1.2 or even higher.

## What I did

TextArea will add a span at load time, staying hidden, allowing the calculation of the actual height of a line according to the font used.
We needed to introduce the notion of fonts ready to allow for the correct calculation after the fonts were loaded.

## How to test

Tested in storybook on Chrome and Edge. Also tested in Firefox but, while the expand works perfectly, it also adds a small scrollbar on the side of the textarea when adding more rows.

Screen recording: https://share.getcloudapp.com/2NuPqvOD

### DIFF
--- a/packages/components/src/shared/src/useFontFaceReady.ts
+++ b/packages/components/src/shared/src/useFontFaceReady.ts
@@ -1,0 +1,15 @@
+import { useEffect, useState } from "react";
+
+const useFontFaceReady = () => {
+    const [ready, setReady] = useState(false);
+
+    useEffect(() => {
+        document.fonts.ready.then(() => {
+            setReady(true);
+        });
+    }, []);
+
+    return ready;
+};
+
+export { useFontFaceReady };
--- a/packages/components/src/shared/src/useFontFaceReady.ts
+++ b/packages/components/src/shared/src/useFontFaceReady.ts
@@ -4,9 +4,14 @@ const useFontFaceReady = () => {
     const [ready, setReady] = useState(false);
 
     useEffect(() => {
-        document.fonts.ready.then(() => {
-            setReady(true);
-        });
+        let isCancelled = false;
+        const loadFonts = async () => {
+            await document.fonts.ready;
+            if (!isCancelled) { setReady(true); }
+        };
+        loadFonts();
+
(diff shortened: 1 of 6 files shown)


packages/components/src/shared/src/useFontFaceReady.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from "react";
+
+const useFontFaceReady = () => {
```
alexasselin008, line 3: I beleive the current implementation could result in error like this one
image

This is due to the fact that the hook could be unmounted before the callback has been resolved, resulting in calling setReady on an unmounted hook

I believe this is a safer approach to this hook

```
const useFontFaceReady = () => {
    const [ready, setReady] = useState(false);

    useEffect(() => {
        let isCancelled = false;
        
        const loadFonts = async () => {
            await document.fonts.ready;
            if (!isCancelled) {
                setReady(true);
            }
        };

        loadFonts();

        return () => {
            isCancelled = true;
        };
    }, []);

    return ready;
};
```

packages/components/src/text-area/src/TextArea.tsx
```diff
@@ -61,6 +61,36 @@ const pxToInt = (value?: string) => {
     return !isNil(value) ? parseInt(value.replace("px", ""), 10) : 0;
 };
 
+function useCalculateLineHeight(inputRef: any) {
```
alexasselin008, line 64: Pass only the inputRef.current, and type it with 

`inputRef.current as HTMLTextAreaElement`

alexasselin008 (dismissed):

alexasselin008 (commented):
image
do you mind fixing this

alexasselin008 (approved):

Issue #740, opened 2021-09-01
🐛 Message: Flicker on first state change


LaurentDario:
## Describe the bug

We use Message in the Settings page, alongside inputs that are updating the dates displayed in the Messages.
The state update seems to be provoking a flicker on all the Messages under that state, that we are able to reproduce in the documentation.

## Steps to reproduce

Include the Message in a State change. The Message then flickers for the first change. This does not happen then for any further changes:
![orbit-message-flicker](https://user-images.githubusercontent.com/234418/131693664-7c5b195a-1091-40cf-8450-7bcbf2a579f9.gif)

## Expected results

No flicker on change.

patricklafrance:
Added the issue to our development board to investigate what's going on.

alexasselin008:
Repro code

https://orbit.sharegate.design/?path=/docs/message--default-story#controlled


```
() => {
    const [count, setCount] = useState(0);

    const toggleVisibility = useCallback(() => {
        
    }, [setCount]);

    return (
        <Stack minHeight="110px">
            <Button variant="secondary" onClick={() => setCount(x => x + 1)}>
Increment Count
            </Button>
            <Message show>
                <Content>{count}: We offer the safest space travel of the industry!</Content>
                <Button>Book now</Button>
            </Message>
        </Stack>
    );
};
```

PR #772, opened 2021-11-15
Ensure that the use initial render ref behave properly

alexasselin008:
PR for #740

### DIFF
--- a/packages/components/src/shared/src/useIsInitialRender.ts
+++ b/packages/components/src/shared/src/useIsInitialRender.ts
@@ -1,7 +1,7 @@
 import { useEffect } from "react";
 import { useRefState } from "./useRefState";
 
-export function useIsInitialRenderRef() {
+export function useIsInitialRender() {
     const [isInitialRef, setIsInitial] = useRefState(true);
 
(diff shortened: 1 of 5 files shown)


LaurentDario (approved):
Riding the current!

packages/components/src/shared/src/useIsInitialRender.ts
```diff
@@ -8,5 +8,5 @@ export function useIsInitialRender() {
         setIsInitial(false);
     }, [setIsInitial]);
 
-    return isInitialRef.current;
+    return isInitialRef;
```
patricklafrance, line 11: These kind of changes are dangerous in the long run.

By that, I mean, if you had to design this hook first, you would probably have returned a bool instead of a ref.

Now you have to solve a specific issue somewhere in the code base so you change the API to return a ref instead.

Maybe the problem is somewhere else or there is another solution which doesn't involve reworking the API of a core hook?

alexasselin008: I fixed it another way, by wrapping the useInitialRender in a committedRef.

PR #775, opened 2021-11-16
Move colors doc around and edit names and description

LaurentDario:
This PR moves the Color section of Storybook around to gather the Color Palette and Semantic Colors together under one folder.
Before:
image

After:
image


patricklafrance:
It’s your call but from what I remembered, devs love to use ctrl+f to search in documentation.

By splitting the docs in 2 pages, you make it harder for them to find a color since they must search in 2 pages.

I doubt they really make a difference between color palettes and semantic colors, at least, not until they used Orbit for a while!

My 2 cents ;)

fraincs:
The color variables should never be used from now on, if we want to support a dark theme. I am ok with splitting both as using a color directly is frowned upon.

patricklafrance:
Got it! Still, it’s hard from a developer standpoint to understand what you wrote after reading the documentation since it’ not mentionned anywhere.

I guess at lesst the typescript typings only allow to specify semantic colors? If it does, that’s that.

I wonder though what’s the point of documenting the color palettes if they can’t be used?

fraincs:
I wondered the same regarding documenting colors, this is a in-between state where eventually the color documentation could be removed / merged with the semantic color doc. Since some products won't move to semantic color right away, I'd rather have both documented but slowly deprecate the use of named colors.

Issue #783, opened 2021-11-23
🐛 Spinner should be visible when an async tab is loading


patricklafrance:
## Describe the bug

Spinner should be visible when an async tab is loading

![image](https://user-images.githubusercontent.com/794579/142975373-3945d78f-64de-4add-aaa4-4ad9508e8e66.png)

## Steps to reproduce

Go to: https://deploy-preview-779--sg-storybook.netlify.app/?path=/docs/tabs--default-story

Switch between tabs

fraincs:
fixed on 22.0.1

PR #788, opened 2021-11-23
Bug: AsyncTabs loader not showing

LaurentDario:
Issue: 

## Summary

The loader in the Async Tabs example wasn't showing, as reported in #783 

## What I did

Fixed the colour name used from `--o-ui-alias-low-break` to `--o-ui-b-alias-low-break`.


alexasselin008:
Looking at this PR, i checked if there was other --o-ui-alias variable used, and there is one located in docs.css, could you fix it also please?
image

LaurentDario:
Fixed @alexasselin008 !

Issue #792, opened 2021-11-23
🐛 Docs theme switcher only have a single choice


patricklafrance:
## Describe the bug

Since the theme switcher in Orbit docs only have a single choice, you should consider removing it if you don't intend to support multiple theme for the docs.

![image](https://user-images.githubusercontent.com/794579/143133618-7669941d-e9f0-4756-b83e-12d5ddf77ec7.png)

fraincs:
It makes sense to remove it since it serves no purposes.

Issue #791, opened 2021-11-23
🐛  Secondary Buttons shows artifacts 


fraincs:
## Describe the bug

![image](https://user-images.githubusercontent.com/361632/143097942-b0ae9c72-1a77-42c1-b940-ecbcf39311fc.png)

In certain circumstances, buttons with outlines using box shadows shows some artifacts. I suspect a regression in Chrome engine, yet to be validated.

## Steps to reproduce

https://orbit.sharegate.design/?path=/docs/menu--default-story#default

This seems to only happen on and off. No results are guaranteed. Moving to border should be safe as it's nothing new.

## Expected results

No artifacts are presents.

## Reproducible demo

https://orbit.sharegate.design/?path=/docs/menu--default-story#default

fraincs:
https://lzomedia.com/blog/inset-box-shadow-visual-artifacts-in-google-chrome/ here's an identical issue.

fraincs:
Checkbox and Tag were also problematic

PR #797, opened 2021-11-26
box shadow artifacts fix - removed theme in doc

fraincs:
Issue: 

## Summary

#791 / #792 

## What I did

Every components using box-shadow for a fake border reverted back to using a border instead.

Remove the ThemeSwitcher in the Doc. If we ever have two themes it'll be easy to have it back.

## How to test

- #791 : /?path=/docs/button--default-story#icon /?path=/docs/tag--default-story /?path=/docs/checkbox--default-story
- #792 : Check in the doc top bar that the theme picker is gone

### DIFF
--- a/packages/components/src/button/src/Button.css
+++ b/packages/components/src/button/src/Button.css
@@ -38,10 +38,10 @@
 .o-ui-button:after {
     opacity: 0;
     content: "";
-    top: -1px;
-    left: -1px;
-    right: -1px;
-    bottom: -1px;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
     display: block;
     position: absolute;
     box-shadow: 0 0 0 var(--o-ui-focus-ring-thickness-md) var(--o-ui-focus-ring-color);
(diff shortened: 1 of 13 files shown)


packages/components/src/button/src/Button.css
```diff
@@ -116,10 +116,17 @@ a.o-ui-button {
...
     color: var(--o-ui-text-alias-primary);
 }
 
+.o-ui-button-secondary:after {
+    top: -1px;
+    left: -1px;
+    right: -1px;
+    bottom: -1px;
+}
+
```
LaurentDario, line 129: Just curious about the context of that change. I guess it helps with the artifact issue but do we know what it does?

LaurentDario (approved):

packages/components/src/button/src/Button.css
fraincs: Since we are using a border instead of a box shadow for outline components the focus style needs to move as a border takes space in the layout while a box shadow does not affect an element dimensions.

Issue #801, opened 2021-11-30
🐛 Modals: Focus is not working when focusable elements are hidden under Disclosure


LaurentDario:
The focus sequence seems to break in a modal where focusable elements are hidden.

## Describe the bug

If a Modal contains disclosures with focusable elements (such as Checkbox in the example provided bellow), the focus seems to be blocked just before that hidden element. Best example is with a Disclosure including a focusable element.

This is likely due to the trapFocus trying to reach the hidden element not being able to find it.

Recording: https://share.getcloudapp.com/qGuRGlAY

## Steps to reproduce

Reproducible in the doc of Modal by adding a series of Disclosures with Checkboxes inside of them.

Example of code: 
```
<ModalTrigger>
    <Button variant="secondary">Open</Button>
    <Modal>
        <Heading>Apollo 11 movie</Heading>
        <Content>
            <Paragraph>
                Apollo 11 is a 2019 American documentary film edited, produced and directed by Todd
                Douglas Miller. It focuses on the 1969 Apollo 11 mission, the first spaceflight from
                which men walked on the Moon.
            </Paragraph>
            <Paragraph>
                The film consists solely of archival footage, including 70 mm film previously
                unreleased to the public, and does not feature narration, interviews or modern
                recreations. The Saturn V rocket, Apollo crew consisting of Buzz Aldrin, Neil
                Armstrong, and Michael Collins, and Apollo program Earth-based mission operations
                engineers are prominently featured in the film.
            </Paragraph>
            <Disclosure>
              <Button>Open Disclosure 1</Button>
              <Content>
                <Checkbox key="1_1">Checkbox 1</Checkbox>
                <Checkbox key="1_2">Checkbox 2</Checkbox>
                <Checkbox key="1_3">Checkbox 3</Checkbox>
              </Content>
            </Disclosure>
            <Disclosure>
              <Button>Open Disclosure 2</Button>
              <Content>
                <Checkbox key="2_1">Checkbox 1</Checkbox>
                <Checkbox key="2_2">Checkbox 2</Checkbox>
                <Checkbox key="2_3">Checkbox 3</Checkbox>
              </Content>
            </Disclosure>
            <Select>
              <Item key="select1">Item 1</Item>
              <Item key="select2">Item 2</Item>
              <Item key="select3">Item 3</Item>
            </Select>
        </Content>
    </Modal>
</ModalTrigger>
```
[...]

PR #806, opened 2021-12-01
Bug/801 modals trap focus

LaurentDario:
Issue: #801 

## Summary

The Focus Trap gets locked on any element placed before a hidden focusable element (like a radio button inside of a Disclosure)

## What I did

Adding a check on the elements Trap Focus is going through to make sure they are visible at the time the user tabs through the focusable elements.

Recording: https://share.getcloudapp.com/llukpR2J

### DIFF
--- a/packages/components/src/card/src/Card.css
+++ b/packages/components/src/card/src/Card.css
@@ -3,6 +3,7 @@
     background-color: var(--o-ui-bg-alias-default);
     border-radius: var(--o-ui-shape-rounded);
     border: 1px solid var(--o-ui-b-alias-low-break);
+    height: max-content;
 }
 
 /* SIZES */
--- a/packages/components/src/shared/src/focusableTreeWalker.ts
+++ b/packages/components/src/shared/src/focusableTreeWalker.ts
@@ -1,6 +1,6 @@
 // Tree walker code have been copied from: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/focus/src/FocusScope.tsx.
 
-const FocusableElement = [
+const FocusableElements = [
     "input:not([disabled]):not([type=hidden])",
     "select:not([disabled])",
     "textarea:not([disabled])",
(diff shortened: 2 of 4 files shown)


packages/components/src/card/src/Card.css
```diff
@@ -3,7 +3,6 @@
     background-color: var(--o-ui-bg-alias-default);
     border-radius: var(--o-ui-shape-rounded);
     border: 1px solid var(--o-ui-b-alias-low-break);
-    height: max-content;
```
LaurentDario, line 6: Francis' sneaky fix!

packages/components/src/shared/src/useFocusManager.ts
```diff
@@ -141,18 +145,20 @@ export class FocusManager {
     focusNext({ canFocus, ...options }: FocusOptions = {}) {
         const { elements } = this.scope;
 
+        const visibleElements = this.getVisibileElements(elements);
```
alexasselin008, line 148: Au lieu de filtré les elements localement, on pourrait plutot essayer de fixer le issue à la source. 

L'inspiration de ce concept viens de spectrum. Ils semble ajouter des selectors qu'on a pas, par exemple 

const TABBABLE_ELEMENT_SELECTOR = focusableElements.join(':not([hidden]):not([tabindex="-1"]),');

https://github.com/adobe/react-spectrum/blob/34717616f3a6555d3f4d6739635d3ac347b50a58/packages/%40react-aria/focus/src/FocusScope.tsx#L224


tu peux regarder dans focusableTreeWalker.ts ln 21. Tu vas voir que c'est pas mal semblable à ce que je t'ai envoyé

packages/components/src/shared/src/useFocusManager.ts
```diff
@@ -100,6 +100,10 @@ export class FocusManager {
         return element;
     }
 
+    getVisibileElements(elements: HTMLElement[]) {
+        return elements.filter(x => (x.offsetWidth > 0 && x.offsetHeight > 0));
```
fraincs, line 104: this is a nice approach

packages/components/src/shared/src/useFocusManager.ts
patricklafrance: I agree with Alex.

packages/components/src/shared/src/useFocusManager.ts
patricklafrance: @fraincs haha

packages/components/src/shared/src/focusableTreeWalker.ts
```diff
@@ -16,14 +16,62 @@ const FocusableElement = [
     "[contenteditable]"
 ];
 
-export const FocusableElementSelector = [...FocusableElement, "[tabindex]"].join(",");
+export const FocusableElementSelector = FocusableElements.join(":not([hidden]),") + ",[tabindex]:not([disabled]):not([hidden])";
+FocusableElements.push("[tabindex]:not([tabindex=\"-1\"]):not([disabled])");
```
alexasselin008, line 20: I would change the synthax here so its more like before, instead of the spectrum synthax

Issue #811, opened 2021-12-06
🐛  Autocomplete color is not dark mode friendly


fraincs:
## Describe the bug

When autofill is applied to a field, Chrome applies a specific color, this is good in light mode but does not behaves well in dark mode as it's too light.

![image](https://user-images.githubusercontent.com/361632/144867958-a7ae34fd-2071-479d-b22f-57562b789ea1.png)

## Steps to reproduce

Add autocomplete="name" to a text input, pick your name.

## Expected results

The autofill background color hint is darker.

This is a possible hint toward a fix: 

https://stackoverflow.com/questions/2781549/removing-input-background-colour-for-chrome-autocomplete

fraincs:
Fixed on https://github.com/gsoft-inc/sg-orbit/releases/tag/22.0.3

Issue #810, opened 2021-12-02
🚀  Handle breaking words in a more sensitive way


fraincs:
## Description

Using overflow-wrap: break-word; on paragraph and headings would be a more sensitive way to cut overflowing text.

without it 
![image](https://user-images.githubusercontent.com/361632/144502605-6e10a6ba-5b8d-4031-b7da-8773e7c49bb2.png)

with the property
![image](https://user-images.githubusercontent.com/361632/144502678-a36565e9-141b-4acd-b1cf-50942f9887d7.png)

fraincs:
Fixed on https://github.com/gsoft-inc/sg-orbit/releases/tag/22.0.3

Issue #805, opened 2021-12-01
🐛 ButtonAsLink should be exported with the button slot


alexasselin008:
## Describe the bug

A card with a button, that works
```
<Card>
    <Heading>NASA Headquarters</Heading>
    <Header>No unexpected visitors allowed.</Header>
    <Content>
        <Paragraph>
            NASA Headquarters, officially known as Mary W. Jackson NASA Headquarters or NASA HQ and
            formerly named Two Independence Square, is a low-rise office building in the
            two-building Independence Square complex at 300 E Street SW in Washington, D.C.
        </Paragraph>
    </Content>
    <Button variant="secondary">Plan a visit</ButtonAsLink >
</Card>
```

A card with a buttonLink, wich doesnt work
```
<Card>
    <Heading>NASA Headquarters</Heading>
    <Header>No unexpected visitors allowed.</Header>
    <Content>
        <Paragraph>
            NASA Headquarters, officially known as Mary W. Jackson NASA Headquarters or NASA HQ and
            formerly named Two Independence Square, is a low-rise office building in the
            two-building Independence Square complex at 300 E Street SW in Washington, D.C.
        </Paragraph>
    </Content>
    <ButtonAsLink variant="secondary">Plan a visit</ButtonAsLink >
</Card>
```

## Expected results

The second scenario should work without having to specify a slot

kingbaub3:
Similar to this issue, the following code works with a Button but doesn't work with a ButtonAsLink.

Button:
![image](https://user-images.githubusercontent.com/6119223/144907550-3466b0b8-74d9-4eec-8b50-ace36f88b995.png)

ButtonAsLink:
![image](https://user-images.githubusercontent.com/6119223/144907637-05b11b01-ddd6-407a-8d96-a8c2a3ac4708.png)

patricklafrance:
This issue makes me wonder if providing the ability to have a link which look like a button was a good idea?

From a user standpoint, if you see something which looks like a button, you'll expect an action. But in this case, once you click on the button you will navigate somewhere.

This is not a good user experience has UI should behave the way user expect them to behave.

Beware of this pattern and maybe try to raise this concern to your designer before implementing the design?

At least, something similar to the `<Link />` component `external` prop should be added to `<ButtonAsLink />` component so the link will be opened in a new tab and the `rel` prop would be set to "noopener noreferrer"

alexasselin008:
I was thinking that too for the external props, it seems like a good idea. 

[...]

Issue #804, opened 2021-12-01
🐛 Cards do not stretch vertically


fraincs:
## Describe the bug

When side by side in a flex container with the align-items: stretch; css property applied, Cards do not stretch.

## Steps to reproduce

Use case: 

```
    <Inline alignItems="stretch">
        <Card size="xs">
            <Heading>NASA Headquarters</Heading>
            <Content>
                300 E. Street SW, Suite 5R30
                <br />
                Washington, DC 20546
                <br />
                (202) 358-0001 (Office)
            </Content>
        </Card>
        <Card size="sm">
            <Heading>NASA Headquarters</Heading>
            <Content>
                300 E. Street SW, Suite 5R30
                <br />
                Washington, DC 20546
                <br />
                (202) 358-0001 (Office)
                <br />
                (202) 358-4338 (Fax)
            </Content>
        </Card>
        <Card>
            <Heading>NASA Headquarters</Heading>
            <Content>
                300 E. Street SW, Suite 5R30
                <br />
                Washington, DC 20546
                <br />
                (202) 358-0001 (Office)
                <br />
                (202) 358-4338 (Fax)
            </Content>
        </Card>
    </Inline>
```

## Expected results

The first card stretches vertically to be the same height of the other cards.

fraincs:
Fixed on https://github.com/gsoft-inc/sg-orbit/releases/tag/22.0.3

PR #813, opened 2021-12-07
Feature/bundle issues

fraincs:
Issue: 

## Summary

#805 / #804 / #810 / #811 

## What I did

Fixed above issues


alexasselin008 (approved):

packages/components/src/input/src/Input.css
```diff
@@ -34,11 +34,20 @@
...
     background-color: var(--o-ui-bg-alias-input-selection);
 }
 
+/* AUTOFILL */
+input:-webkit-autofill,
+textarea:-webkit-autofill,
+select:-webkit-autofill {
+    -webkit-box-shadow: 0 0 0 1000px var(--o-ui-bg-alias-accent-faint) inset !important;
+    -webkit-text-fill-color: var(--o-ui-text-alias-primary) !important;
+}
```
patricklafrance, line 49: Your selectors are too generic, you should only match orbit elements meaning, `.o-ui-input` selectors.

fraincs: Totally, copy pasta gone wrong

patricklafrance: Also, it's not a big deal but, adding code for "select" in this file is not great from a maintenance standpoint. This file is scope to text inputs.

fraincs:
Yup it has been removed.

PR #822, opened 2021-12-13
Focus management improvement

patricklafrance:
## Summary

Improved the focus management code.  Haven't been able to achieve everything I wanted. My goal was to consolidate all the focus code within a focus manager but `useRestoreFocus` still need a dependency to a tree walker because it has to search at the document body level.

Also fixed the restore focus of `Popover` which was broken, even in prod.

Anyway, I think the focus management is now much easier to understand. I split the `FocusManager` into a few pieces... A FocusScopeIterator, which is reusable, a DomFocusManager and a VirtualFocusManager. 

Also added a lot of comments in `useRestoreFocus` and another 50 Jest tests.


packages/components/src/overlay/src/useRestoreFocus.ts
```diff
@@ -26,36 +30,43 @@ export function useRestoreFocus(scope: DomScope, { isDisabled }: UseRestoreFocus
...
+                        // Must prevent the browser from doing is thing since we will takeover and handle the tab ourself.
                         event.preventDefault();
 
                         if (!isNil(nextElement)) {
+                            // If we found a tabbable element, focus the element and exit.
+                            // This is a scenario were we restored the tab order. Instead of focusing the first element available under the overlay, we focus the element next to the overlay trigger.
                             nextElement.focus();
                         } else {
+                            // If we can't find any tabbable element to focus, restore the focus on the overlay trigger element.
+                            // This is also a scenario were we restored the tab order. Instead of focusing the first element available under the overlay, we made sure to focus back to the overlay trigger.
```
LaurentDario, line 69: small type: Where

LaurentDario (approved):
Just a tiny typo, other than that it makes it clearer so can't complain! I'll do some quick QA today of the different focus cases we were covering to make sure it's all gucci

packages/components/src/overlay/src/useRestoreFocus.ts
patricklafrance: > Just a tiny typo, other than that it makes it clearer so can't complain! I'll do some quick QA today of the different focus cases we were covering to make sure it's all gucci

Perf, let me know if you find something.

PR #830, opened 2021-12-17
tile focus position / loader was off centered / card fluid image/illu…

fraincs:
Issues: 

- Loader was not centered in the loading tab sample
- Card's images were not the right width when in a fluid Card.
- Tile focus position was not right.

## What I did

- Added a new loader animation as using the button one made no sense in a Tab sample
- When fluid the image/illustration will be 8rem like every other cards.
- Focus pseudo element is positioned 1px  negatively which is the width of the element's border.


packages/components/src/tabs/docs/loader.css
```diff
@@ -17,6 +17,15 @@
     border-radius: 50px;
     border: 2px solid var(--o-ui-b-alias-low-break);
     border-top-color: var(--o-ui-white);
-    animation: o-ui-button-spinner 1s linear infinite;
+    animation: o-ui-sb-spinner 1s linear infinite;
     transform: translate(-50%, -50%) rotate(0deg);
 }
+
+@keyframes o-ui-sb-spinner {
```
patricklafrance, line 24: is this spinner specific to a tab docs example or any other exemple in the docs could use it?

fraincs: Any other examples could use it

fraincs: I guest this could be moved then

patricklafrance: yeah should go in `docs.css`

patricklafrance (approved):

Issue #861, opened 2022-01-11
🐛 Message close button overflow outside it's container when it's too small


patricklafrance:
## Describe the bug

A message close button overflow outside it's container when it's too small

![image](https://user-images.githubusercontent.com/794579/148887758-a2484c9a-ac85-4680-bb9a-002bde408074.png)

Issue #856, opened 2022-01-11
🐛 ButtonGroup should wrap on multiple lines before breakinging text


patricklafrance:
## Describe the bug

ButtonGroup should wrap on multiple lines before breakinging text.

![image](https://user-images.githubusercontent.com/794579/148885107-77a1e463-bbc0-4043-a921-c6c97c3f2dc8.png)

fraincs:
Button group should wrap by default.

Issue #838, opened 2021-12-23
🐛 Doc: background colours are missing Alias


LaurentDario:
3 BG colours are missing `alias` in their var name.
![image](https://user-images.githubusercontent.com/234418/147261666-012f31ac-9784-4ba9-a850-859af914d243.png)

`soft-break` `mid-break` & `hard-break`

Issue #823, opened 2021-12-13
🐛 Button content should not wrap in messages


alexasselin008:
## Describe the bug

![image](https://user-images.githubusercontent.com/38871812/145837991-1219fef1-09ec-494a-afd5-cf418bd5b40f.png)

Button Got it is wrapped on 2 lines

## Steps to reproduce
https://orbit.sharegate.design/?path=/docs/message--default-story#rich-content

```
<Message>
    <Content>
        <Heading>There are no more activations left!</Heading>
        Need to use ShareGate on a different workstation? Remove the activation below <span role="img" aria-label="" aria-hidden="false" class="o-ui-html-span">👇</span> then log into the migration tool with your ShareGate account. The license will be automatically activated on that workstation.
    </Content>
    <Button>Got it</Button>
</Message>
```

## Expected results

Got it should be on the same line

patricklafrance:
I am not sure if it's the button text who shouldn't wrap. We can't know how long the button text could be. I agree for a short text it's weird but what about if the text "Click me for a supper dupper offer" ?

Something have to be done but I would also investigate other options.

alexasselin008:
I would prefer to have to specify on my button that it should wrap when it's too long, rather than specifying on all buttons in message that i don't want them to wrap

fraincs:
One option would be to always reserve a spot for a button, the content length is too long anyway (The ideal width of a paragraph is 40-55 characters) it's currently way more in this example. Also I am wondering what would happen if there's 2 messages and both with a different button, the ending of the lines would be non identical.

patricklafrance:
> I would prefer to have to specify on my button that it should wrap when it's too long, rather than specifying on all buttons in message that i don't want them to wrap

Adding this kind of property on a Button is, in my opinion dangerous. Eventually, with this philosophy, you'll end up with a bunch of incoherent properties on the component to maintain.

I prefer a solution similar to what Francis propose. It needs to work for all the breakpoints though.

fraincs:
![image](https://user-images.githubusercontent.com/361632/146070643-1e97b039-0fdc-4eb8-a3b5-fb2c1016dea6.png)
Here's an option I've seen in another Design System, realistically a button on two lines is not very great looking design wise, limiting the buttons dimensions and adding ellipsis if needed would be a graceful way to solve the problem.

patricklafrance:
[...]

PR #863, opened 2022-01-12
#823 / #838 / #856 / #861

fraincs:
Issue: 

## Summary

#823 - Button content should not wrap in messages
#838 - Doc: background colours are missing Alias
#856 - ButtonGroup should wrap on multiple lines before breaking text
#861 - Message close button overflow outside it's container when it's too small

### DIFF
--- a/packages/components/src/button/src/ButtonGroup.tsx
+++ b/packages/components/src/button/src/ButtonGroup.tsx
@@ -39,6 +39,7 @@ export function InnerButtonGroup(props: InnerButtonGroupProps) {
         children,
         disabled,
         fluid,
+        wrap = true,
         forwardedRef,
         orientation = "horizontal",
         size,
@@ -66,7 +67,8 @@ export function InnerButtonGroup(props: InnerButtonGroupProps) {
                     gap: Gap[orientationValue][normalizeSize(sizeValue)],
                     orientation,
                     ref: forwardedRef,
-                    role: !isInField ? "group" : undefined
+                    role: !isInField ? "group" : undefined,
+                    wrap
                 }
             )}
         >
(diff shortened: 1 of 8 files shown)


packages/components/src/button/src/ButtonGroup.tsx
```diff
@@ -39,6 +39,7 @@ export function InnerButtonGroup(props: InnerButtonGroupProps) {
         children,
         disabled,
         fluid,
+        wrap = true,
```
LaurentDario, line 42: I might be missing some context so please forgive me if I'm not understanding it correctly!
We had this discussion a long time ago also with Alex, but I'm not a big fan of booleans defaulting to `true` in props.

That means that if I want the `InnerButtonGroup` to be no wrap, I don't to specifically say `wrap={false}`. In other words: `<InnerButtonGroup wrap={false}>` works but `<InnerButtonGroup wrap>` is useless.

In that case would it be ok to with `noWrap = false` here, and on line 71 something along the lines of `wrap: !noWrap`.
Or maybe `wrap` (with no default set on line 42) and, on line 71, `wrap: noWrap ? undefined : true`.

Opening this up for discussion, ultimately the solution you have now works just fine so not bothered if we go with it!

LaurentDario (approved):
Just one tiny comment

packages/components/src/button/src/ButtonGroup.tsx
fraincs: I am open to any of these solutions, what do you think @alexasselin008

alexasselin008: In this case, the props with the name wrap are in multiple components. Therefore, i wouldnt not change the name, even if i agree with what you said about true default values.

Issue #874, opened 2022-01-18
🚀  Border Radii should be exposed


fraincs:
## Description

Border radii are not available as tokens like spacing / sizing tokens, it would be useful to have access to the same values as the components.

E.g. pill / rounded 

## Checklist

- [x] Chromatic tests
- [x] Jest tests
- [x] Documentation
- [x] Figma

Issue #873, opened 2022-01-17
🐛 Font size alias headline and subheadline should be renamed


fraincs:
## Describe the bug

Font size aliases follow a numeric logic except for headline and subheadline, this was to follow Tachyons way of naming things. Renaming them to 10 and 11 would be more consistent.

Issue #842, opened 2022-01-06
💡 Documentation materials page should be merged in a single "Tokens" page.


patricklafrance:
## Problem to solve

It's hard right now for a documentation reader to quickly find all the CSS Variables and Design Assets available in Orbit.

Furthermore, our current "Materials" pages offer a lot of information which are useless for a developer. Our original idea was that the documentation should be for both developers and designers but let's be honest in the end it's only for the developers.

## Idea

A "Tokens" page could offer a centralized and more concise documentation experience. Something similar to https://paste.twilio.design/tokens

A token entry could include: **Style Prop Value (when available)** | **CSS Variable** | **Value**  | **Exemple**

Maybe the word "token" is not appropriate since we don't have actual tokens in Orbit but I couldn't find a better name.

Once we have this page, maybe we could remove the "Reference" section of the style props page.

fraincs:
> Once we have this page, maybe we could remove the "Reference" section of the style props page.

I am unsure about this one as the reference tables are providing more information like if a styleProps supports pseudo states, the CSS Property it affects, etc.

PR #882, opened 2022-01-21
Feature/842

fraincs:
Issue: 

#842 - 💡 Documentation materials page should be merged in a single "Tokens" page.
#874 - Border Radii should be exposed.
#873 - Font Size Headline and subheadline should be renamed.

## Summary

Moved Spacing, Sizing, Colors, Typography and Shadows pages to a new token page.
Exposed and renamed some tokens.

## What I did

- Deleted Spacing, Sizing, Colors, Typography and Shadows pages
- Modified SemanticColorTable(now named TokenTable) to accommodate the new tokens tables
- Renamed Headline and subheadline to 10 and 11
- Exposed border radius shapes


patricklafrance:
I don't think we should have a column per color scheme for the value. The value should change bases on the active color scheme.

Otherwise everytime we had a new color scheme we have to add a new column. It doesn't scale well.

patricklafrance:
I personally find it easier to read with a font-size of "fs-3" like we used to have for every table.

patricklafrance:
I personally wouldn't change the background color for static color like white:

image

It's static, you should expect to see nothing on a light color scheme, switch to dark if you want to see.

patricklafrance:
I don't feel the need to box the values for line-height. Pretty sure if someone want to see more about the line-height he/she will open dev tools and have a look.

docs/features/tokens/Tokens.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+# Tokens
+
+Tokens are Orbit's reusable values, providing a common language between developers and designers. These can be used as `CSS` variables or [Style Props](/docs/style-props--page) values.
+
+Aliased tokens are named for their use case, rather than their value. For example `alias-alert` is named to indicate the appropriate background color for an error message background, as opposed to `red-5` which is less specific and could be used in a number of ways depending on context.
+
+This makes their intended use clear and intentional, and allows us to develop a scalable and consistent visual language while also being the building block of Orbit's [theming](/docs/theming--page).
+
+### Usage
```
patricklafrance, line 21: Should he h2

docs/features/tokens/Tokens.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+#### As Style Props
```
patricklafrance, line 23: I don't think it should be an header.

docs/features/tokens/Tokens.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+```jsx
+    <Div backgroundColor="alias-default">...</Div>
+```
+
+#### As CSS Variable
```
patricklafrance, line 29: I don't think it should be an header.

docs/features/tokens/Tokens.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+        { token: "alias-alert", variable: "--o-ui-icon-alias-alert", valueLight: "#D93C37", valueDark: "#F88F8D", itemRenderer: iconRenderer },
+        { token: "alias-warning", variable: "--o-ui-icon-alias-warning", valueLight: "#F6BA42", valueDark: "#F6BA42", itemRenderer: iconRenderer },
+        { token: "alias-success", variable: "--o-ui-icon-alias-success", valueLight: "#2D8739", valueDark: "#17A93C", itemRenderer: iconRenderer },
+        { token: "alias-static-white", variable: "--o-ui-icon-alias-static-white", valueLight: "#FFFFFF", valueDark: "#FFFFFF", itemRenderer: iconRenderer }
+    ]}
+/>
+
+## Font Sizes
+
+> In Orbit, **1 rem unit = 16 pixels**.
```
patricklafrance, line 120: I would display this info at the top of the page liste Paste did because it's not only font sizes which use rem. For example, Radii also use rem values.

patricklafrance:
Seems to be an issue with this one:

image

patricklafrance:
Should the exemple  be vertically centered?

image

docs/features/tokens/Tokens.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+import { Table } from "@stories/components";
+
+<Meta
+    title="Features/Tokens"
+    id="tokens"
+/>
+
+# Tokens
+
+Tokens are Orbit's reusable values, providing a common language between developers and designers. These can be used as `CSS` variables or [Style Props](/docs/style-props--page) values.
```
patricklafrance, line 15: I think you could rework a little bit this text to move "These can be used as `CSS` variables or [Style Props](/docs/style-props--page) values" in the usage section since this is describing the usage.

patricklafrance (changes requested):

fraincs:
> I don't think we should have a column per color scheme for the value. The value should change bases on the active color scheme.
> 
> Otherwise everytime we had a new color scheme we have to add a new column. It doesn't scale well.

Realistically how many themes are we going to have? The two column choice was one to expedite the release and avoid some coding as well as make it easier to `ctrl + f` a value. Paste way of doing it works since they have a search input in the page.

fraincs:
> It's static, you should expect to see nothing on a light color scheme, switch to dark if you want to see it.

Static white would never be on a white background but a darker colour, this shows a real use case versus questioning if there is a render bug or not.

fraincs:
> I don't feel the need to box the values for line-height. Pretty sure if someone want to see more about the line-height he/she will open dev tools and have a look.

image

without them it's hard to see the difference

fraincs:
> I personally find it easier to read with a font-size of "fs-3" like we used to have for every table. I never understood why it's seems right to ask a reader to squint his eyes because it's a list.

We are starting to be tight on space, taking example on others they mostly go for 14px even 12px for these tables.

fraincs:
> I don't think we should have a column per color scheme for the value. The value should change bases on the active color scheme.
> 
> Otherwise everytime we had a new color scheme we have to add a new column. It doesn't scale well.

Ended up coding it.

fraincs:
> > I personally find it easier to read with a font-size of "fs-3" like we used to have for every table. I never understood why it's seems right to ask a reader to squint his eyes because it's a list.

Went back to the original font size

patricklafrance (approved):

PR #883, opened 2022-01-21
First exploration into a solve for the Popover issue with trapfocus

LaurentDario:
The inspiration: https://github.com/Shopify/polaris-react/pull/2681/files#diff-56933e8df958aa3996a19f9c20624f35838f1ab930bd862836dff77602609bc3R27

I think the idea is here but the execution isn't right.
The idea would be to keep track of which trap focus are ongoing and if there's already one, avoid trying to trap focus again until the focus is "released". But in the current case I'm never removing the focus trap from the list and rely on it re-initializing, not sure this is viable.

Issue: 

## Summary

Opening 2 Popover at the same time causes an error

## What I did

List the focus traps in `trapFocusList` to keep track of when the focus trap is already triggered.


patricklafrance:
This is a good lead.

However, I think it ignores the real issue. I would try to figure out what's going on with `useTrapFocus` before going this way. In Orbit an overlay have the capability to close on focus leave so opening a second Popover (by clicking on a trigger) should automatically close the first one. Therefore you should never have more than one Popover opened.

Since every Overlay components use it's own FocusManager I am not sure how it would know that multiple overlay components are opened. It would require the FocusManager to be defined at the root of the App, for example with a Context.

I think Orbit FocusManager and Polaris FocusManager implementations have different intent. In Orbit, it's an helper to facilitate moving the focus around in a given scope.

Your code fix the problem but I am not sure if it's intentional. By this, I mean that the idea behind your fix is to have a centralized FocusManager who's aware of which overlay component is currently opened so the `useTrapFocus` hook could have access to this information and take a decision based on this information. It does work but since the FocusManager instances are specific to a single overlay component a FocusManager can only know information about one overlay component which kind of invalidate the idea behind the fix.

Based on this, I guess some kind of flag in `useTrapFocus` could fix the problem but it's hard to tell without knowing exactly what's going on with `useTrapFocus.`.

Before trying to apply a fix, I think it would be a good idea to understand what's going on.

patricklafrance:
Let me know if I haven't correctly understood the fix.

LaurentDario:
I'm still digging into this problem but based on a few tests it looks like the `focusin` event is triggered from the second Popover to the first. So we end up on a loop that hides the 1st Popover but then focuses in it, then trigger the trigger button of the second one. I'm having a hard time following the trail of where all these events are going.

patricklafrance:
@LaurentDario I think the following code fix the issue.

```
    // If a focus event occurs outside the scope (e.g. user tabs from browser location bar),
    // restore focus to the previously focused node or the first tabbable element in the active scope.
    const handleFocus = useEventCallback((event: FocusEvent) => {
        console.log("useTrapFocus - handleFocus", event);

        // When then focus was previously outside of the viewport (like the browser localtion bar),
        // the relatedTarget (aka the previous focused element) will be null.
        if (isNil(event.relatedTarget)) {
            const target = event.target as HTMLElement;

            if (!focusManager.isInScope(target, { includeChildScopes: true })) {
                if (!isNil(focusedElementRef.current)) {
                    focusedElementRef.current.focus();
                } else {
                    const element = focusManager.focusFirst({ tabbableOnly: true });
                    setFocusedElement(element);
                }
            } else {
                setFocusedElement(target);
            }
        }
    });
```

The addition is `if (isNil(event.relatedTarget))  { ... }`

The problematic code is the code which ensure that the focus is moved back into the traped focus scope when the user switch back to the viewport from the outside (like from the browser location bar).

This fix detect when the focus is coming back into the viewport from the outside by leveraging the fact that the `relatedTarget` will be `null`.

It seems to work fine.

The main usecase for this feature in trap focus is for a dialog... 

Without the code, you can see that when the focus move back to the viewport, the dialog trigger is focused instead of the button in the dialog content:

without_code

With the code:

with_code

LaurentDario:
Amazing!

I'm trying to understand why the solve works, does this mean that when the Popover was closing, it would focus back outside of the page and then try to focus back in onto the trigger of the first Popover instead of the new Popover that just opened?

patricklafrance:
> Amazing!
> 
> I'm trying to understand why the solve works, does this mean that when the Popover was closing, it would focus back outside of the page and then try to focus back in onto the trigger of the first Popover instead of the new Popover that just opened?

No, well there was 2 cases:

1- One of the issue was that the Popover wasn't closing at all when the focus moved to another component (for example a Button). When the focus was moving to another component, the `handleFocus` code was interferring and moving the focus back to the Popover. This was preventing the Popover from closing.

You can replicate the issue with the following code:

```
        <Flex>
            <PopoverTrigger>
                <Button variant="secondary">Toggle</Button>
                <Popover>
                    <Heading>Space News</Heading>
                    <Content>SpaceX designs, manufactures, and launches the world's most advanced rockets and spacecraft.</Content>
                </Popover>
            </PopoverTrigger>
            <Button marginLeft="100px" variant="secondary">Test</Button>
        </Flex>
```

2- The second issues was when having multiple Popover rendered on the same page. When the first Popover is rendered, the autofocus and the trap focus code if the first Popover is executed. So the first Popover have focus. Then the second Popover is rendered and his autofocus and trap focus code is also executed. The second Popover try to get focus (autofocus) but the first Popover trap focus code take it back. Then the second Popover trap focus code also take it back. You end up with an infinite loop.

You can replicate the issues with the following code (WARNING, YOUR BROWSER TAB WILL FREEZE):

```
```
        <>
                <Popover>
                    <Heading>Space News</Heading>
                    <Content>SpaceX designs, manufactures, and launches the world's most advanced rockets and spacecraft.</Content>
                </Popover>
                <Popover>
                    <Heading>Space News</Heading>
                    <Content>SpaceX designs, manufactures, and launches the world's most advanced rockets and spacecraft.</Content>
                </Popover>
        </>
```
```

This was all due to the behavior of the code in `handleFocus`. By scoping the execution of the code in `handleFocus` to execute only when the focus come back from the outside, it solve the problem since it won't execute for the 2 cases I explained.

patricklafrance:
@LaurentDario I think you can remove the failing test.

Also, could you please keep the comment, I think future us will like it.

```
        // When then focus was previously outside of the viewport (like the browser localtion bar),
        // the relatedTarget (aka the previously focused element) will be null.
```

patricklafrance (approved):

Issue #876, opened 2022-01-18
🐛 Missing box-sizing on custom component as Button


LaurentDario:
When using the `as` function to render a non-Orbit component as a Button (in that example, a React Router Link), the box-sizing is left to default `content-box`, making the button larger than it should be.

## Describe the bug

After removing Tachyons (which was forcing the rule `box-sizing: border-box` for `*, :after, :before`) we noticed that the size of buttons is not consistent.
![image](https://user-images.githubusercontent.com/234418/150028519-306b8a93-aab8-416f-8b43-1be45da92e34.png)

The button being fluid, the missing box-sizing forces it out of the container.

## Steps to reproduce

We use a `NavHashLink` as button:
```
const NavHashLinkAsButton = as(Button, NavHashLink);
```

Contrary to a Link as Button, it does not get added the `o-ui-html-a` class adding the right `box-sizing`.

## Expected results

Not sure if this falls onto Orbit or if it should be handled by the consumer.
If it should be handled by the consumer, a `boxSizing` prop could be added to the styled system.

patricklafrance:
I think it should be handled by Orbit, what do you think @fraincs ?

Is is intentional?

fraincs:
It is not intentional, the issue seems to be due to the fact that Orbit can't detect the NavHashLink is an `a` element. Should Orbit aware of third parties?

patricklafrance:
@fraincs how would you do this?

alexasselin008:
We have an Orbit Button that renders anything with the as property. Instead of trying to detect the anything and apply a box sizing to that, should we add the box sizing to all the orbit components directly?

if o-ui-html-button has box sizing set, why wouldnt o-ui-button have it?

fraincs:
> @fraincs how would you do this?

I was not implying it was possible.

fraincs:
> 

We were aiming to only set the box sizing on html elements and not other classes, in theory all Orbit components should use Orbit html elements, in this particular situation the rendered `<a>` is not an Orbit `a` element since it's provided by React Router.

patricklafrance:
I think @fraincs nailed the problem and it's out of scope of Orbit.

patricklafrance:
2- Wild idea...

1- I think we could refactor the "default style" classes to be `p.o-ui-html` instead of `o-ui-html-p`. Therefore, we could somehow add a "o-ui-html" class to an external component and it would benefit from orbit default styles. Maybe it could be added by the "as" function.

[...]

Issue #867, opened 2022-01-14
💡 Storybook doc: Only disable one of the buttons in Alert Disabled button example


LaurentDario:
In storybook we're offering an example for Disabled Buttons, in Alerts. This example blocks the user in an Alert they can't dismiss because all buttons are disabled.

![alert-stuck](https://user-images.githubusercontent.com/234418/149545105-96d62daa-1338-4a46-b851-e196b220ed8d.gif)

## Idea

We could limit the disabled buttons to 1 of the 2 buttons. Or we could offer two examples with one disabled button in each example (to show that the user can disable any of the 2 buttons).

patricklafrance:
Yeah!

PR #885, opened 2022-01-23
Add a "normalizeStyles" option to the "as" function

patricklafrance:
Issue: https://github.com/gsoft-inc/sg-orbit/issues/876 https://github.com/gsoft-inc/sg-orbit/issues/867

## Summary

When using the `as` function with a third party component there is currently no way to apply the default Orbit element styles to the third party component.

## What I did

My initial though was to introduce a `toOrbit(component)` function. This function would add Orbit style props capabilities and Orbit default element styles to the third party component.

From a library standpoint it seemed like a good idea. However, I don't think ShareGate have any use cases for this so it was adding useless complexity. Please correct me if I'm wrong and we do have use cases for this @LaurentDario, @alexasselin008.

The solution implemented in this PR is to extend the `as` function to accept a `normalizeStyles` option which would apply all Orbit default element styles on the resulting component.

```jsx
const NavHashLinkAsButton = as(Button, NavLink, { normalizeStyles: true });
```

This is something the `as` function could have done by default but I think most of the usage for the `as` function are only using Orbit components so I prefered to make it opt-in.

## How to test

- Is this testable with Jest or Chromatic screenshots? No

- Does this need an update to the documentation? Yes


patricklafrance:
@fraincs the new html elements CSS breaks the specifity because since it introduced a type selector...

So I introduced a [:where pseudo selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:where).

```
button:where(.o-ui-element) {
    background-color: transparent;
    background-image: none;
    cursor: pointer;
}
```

I don't think we care for the "Support for forgiving" so according to caniuse, it seems well supported: https://caniuse.com/?search=%3Awhere

However, the problem occurs because Orbit default element styles try to normalize all the elements. I am not sure if this is something we should be.

IMO, every elements which are not rendered by an Orbit component should use the browser default style. I don't consider a "<A>", "<Span>", "<HtmlButton>", ... Orbit components.

So I would basically remove everything from "html.css" and only keep:

```
.o-ui-element,
.o-ui-element::after,
.o-ui-element::before {
    box-sizing: border-box;
}
```

I guess there is a reason why you did this? Do you remember why?

fraincs:
> @fraincs the new html elements CSS breaks the specifity because it introduced a type selector...
> 
> So I added a [:where pseudo selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) to reset the specifity to 0.
> 
> ```
> button:where(.o-ui-element) {
>     background-color: transparent;
>     background-image: none;
>     cursor: pointer;
> }
> ```
> 
> I don't think we care about the "forgiving" feature of `:where` so according to caniuse, it seems well supported: https://caniuse.com/?search=%3Awhere
> 
> However, the problem occurs because Orbit default elements styles try to normalize all the default element style. I am not sure if this is something Orbit should do.
> 
> IMO, every elements which are not rendered by an Orbit component should use the browser default styles. I don't consider a "A", "Span", "HtmlButton" ... Orbit components.
> 
> So I would remove everything from "html.css" and only keep:
> 
> ```
> .o-ui-element,
> .o-ui-element::after,
> .o-ui-element::before {
>     box-sizing: border-box;
> }
> ```
> 
> I guess there is a reason why you did this? Do you remember why?

More of a defensive approach, removing these would need some additional CSS in components, which I am not against, but would delay the merge.

patricklafrance:
@fraincs got it, let's just keep this in mind that we might have an issue here.

fraincs:
> got it, let's just keep this in mind that we might have an issue here.

By issue you mean by not removing the other normalize classes?

patricklafrance:
@fraincs yes

Issue #868, opened 2022-01-15
❓ Alignement of a popover X button


patricklafrance:
When I see it live, I am not sure about the position of the dismiss button of a popover?

![image](https://user-images.githubusercontent.com/794579/149603539-e5451c7f-c08f-4da0-b39f-f4a6ce5e4bc8.png)

Personally, I would align the end of the X with the end of the "Got it" button. And I would add more space over the dismiss button.

fraincs:
@Chicoalvi thoughts?

Chicoalvi:
Currently, the best alignment rendering (which I agree with) is with a margin-top = 8px and margin-right= 8px with a cross Button `<CrossButton aria-label="Clear"

<img width="1680" alt="Capture d’écran, le 2022-01-18 à 14 52 06" src="https://user-images.githubusercontent.com/16927274/150009121-6fc1d16a-752c-4acc-a377-c0cd0c06b3da.png">
`

patricklafrance:
8px looks good to me!

fraincs:
Do you have an example with a button? How would the title react if too long?

Chicoalvi:
it looks like this 

<img width="611" alt="Capture d’écran, le 2022-01-18 à 16 18 19" src="https://user-images.githubusercontent.com/16927274/150020390-d18ab301-3792-475a-9066-c037812a7e65.png">

Chicoalvi:
With a long title
I put a 2 rem margin-right on the title's div

<img width="497" alt="Capture d’écran, le 2022-01-18 à 16 35 30" src="https://user-images.githubusercontent.com/16927274/150022548-388ee959-f2f6-47cf-8d39-e3ef2ee4f4c7.png">

On the side of the design system of Salesforce, they have privileged a margin around the cross Button

<img width="1180" alt="Capture d’écran, le 2022-01-18 à 16 29 11" src="https://user-images.githubusercontent.com/16927274/150023283-d920d657-4c1c-4a80-be7d-96b094da80a5.png">

fraincs:
LGTM

Issue #862, opened 2022-01-11
❗ Badge should allow different background colors


alexasselin008:
## Description

When rendering a Badge with the Dot variant. The idea is to quickly indicate to the user that something new is available. However, the gray "hard-break" color doesn't seem an appropriate color to capture the user's attention. We should be able to customize the
dot color, or have a set of defined colors we can use.

Exemple : 
```
        <ToggleIconButton aria-label="Whats News" variant="tertiary">
          <Badge slot="icon" variant="dot" overlap="icon">
            <NotificationIcon />
          </Badge>
        </ToggleIconButton>
```
![image](https://user-images.githubusercontent.com/38871812/148952561-7c045716-6518-46b5-8357-04bcbde12c42.png)
![image](https://user-images.githubusercontent.com/38871812/148952573-f0a9376f-11de-4b1b-a312-bc1b4097ab80.png)

fraincs:
Something to keep in mind is that a badge can contain text, in order to keep a good contrast we should limit the options or it can lead to some accessibility issues.

patricklafrance:
Badges used to be red to get the user attention but it's been updated to gray with the new DA. Maybe the solution is not to make the backgroundColor customizable but instead bring this issue to the designers and ask for a new color?

I am not sure what use cases would required to have customizable backgroundColor for badges. IMO, the issue here might just be that the current backgroundColor is not the right one. The badge backgroundColor should be standardize accross an app.

If we do want to allow custom backgroundColor then there are a few options...

The rendered HTML is the following:

```
<div class="o-ui-badge o-ui-badge-dot o-ui-badge-over-circle o-ui-html-div">
        <div class="o-ui-badge-anchor o-ui-html-div"></div>
        <div class="o-ui-html-div"></div>
</div>
```

Right now we will render the additional props (e.g styling props) on the outer element with the `o-ui-badge` class. Which is as expected. What we want is to have a way to provide style props to the inner element with the `o-ui-badge-anchor` class.

1- The obvious one is to forward the backgroundColor prop (and maybe the color prop) to the inner component. It would work but I ususally don't like forwarding specific props. Things could get messy and hard to understand from a consumer standpoint. I tend to move away from this when there are alternative solutions.

2- We could update the Badge component to always receive as children the actual badge component AND the component to anchor to. It's already happenning for an icon badge and kind of happening for a text badge. The only one which is not is the dot variant. The dot variant could receive a Dot component.

```
<Badge>
[...]

Issue #847, opened 2022-01-07
🐛  Focus contrast in darkmode could be revised.


fraincs:
## Describe the bug

When focusing on an element in dark mode the focus ring colour contrast could be enchanced.

![image](https://user-images.githubusercontent.com/361632/148576274-08c8b76c-fc5a-45ee-8be9-0a5fb971e0f3.png)

## Steps to reproduce

Focus on an input in dark mode.

## Expected results

The contrast is better.

GateNk:
![image](https://user-images.githubusercontent.com/41477329/150203184-4ac0ebb3-4379-48ff-af9d-162fabba23e4.png)

Voici une nouvelle couleur pour le dark mode qui a un contrast ratio similaire pour le dark mode:
#3590DD  35%

fraincs:
> 

magnifico

PR #886, opened 2022-01-24
Feature 847 862 868

fraincs:
Issue: 

## Summary

#847 - Focus contrast in darkmode could be revised.
#862 - Badge should allow different background colors
#868 - Alignment of Popover X button

## What I did

- Created two now colour aliases to swap the focus ring color by colorScheme.
- Changed badge colour to alias-accent, according to designs.
- Updated position of close icon/


packages/components/src/image/src/Image.css
```diff
@@ -9,6 +9,6 @@
 }
 
 /* DARK MODE */
-.o-ui-dark-color-scheme .o-ui-image {
+.o-ui-shargate-dark .o-ui-image {
```
patricklafrance, line 12: 2 issues here:

1- you misspelled "sharegate"
2- I don't like the fact that this selector is coupled to a specific theme. I would prefer a generic class which is not dependent of the theme name

packages/components/src/styling/src/theming/orbitTheme.ts
```diff
@@ -162,6 +162,8 @@ export interface ColorAliases {
     "text-alias-success": AliasValue;
     "text-alias-static-white": AliasValue;
     "text-alias-input-placeholder": AliasValue;
+    // Focus
+    "focus-ring": AliasValue;
```
patricklafrance, line 166: should the name contains "alias" to match the naming convention of the others?

patricklafrance (changes requested):

patricklafrance:
Dot badges are broken

image

fraincs:
> I don't understand why Chromatic didn't flag the change?

it did I somehow read square in the test name and didn't think much of it 😰  fixed

packages/components/src/styling/src/theming/orbitTheme.ts
fraincs: in that case it should be named something like `focus-ring-color-alias-default`

packages/components/src/image/src/Image.css
fraincs: You would add a class to the image based on the colorscheme?

alexasselin008 (approved):

packages/components/src/image/src/Image.css
patricklafrance: It seems to be your intent.

Why did you made this change?

fraincs: `dark-color-scheme` was not applied due to an old renaming of things, my intent was to fix this regression, I will go the token route, as it's a more robust solution.

patricklafrance: This is one way to do it. you could also update the `getColorSchemeClassName` function used by the ThemeProvider to render an additional `o-ui-[light|dark]` class.

Since this class wouldn't be specific to a theme, it's fine to use it anywhere.

PR #900, opened 2022-01-31
Added an explanation about type scale in the text component docs

patricklafrance:
Issue: 

## Summary

Added an explanation about type scale in the text component docs


fraincs (approved):
GG

Issue #891, opened 2022-01-26
🐛 Horizontal Illustrations should provide better positionning for heading and content


alexasselin008:
## Describe the bug

![image](https://user-images.githubusercontent.com/38871812/151234377-bebc80ff-e8e0-43bb-a312-8f0790b1be0f.png)

The more i think about it, the more i think we should maybe split this component in 2. Illustration and illustratedImage.
because the position props makes it really weird. 

## Steps to reproduce
https://orbit.sharegate.design/?path=/docs/illustration--horizontal
```
<Illustration orientation="vertical">
    <SvgImage src={NoResults} stroke="alias-primary" />
    <Heading>Can't find "Saturn"</Heading>
    <Content>Please try another search</Content>
</Illustration>
```

## Expected results

In this case, the expected result would be something like this: 
![image](https://user-images.githubusercontent.com/38871812/151234579-9f4dd0e1-8daf-43d0-9769-5fdb95f9a764.png)

GateNk:
Si ça peut aider, voici les règles qu'AlexV m'a partagé concernant les tailles d'images et le positionnement des textes:

> - Pour des images verticales, limitation à 320px de hauteur
> - Pour des images horizontales, limitation à 400px
> - Pour les images carrés ou presque, limitation à 320px
> - Espacement entre l’illustration et son bloc de texte à 40px.

Je n'étais pas au courant qu'on avait une composante ''Illustration'' qui gérait à la fois l'illu et le texte! :O 

On a des exemples ici de ce que ça donnerait pour nos empty states dans Teams management: https://www.figma.com/file/vcOOa3wctqiv1CJ57Akt5T/Empty-states?node-id=25%3A22905

fraincs:
> i think we should maybe split this component in 2. Illustration and illustratedImage.

Totally, an Illustration should not deal with text. An IllustratedMessage using Illustration would make sense.

alexasselin008:
It would fix the issue where a "vertical" illustration has the image and the text on an "horizontal" line, wich is really confusing

patricklafrance:
We could seggregate the 2 components but I don't think we really have problem here. It's only a CSS bug. The text should spaced from the image.

fraincs:
> The second does make sense and it could easily be fix.

The problem I am seeing is that when horizontal an illustration needs a dimension. 
[...]

PR #904, opened 2022-02-07
Feature/891 illustrated message

LaurentDario:
Introducing a new IllustratedMessage component. Also strip out the Heading and Content slots from Illustration to leave it as simply a wrapper for images.

Issue: #891 

## Summary

New component allowing for a more complex handling of the text next to an illustration.
image


fraincs:
I would add an alt to Horizontal Illustration sample as it's flagged as an accessibility problem.

fraincs:
image

I would modify these colours to be semantic colours in order to react to a dark theme.

fraincs:
image

I would remove the padding to the content as if the heading is already pushing it and without a heading we don't want the content pushed lower.

patricklafrance:
> image
> 
> I would remove the padding to the content as if the heading is already pushing it and without a heading we don't want the content pushed lower.

Why should an IlustratedMessage have borders and a background in the first place?

These things should only be specific to an Illustration component.

patricklafrance:
I really don't get the new components.

I though an Illustration would be:

image

and

image

Basically, an image centered in a container with optional rounded borders. The use case is to had them as some kind of banner in a Modal.

I though IllustratedMessage would be by default:

image

and if an horizontal orientation is specified:

image

I don't understand why it should be more. The idea was inspired by https://react-spectrum.adobe.com/react-spectrum/IllustratedMessage.html.

The use case for an IllustrationMessage is to use it to display states like an empty list state or a 404 not found.

I can't find any reason to have some kind of hybrid UI between both components like this:

image

alexasselin008:
I think the component is great, but the doc could be simplified a bit. 

Horizontal
An illustrated message can render horizontally.

> I would change the image to a planet, the image is too large for a horizontal alignement and it makes it look weird

```
  <IllustratedMessage>
      <SvgImage src={Planet} aria-label="Galaxy" stroke="alias-primary" />
      <Heading>Can't find "Saturn"</Heading>
      <Content>Please try another search</Content>
  </IllustratedMessage>
```

Vertical
An illustrated-message can render vertically.

> I would remove all the unnecessary props on the illustrated. The message component isnt made to be in a container with a backgroun image with padding

> Change the Image to the NoResults Image, it makes more sense here

```
<IllustratedMessage  orientation="vertical">
    <Image src={Planet} alt="Planet over another planet" />
    <Heading>Can't find "Saturn"</Heading>
    <Content>Please try another search</Content>
</IllustratedMessage>
```


Shape
An illustrated-message can have a rounded shape.

> Should be removed entirelly. This property doesnt make sense on this component. If we want a background behind, we will put this component inside another.

packages/components/src/illustrated-message/src/IllustratedMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+export interface InnerIllustratedMessageProps extends SlotProps, InternalProps, StyledComponentProps<typeof DefaultElement> {
+    /**
+     * React children.
+     */
+    children: ReactNode;
+    /**
+     * The orientation of the illustrated message.
+     */
+    orientation?: ResponsiveProp<"horizontal" | "vertical">;
+    /**
```
alexasselin008, line 19: as mentionned in my other comment, remove this

packages/components/src/illustrated-message/docs/IllustratedMessage.stories.mdx
```diff
@@ -0,0 +1,79 @@
+import { ArgsTable, Meta, Story } from "@storybook/addon-docs";
+import { ComponentInfo, Preview, Tagline } from "@stories/components";
+import { Content } from "@components/placeholders";
+import { Heading } from "@components/typography";
+import { IllustratedMessage, InnerIllustratedMessage } from "@components/illustrated-message";
+import { Image, SvgImage } from "@components/image";
+import { ReactComponent as NoResults } from "./assets/no-results.svg";
```
patricklafrance, line 7: Should be imported directly from "./assets"

packages/components/src/illustrated-message/docs/IllustratedMessage.stories.mdx
```diff
@@ -0,0 +1,79 @@
+import { ArgsTable, Meta, Story } from "@storybook/addon-docs";
+import { ComponentInfo, Preview, Tagline } from "@stories/components";
+import { Content } from "@components/placeholders";
+import { Heading } from "@components/typography";
+import { IllustratedMessage, InnerIllustratedMessage } from "@components/illustrated-message";
+import { Image, SvgImage } from "@components/image";
```
patricklafrance, line 6: Should be imported directly from "./assets"

packages/components/src/illustrated-message/docs/IllustratedMessage.stories.mdx
patricklafrance, line 7: Should be imported directly from "./assets" instead of "./assets/no-results.svg"

packages/components/src/illustrated-message/src/IllustratedMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+     */
+    shape?: "straight" | "rounded";
+}
+
+export function InnerIllustratedMessage({
+    as = DefaultElement,
+    children,
+    forwardedRef,
+    orientation = "horizontal",
+    shape = "straight",
```
patricklafrance, line 30: Shouldn't have a shape.

packages/components/src/illustrated-message/src/IllustratedMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+     * The illustrated message shape.
+     */
+    shape?: "straight" | "rounded";
+}
+
+export function InnerIllustratedMessage({
+    as = DefaultElement,
+    children,
+    forwardedRef,
+    orientation = "horizontal",
```
patricklafrance, line 29: I think the orientation should be "vertical" by default

patricklafrance (changes requested):

packages/components/src/illustrated-message/src/IllustratedMessage.tsx
alexasselin008: shape

packages/components/src/illustrated-message/src/IllustratedMessage.css
```diff
@@ -0,0 +1,44 @@
...
+    gap: var(--o-ui-sp-8);
+}
+
+.o-ui-illustrated-message-vertical {
+    flex-direction: column;
+    text-align: center;
+    gap: var(--o-ui-sp-4);
+}
+
+/* SHAPES */
```
patricklafrance, line 31: Shapes CSS is not used anymore

packages/components/src/illustrated-message/src/IllustratedMessage.tsx
```diff
@@ -0,0 +1,77 @@
...
+    children,
+    forwardedRef,
+    orientation = "vertical",
+    ...rest
+}: InnerIllustratedMessageProps) {
+    const orientationValue = useResponsiveValue(orientation);
+
+    const { content, heading, image } = useSlots(children, useMemo(() => ({
+        _: {
+            required: ["image"]
```
patricklafrance, line 32: It doesn't really make sense anymore to use an IllustratedMessage component without an header and a content.

Therefore, all slots should be required, meaning: `required: ["image", "content", "heading"]`

packages/components/src/illustrated-message/tests/chromatic/IllustratedMessage.chroma.jsx
```diff
@@ -0,0 +1,128 @@
...
+        </IllustratedMessage>
+    )
+    .add("very long content", () =>
+        <IllustratedMessage width="700px" height="200px" orientation="horizontal">
+            <Image src={Nasa} alt="Nasa" />
+            <Heading>Can't find "Saturn"</Heading>
+            <Content>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nunc suscipit metus neque, non pharetra enim tincidunt dictum. Fusce in ultricies turpis, vitae finibus nunc. Quisque laoreet sit amet eros eget volutpat. Pellentesque non nulla dui. Sed nec felis quam. Vestibulum velit magna, fringilla ut neque cursus, porta rhoncus nulla. Suspendisse auctor sollicitudin tortor, quis viverra tellus egestas sed. Pellentesque ut dignissim nisi. Duis sit amet ex bibendum, pharetra purus eget, varius massa. In pulvinar dui quis dignissim commodo. Nulla facilisi..</Content>
+        </IllustratedMessage>
+    )
+    .add("no title", () =>
```
patricklafrance, line 71: I don't think we should allow this use case.

packages/components/src/illustration/src/Illustration.tsx
```diff
@@ -31,19 +30,10 @@ export function InnerIllustration({
 }: InnerIllustrationProps) {
     const orientationValue = useResponsiveValue(orientation);
 
-    const { content, heading, image } = useSlots(children, useMemo(() => ({
+    const { image } = useSlots(children, useMemo(() => ({
```
patricklafrance, line 33: An Illustration component shouldn't use "slots" anymore since it now only accept a single children.

packages/components/src/illustration/src/Illustration.tsx
```diff
@@ -65,8 +55,6 @@ export function InnerIllustration({
             )}
         >
             {image}
```
patricklafrance, line 57: Then it would be {children} here.

packages/components/src/illustration/src/Illustration.tsx
```diff
@@ -31,19 +30,10 @@ export function InnerIllustration({
...
-            as: Text,
-            className: "o-ui-illustration-content"
-        },
-        heading: {
-            as: "h3",
-            className: "o-ui-illustration-heading",
-            size: "sm"
-        },
         image: {
             className: "o-ui-illustration-image"
```
patricklafrance, line 38: To add the class "o-ui-illustration-image" to the image, you can use the `augmentElement` function

patricklafrance (changes requested):

packages/components/src/illustrated-message/tests/chromatic/IllustratedMessage.chroma.jsx
fraincs: The components allows it, I can see some uses cases where a title would be too much and some italic explanation would be enough.

patricklafrance: I would validate with the designers. I personally would see those cases as out of brand.

alexasselin008:
Probably not the right behavior : 
image

alexasselin008:
image
3 way of writing the component name in the documentation

fraincs:
> 3 way of writing the component name in the documentation

illustrated message is the way to go if we want to follow the other camelCased components.

alexasselin008 (approved):

Issue #859, opened 2022-01-11
💡 Tag should have rounded corner (like a button)


patricklafrance:
## Problem to solve

In term or border-radius, tags are currently the exception in Orbit. Their border-radius is more accentuated than the other components.

Because of this, they don't fit well when they are inlined with buttons:

![image](https://user-images.githubusercontent.com/794579/148886858-c4c11f39-6b53-427c-b032-83661cd0eddd.png)

And their close button looks quite compact and it's border border-radius doesn't fit with the tag:

![image](https://user-images.githubusercontent.com/794579/148887013-fec30ed0-9c69-4faa-84fe-73065f5e77e9.png)

## Idea

Best solution IMO: Update the tag to use the same border-radius has a button.

Consolution: Introduce a variant of the close button with the same border-radius as the current tag.

PR #906, opened 2022-02-11
Feature/859

fraincs:
Issue: 

#859 - Tag should have rounded corner (like a button)

## What I did

- Tags now have rounded corners.
- Tags have a revamped padding  when a dot / icon is present.
- Buttons now follow the same rules as Tags padding.
- Fixed an issue where the documentation Planet image could trick you into thinking that a component had too much padding.
- Fixed an issue where in certains situations a Switch control could go further than it's parent.


patricklafrance:
Yyyyeeeesssss!!!

packages/components/src/button/src/Button.tsx
```diff
@@ -144,7 +144,7 @@ export function InnerButton(props: InnerButtonProps) {
         text: {
             "aria-hidden": loading,
             className: "o-ui-button-text",
-            size: condensed ? condensedTextSize(sizeValue) : sizeValue
+            size: condensed ? condensedTextSize(sizeValue) : "md"
```
patricklafrance, line 147: So you don't want the button text to accept a responsize size when it's not condensed?

packages/components/src/tag/src/Tag.tsx
```diff
@@ -81,7 +81,7 @@ export function InnerTag({
         text: {
             className: "o-ui-tag-text",
             color: "inherit",
-            size: sizeValue
+            size: "md"
```
patricklafrance, line 84: So you don't want a tag text to accept a responsize size value?

patricklafrance:
309 chromatic changes 😱 I hope you won’t miss anything!

Would be nice to find a way to avoid combining changes which results in to much changes in chromatic.

packages/components/src/button/src/Button.tsx
fraincs: It has been decided that the current font size on small buttons is too small. md is the minimum a button text should be only the height changes between a small and medium button now.

packages/components/src/tag/src/Tag.tsx
fraincs: It has been decided that the current font size on small tags is too small. md is the minimum a tag text should be only the height changes between a small and medium tag now.

fraincs:
> 309 chromatic changes 😱 I hope you won’t miss anything!

This is what happens when you change one thing on a button. Buttons are everywhere.

packages/components/src/button/src/Button.tsx
patricklafrance: Got it but it should still allow a responsive syntax.

fraincs: enlighten me on the how to do it :)

Chicoalvi:
> Should a Button and a Tag have the same border color?
> 
> image

It's just my POV.

The whole area of a button is cleavable ≠ The clickable area (for closing) of the tag is just the area of the cross button
>so the graphic rendering of hovering is different

As user, i easy identify the differents components side by side. I know that are different and that they haven’t the same function

packages/components/src/button/src/Button.tsx
patricklafrance: > > Should a Button and a Tag have the same border color?
> > image
> 
> It's just my POV.
> 
> The whole area of a button is cleavable ≠ The clickable area (for closing) of the tag is just the area of the cross button
> 
> > so the graphic rendering of hovering is different
> 
> As user, i easy identify the differents components side by side. I know that are different and that they haven’t the same function

@Chicoalvi make sense, thanks!

patricklafrance: > enlighten me on the how to do it :)

You can do it with `useResponsiveValue`. To enforce the the minimum size constraint, you either use either `createSizeAdapter` or `adaptSize`.

DM me if you need help.

fraincs:
> As user, i easy identify the differents components side by side. I know that are different and that they haven’t the same function

I think the difference is not subtle enough. Thoughts?

fraincs:
It has been discussed that a TagList should be solid by default.

image

patricklafrance (approved):

Issue #897, opened 2022-01-29
🐛 A tile with row size is not responsive


patricklafrance:
## Describe the bug

The tiles should try to take the available space.

(https://orbit.sharegate.design/?path=/docs/tile--tile#row-size)

![Tile row size](https://user-images.githubusercontent.com/794579/151671190-ced36b8c-905f-4e07-afe2-a3a0e511e80a.png)

fraincs:
Layout drives responsiveness the sample has no layout, all those responsive issues should be in an umbrella issue, components are responsive per se but a responsive layout is needed, to tell them how to behave.

patricklafrance:
I don't agree.

A tile group have a layout, even more with a row-size option.

If the consensus is that a tile group doesn't have a layout then I think the TileGroup component have to be removed and maybe a hook should be introduced for the group selection behavior.

patricklafrance:
You can create an umbrella issue if you feel it will help you :)

Issue #896, opened 2022-01-29
🐛 A modal with choices is not responsive


patricklafrance:
## Describe the bug

(https://orbit.sharegate.design/?path=/docs/modal--default-story#choice)

For small viewports, the choices, should switch from a column layout to a row layout.

In the second choice, there is a huge padding before the title.

![Modal choices](https://user-images.githubusercontent.com/794579/151671023-87c53242-7e12-4bc6-97d4-58e21ff2dc74.jpg)

Issue #894, opened 2022-01-29
🐛 Form fieldset is not responsive


patricklafrance:
## Describe the bug

Fields should collapse on multiple line.

https://orbit.sharegate.design/?path=/docs/form--vertical-layout#fieldset

![Fieldset](https://user-images.githubusercontent.com/794579/151670737-d14fdd41-aeb9-4577-a4df-feb66a0f14bc.png)

fraincs:
Ended up refactoring a little bit of the Form css in order to better react to mobile layouts.

Issue #893, opened 2022-01-29
🐛 Form vertical layout is not responsive


patricklafrance:
## Describe the bug

https://orbit.sharegate.design/?path=/docs/form--vertical-layout#vertical-layout

![Vertical form](https://user-images.githubusercontent.com/794579/151670694-dcdab12f-9160-42e7-b070-422ca109432f.png)

PR #909, opened 2022-02-18
Mobile doc fix - 

fraincs:
Issue: 

## Summary

Many doc samples were not mobile friendly - this PR aims at fixing most of them. This PR also introduces a bugfix where small Tag and Buttons were using icons that were too small compared to the font size.

## What I did

#859 - 💡 Tag should have rounded corner (like a button)
#897  - 🐛 A tile with row size is not responsive
#896 - 🐛 A modal with choices is not responsive
#894 - 🐛 Form fieldset is not responsive
 #893 - 🐛 Form vertical layout is not responsive

### DIFF
--- a/packages/components/src/form/docs/Form.stories.mdx
+++ b/packages/components/src/form/docs/Form.stories.mdx
@@ -4,7 +4,7 @@ import { Checkbox } from "@components/checkbox";
 import { ComponentInfo, Preview, Tagline } from "@stories/components";
 import { Field, HelpMessage, Label } from "@components/field";
 import { Fieldset, Form, InnerFieldset, InnerForm } from "@components/form";
-import { Grid, Inline, repeat } from "@components/layout";
+import { Flex, Grid, Inline, repeat } from "@components/layout";
 import { PasswordInput, TextInput } from "@components/text-input";
 
 <Meta
@@ -147,10 +147,10 @@ Form fields can be grouped together within fieldsets.
     <Story name="fieldset">
         <Form>
             <Fieldset label="Shipping Address">
-                <Grid
-                    templateColumns={{
-                        base: "1fr",
-                        sm: repeat(2, "1fr")
+                <Flex
+                    flexDirection={{
+                        base: "column",
+                        sm: "row"
                     }}
                 >
                     <Field required>
@@ -161,11 +161,11 @@ Form fields can be grouped together within fieldsets.
                         <Label>Last name</Label>
                         <TextInput />
                     </Field>
-                </Grid>
-                <Grid
-                    templateColumns={{
-                        base: "1fr",
-                        sm: repeat(2, "1fr")
+                </Flex>
+                <Flex
+                    flexDirection={{
+                        base: "column",
+                        sm: "row"
                     }}
                 >
                     <Field required>
@@ -176,17 +176,17 @@ Form fields can be grouped together within fieldsets.
                         <Label>Appartment</Label>
                         <TextInput />
                     </Field>
-                </Grid>
+                </Flex>
                 <Field required>
                     <Label>Postal code</Label>
                     <TextInput />
                 </Field>
             </Fieldset>
             <Fieldset label="Billing Address">
-                <Grid
-                    templateColumns={{
-                        base: "1fr",
-                        sm: repeat(2, "1fr")
+                <Flex
+                    flexDirection={{
+                        base: "column",
+                        sm: "row"
                     }}
                 >
                     <Field required>
@@ -197,11 +197,11 @@ Form fields can be grouped together within fieldsets.
                         <Label>Last name</Label>
                         <TextInput />
                     </Field>
-                </Grid>
-                <Grid
-                    templateColumns={{
-                        base: "1fr",
-                        sm: repeat(2, "1fr")
+                </Flex>
+                <Flex
+                    flexDirection={{
+                        base: "column",
+                        sm: "row"
                     }}
                 >
                     <Field required>
@@ -212,7 +212,7 @@ Form fields can be grouped together within fieldsets.
                         <Label>Appartment</Label>
                         <TextInput />
                     </Field>
-                </Grid>
+                </Flex>
                 <Field required>
                     <Label>Postal code</Label>
                     <TextInput />


alexasselin008 (approved):

packages/components/src/button/src/Button.tsx
```diff
@@ -142,11 +148,11 @@ export function InnerButton(props: InnerButtonProps) {
         },
         "end-icon": {
             className: "o-ui-button-end-icon",
-            size: condensed ? sizeValue : embeddedIconSize(sizeValue)
+            size: condensed ? sizeValue : iconSize(sizeValue)
```
patricklafrance, line 151: This changes is a smell to me. The embeddedIconSize utility is to ensure all our components use the same sizes for embedded icons.

Does the button component is not standard anymore? Does the embeddedIconSize needs an update becomes it doesn't reflect our DA anymore?

Those questions should be answered before making the change.

packages/components/src/tag/src/Tag.tsx
```diff
@@ -80,11 +85,11 @@ export function InnerTag({
         },
         "end-icon": {
             className: "o-ui-tag-end-icon",
-            size: embeddedIconSize(sizeValue)
+            size: iconSize(sizeValue)
```
patricklafrance, line 88: Same concerns has I mentioned for the button component

packages/components/src/form/docs/Form.stories.mdx
```diff
@@ -37,7 +37,13 @@ A form can stack fields vertically.
 
 <Preview>
     <Story name="vertical layout">
-        <Form fluid width="25%">
+        <Form fluid
```
patricklafrance, line 40: If you set this one as fluid, then should we set all the problematic examples also as fluid? Feels weird though to have all the samples as fluid.

Maybe we could add some hacky CSS only for the docs to update our input min width CSS variables for our small breakpoints. This hacky CSS would be specific to the component and won't alter the components in apps.

Same thing could be done for tiles and other components with a problematic min width.

packages/components/src/form/src/Form.css
```diff
@@ -33,6 +25,8 @@
     margin-bottom: var(--o-ui-sp-3);
 }
 
-.o-ui-fieldset .o-ui-form-field:last-child {
-    margin-bottom: 0;
+.o-ui-fieldset-fields {
+    display: grid;
+    width: 100%;
+    gap: var(--o-ui-sp-5);
```
patricklafrance, line 31: I don't think the `width: 100%` is necessary and I am not sure it should be specified by default since Orbit usually a component to explicit receive a `fluid` property to display something as fluid.

packages/components/src/form/src/Form.css
```diff
@@ -33,6 +25,8 @@
     margin-bottom: var(--o-ui-sp-3);
 }
 
-.o-ui-fieldset .o-ui-form-field:last-child {
-    margin-bottom: 0;
+.o-ui-fieldset-fields {
+    display: grid;
```
patricklafrance, line 29: A grid display without any specific templating doesn't seem right.

patricklafrance:
Horizontal tile is still broken:

image

packages/components/src/form/tests/chromatic/Form.chroma.jsx
```diff
@@ -88,7 +88,7 @@ stories()
                 </Field>
                 <Button variant="secondary">Submit</Button>
             </Form>
-            <Form disabled>
+            <Form disabled alignSelf="end">
```
patricklafrance, line 91: The align-self actually seems to move the form downward on desktop, what's the intention here?

fraincs:
> Horizontal tile is still broken:

swapping to vertical would not show the horizontal btile

packages/components/src/form/tests/chromatic/Form.chroma.jsx
fraincs: Without an align child of flex items stretch by default, making the smaller form stretch, start could be used too, iirc many chromatic tests behaves this way (aligning items at the end).

packages/components/src/form/src/Form.css
fraincs: This makes gap works ;) Flex could be used here too, Grid has the advantage of not messing with the layout direction.

packages/components/src/form/src/Form.css
fraincs: I'll dig down the reason it's been added and follow through

packages/components/src/button/src/Button.tsx
fraincs: > Horizontal tile is still broken:

as for the image height some discussions have started here: https://github.com/gsoft-inc/sg-orbit/issues/898#issuecomment-1042368002

I was planning to at least fix the overflow then tackle the problem of the image not stretching as it's a different issue to me.

A new image will need to be produced too in order to be able to cover the whole image area without stretching the image, I don't like it anyways :D

packages/components/src/form/tests/chromatic/Form.chroma.jsx
patricklafrance: Yeah but the rendering have been altered for a lg breakpoint, not sure this expected

packages/components/src/form/src/Form.css
patricklafrance: Yeah I know, I guess I would have choose Flex since there isn’t any expected templating here.

That being said, should a fieldset really have any layout? I think it’s the content of the fieldset which should be responsable for the layout.

packages/components/src/button/src/Button.tsx
patricklafrance: Got it

fraincs: > embeddedIconSize utility is to ensure all our components use the same sizes for embedded icons.

in that case a review of all the components shall be done, the main goal here is to make sure we don't go lower than 14px in font-size as 12px is hard to read. By doing this it was logical to bump the icon sizes in small components. This might be an error but I think we should then bump all icons sizes to avoid icons that are hard to decipher.

packages/components/src/form/tests/chromatic/Form.chroma.jsx
fraincs: The goal is to make sure the forms don't stretch and keep their initial gap. Since the parent is a flex item it's needed I am not sure what you mean by rendering have been altered for a lg breakpoint compared to the old test?

packages/components/src/form/src/Form.css
patricklafrance: It should be a Flex or Grid layout inside the Fieldset which provide the spacing between the fields.

Fieldsets is only there to provide semantic meaning not layout.

packages/components/src/form/docs/Form.stories.mdx
patricklafrance: Well the core issue doesn't happens in our apps, only in SB because of the code block style.

packages/components/src/form/tests/chromatic/Form.chroma.jsx
patricklafrance: They are not vertically aligned because of the "alignSelf=end".

image

I don't really mind, I didn't though it would escalade, I though it would be a "ah ah oopsy moment" and it would be over quickly. You can keep it this way if you want.

packages/components/src/button/src/Button.tsx
fraincs: Indeed added this to the Sync discussion as it can be not the path we want to take, take Lozenge as an example.

packages/components/src/form/src/Form.css
fraincs: This means the end user will need to add a wrapper, setting the field set as a flex layout with a gap takes this off the consumer hands. Free responsive layout 
image

packages/components/src/form/tests/chromatic/Form.chroma.jsx
fraincs: Remove the align in the dev tools and you'll see what happens.

packages/components/src/form/src/Form.css
patricklafrance: I am not sure I understand what you mean but the code sample for Fieldset already includes a Grid so what else does the user will need to add?

image

packages/components/src/form/tests/chromatic/Form.chroma.jsx
patricklafrance: image

packages/components/src/form/src/Form.css
fraincs: The right amount of gap. Or fieldset could inherit gap?

patricklafrance: What do you mean by the right amound of gap?

My current guess is that you are trying a problem I am not aware exist.

patricklafrance:
@fraincs let me know when alI can review this again after the changes we talked about on our call yesterday are implemented.

Slack is down :(

fraincs:
> Slack is down :(

I was down too :) you can have a look now.

fraincs:
> > Slack is down :(
> 
> I was down too :) you can have a look now.

 @patricklafrance

packages/components/src/form/docs/Form.stories.mdx
```diff
@@ -147,7 +147,12 @@ Form fields can be grouped together within fieldsets.
     <Story name="fieldset">
         <Form>
             <Fieldset label="Shipping Address">
-                <Inline>
+                <Grid
+                    templateColumns={{
+                        base: "1fr",
+                        sm: repeat(2, "1fr")
+                    }}
```
patricklafrance, line 154: I still don't get why someone would use a Grid instead of Flex to horizontally align 2 fields :)

fraincs: Either / or. Both achieve the same results, I prefer grid since the repeat and 1fr are explicit and so much easier to use than flex. I could use flex and swap direction too.

Issue #917, opened 2022-03-07
🚀  Remove Fullscreen Modal option


fraincs:
## Description

It has been discussed that a FullSrceen is a bad idea and should not be used. Therefore removing it would make sure it's never used.

Issue #911, opened 2022-02-24
🐛A fullscreen modal is not wide enough on mobile


patricklafrance:
## Describe the bug

On mobile a fullscreen modal take less horizontal space than a regular modal:

Fullscreen modal:

![image](https://user-images.githubusercontent.com/794579/155442929-194c8806-4808-46bd-9c5d-e66fa919e27a.png)

Regular modal:

![image](https://user-images.githubusercontent.com/794579/155442976-a450cb70-0b0a-49f6-9a31-362effd647ae.png)

On mobile, I would expect a full screen modal to really take the whole space, something similar to this:

![image](https://user-images.githubusercontent.com/794579/155443826-63b2973f-180d-4b6a-8851-9c168bcef9f9.png)

fraincs:
In that case I would change the full screen description too

> A modal can take almost all the available space.

as it's not always true to : 

> A modal can take as much space as possible.

patricklafrance:
It depends what we choose to do on desktop.

fraincs:
> It depends what we choose to do on desktop.

True, I guess it won't be an easy fix after all.

fraincs:
https://github.com/gsoft-inc/sg-orbit/issues/913 has been opened to tackle bigger issues, I will fix the padding of fullscreen modal in mobile per the issue and discussion regarding whether or not a fullscreen modal should take the whole viewport can take place in the issue above.

fraincs:
Since Fullscreen Modals are no more, I am closing this issue.

patricklafrance:
@fraincs it might not be called "fullscreen" anymore but I think the behavior of taking all the available space when on mobile should still be considered.

What do you think?

Issue #910, opened 2022-02-19
🐛 A bordered accordion button have an undesirable gap under the gray zone when touched


patricklafrance:
## Describe the bug

A bordered accordion button have an undesirable gap under the gray zone when touched.

![273793294_937263063651742_5297677744442439909_n](https://user-images.githubusercontent.com/794579/154812283-37ce8ec3-88c8-478e-a3ed-222bde389f73.png)

fraincs:
I guess this is on iOS? Chrome?

patricklafrance:
IOS

fraincs:
Upon inspection this is more a bug with Safari rendering engine 

![image](https://user-images.githubusercontent.com/361632/155763673-b6215d73-9b1e-4744-a3aa-7f2be5709312.png)

Closing in favour of #912

Issue #812, opened 2021-12-06
🐛  Password Icon and Lastpass autofill are competing for attention


fraincs:
## Describe the bug

When a user uses LastPass, LastPass icon is shown above Orbit's Eye icon that allows to show what you've typed.

## Steps to reproduce

Use a password field and LastPass Chrome extension.

## Expected results

The two icons are not on top of each other.

fraincs:
With the way we display our password icon, Password managers that adds their own icons over inputs gets in the way of the reveal password icon. See below

![image](https://user-images.githubusercontent.com/361632/147095212-a90c794e-7125-4027-bb44-51ecc888b99e.png)

While detecting the most common password managers this is not guaranteed to work when a third party updates it's code, nor works for obscure password managers.

This issues aims at finding an elegant solution to make sure these password managers icons never gets in the way.

Chicoalvi:
Is it possible to detect if a user uses a password management software and this software displays its icon in the password field ?

If it is possible, we could adjust the positioning of our eye icon according to the presence or not.

Chicoalvi:
or another solution more simple ![image](https://user-images.githubusercontent.com/16927274/147970474-04b8b3cb-8e58-4e35-afde-aa476e09b2b1.png)

patricklafrance:
If we want to keep an inline "show" icon the common solution is to move all the input style on the wrapper.

For example:

```
<div class="wrapper">
     <input type="password" />
     <button type="button>View</button>
<div>
```

Move the input style (background, border, ...) on the wrapper element. Therefore, from a user standpoint, the button seem to be in the input. For a password manager, it can safely add it's CTA at the end of the input without overlapping our "view" button.

Ex here: https://vaadin.com/docs/latest/example?path=component/passwordfield/password-field-basic.ts

fraincs:
Late answer but I think this is the sensitive solution, moving the input style to the wrapper element.

PR #919, opened 2022-03-09
Issues : 917 - 911 - 910 - 812

fraincs:
Issue: 

#917 / #911 / #910 / #812

## What I did

- Fixed an issue that overlapped password managers icons on reveal password icon.
- Fixed an issue where Modals were not responsive.
- Fixed an issue where Accordions in Safari had a bad hover state position.
- Removed fullscreen Modal as they are of no use.

### DIFF
--- a/packages/components/src/dialog/src/Dialog.css
+++ b/packages/components/src/dialog/src/Dialog.css
@@ -17,10 +17,6 @@
     align-items: flex-start;
 }
 
-.o-ui-dialog-wrapper-fullscreen {
-    padding: 0;
-}
-
 /* DIALOG */
 .o-ui-dialog {
     position: relative;
@@ -62,12 +58,6 @@
     width: 960px;
 }
 
-.o-ui-dialog-fullscreen {
-    width: 100%;
-    min-height: 100%;
-    border-radius: 0;
-}
-
 /* DIALOG | IMAGE & ILLUSTRATION */
 .o-ui-dialog-image,
 .o-ui-dialog-illustration {
@@ -93,6 +83,7 @@
     max-height: 100%;
     overflow: hidden;
     object-fit: cover;
+    object-position: top center;
 }
 
 .o-ui-dialog-illustration {
--- a/packages/components/src/dialog/src/Dialog.tsx
+++ b/packages/components/src/dialog/src/Dialog.tsx
@@ -65,7 +65,7 @@ export interface InnerDialogProps extends AbstractDialogProps<typeof DefaultElem
     /**
      * A dialog can vary in size.
      */
-    size?: ResponsiveProp<"sm" | "md" | "lg" | "fullscreen">;
+    size?: ResponsiveProp<"sm" | "md" | "lg">;
 }
 
 function useHideBodyScrollbar() {
@@ -302,8 +302,7 @@ export function InnerDialog({
                     {
                         className: cssModule(
                             "o-ui-dialog-wrapper",
-                            wrapperHasVerticalScrollbar && "scrolling",
-                            size === "fullscreen" && "fullscreen"
+                            wrapperHasVerticalScrollbar && "scrolling"
                         ),
                         ref: useMergedRefs(wrapperRef, hasVerticalScrollbarRef),
                         zIndex: zIndex + 1
@@ -321,7 +320,7 @@ export function InnerDialog({
                             as,
                             className: cssModule(
                                 "o-ui-dialog",
-                                sizeValue === "fullscreen" ? sizeValue : normalizeSize(sizeValue),
+                                normalizeSize(sizeValue),
                                 focus && "focus"
                             ),
                             id: dialogId,
--- a/packages/components/src/input/src/Input.css
+++ b/packages/components/src/input/src/Input.css
@@ -14,13 +14,23 @@
     transition: var(--o-ui-focus-ring-transition);
     color: var(--o-ui-text-alias-primary);
     width: var(--o-ui-sz-16);
+    min-height: var(--o-ui-sz-6);
+    height: 100%;
 }
 
 .o-ui-input input {
     -webkit-appearance: none;
     outline: none;
-    min-height: var(--o-ui-sz-6);
     border: none;
+    width: 100%;
+    background-color: inherit;
+    color: inherit;
+    height: 100%;
+    min-height: calc(var(--o-ui-sz-6) - 2px);
+}
+
+.o-ui-input textarea {
+    width: 100%;
 }
 
 /* MS EDGE FIX */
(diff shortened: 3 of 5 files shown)


patricklafrance:
Issue here

image

patricklafrance:
Here too:

image

patricklafrance:
Border color is not the same has the input:

image

patricklafrance:
Textarea looks broken

image

patricklafrance:
How come the font-family is monospace?

image

patricklafrance:
When the focused, it's really hard to see the invalid red border of the input

image

patricklafrance:
Doesn't work for NumberInput (you can see that the dashlane icon is under the arrows)

image

fraincs:
> Issue here
> 
> image

most of these has been flagged by chromatic and are currently being fixed

patricklafrance:
It's weird that I only see the bottom of the image on mobile and the close button is not visible since it's under the image:

image

packages/components/src/dialog/src/Dialog.css
```diff
@@ -17,15 +17,27 @@
     align-items: flex-start;
 }
 
+.o-ui-dialog-wrapper-fullscreen {
+    padding: 0;
+}
```
patricklafrance, line 22: Does a dialog still support fullscreen?

packages/components/src/dialog/src/Dialog.css
```diff
@@ -17,15 +17,27 @@
...
     border-radius: var(--o-ui-br-rounded);
     display: grid;
     align-items: stretch;
-    grid-template-areas: "image aside";
-    grid-template-columns: min-content auto;
+    grid-template-areas: "image image" "aside aside";
+    grid-template-rows: min-content auto;
+}
+
+@media screen and (min-width: 768px) {
```
patricklafrance, line 35: I don't like the idea of duplicating the breakpoints values.

Maybe `createThemeVars` could render CSS vars for the breakpoint values then you could use the values here?

Those vars could also be useful for our apps.

packages/components/src/dialog/src/Dialog.css
```diff
@@ -51,26 +63,36 @@
 }
 
 .o-ui-dialog-fullscreen {
-    width: calc(100% - (2 * var(--o-ui-sz-6)));
-    min-height: calc(100% - (2 * var(--o-ui-sz-6)));
+    width: 100%;
+    min-height: 100%;
+    border-radius: 0;
```
patricklafrance, line 68: Does a dialog still support fullscreen?

packages/components/src/dialog/src/Dialog.tsx
```diff
@@ -302,7 +302,8 @@ export function InnerDialog({
                     {
                         className: cssModule(
                             "o-ui-dialog-wrapper",
-                            wrapperHasVerticalScrollbar && "scrolling"
+                            wrapperHasVerticalScrollbar && "scrolling",
+                            size === "fullscreen" && "fullscreen"
```
patricklafrance, line 306: Does a dialog still support fullscreen?

packages/components/src/input/src/Input.css
```diff
@@ -130,10 +121,15 @@
...
 
 /* ICON | MEDIUM */
-.o-ui-input-has-icon input {
+.o-ui-input-has-icon {
     padding-left: var(--o-ui-input-padding-with-icon);
 }
 
+/* PASSWORD */
+.o-ui-password-input{
+    padding-right: 2.25rem;
```
patricklafrance, line 130: I don't think it's necessary since it's override by `.o-ui-input-has-button.o-ui-input`

image

fraincs:
> Border color is not the same has the input:

never has been, it's up to the consumer to set the right border color

fraincs:
> It's weird that I only see the bottom of the image on mobile and the close button is not visible since it's under the image (well not under but it's black on black):

Totally, the specs are currently being written and for the close button it is stated that the close button should always be visible aka not use a dark image. I don't have any designer avail right now to generate a new image, I guess I could make something up.

As for the centering of the image the upcoming doc mention to only use decorative images, as they could be cropped or swap the image at smaller dimensions. No apps use the image modal and it might as well be unsupported as well, since it cause problem in mobile.

packages/components/src/dialog/src/Dialog.css
fraincs: > I don't like the idea of duplicating the breakpoints values.
> 
> Maybe `createThemeVars` could render CSS vars for the breakpoint values then you could use the values here?
> 
> Those vars could also be useful for our apps.

not a bad idea!

packages/components/src/dialog/src/Dialog.css
fraincs: > Doesn't work for NumberInput

Dashlane should not autofill a number input though :D

patricklafrance:
> > It's weird that I only see the bottom of the image on mobile and the close button is not visible since it's under the image (well not under but it's black on black):
> 
> Totally, the specs are currently being written and for the close button it is stated that the close button should always be visible aka not use a dark image. I don't have any designer avail right now to generate a new image, I guess I could make something up.
> 
> As for the centering of the image the upcoming doc mention to only use decorative images, as they could be cropped or swap the image at smaller dimensions. No apps use the image modal and it might as well be unsupported as well, since it cause problem in mobile.

Got it!

For the image, why focus on the bottom of the image instead of the top?

fraincs:
> For the image, why focus on the bottom of the image instead of the top?

This is how object fit cover work actually.

fraincs:
@patricklafrance somehow since the styles read padding are now applied to the input wrapper instead of the input per se, the autocomplete width is still using the input widths, I tried to find where the magic happens but I am not sure exactly how the width is determined here. Any ideas?

image

packages/components/src/dialog/src/Dialog.css
patricklafrance: @fraincs yes for "age" it would :)

packages/components/src/dialog/src/Dialog.css
patricklafrance: :(

patricklafrance:
> @patricklafrance somehow since the styles read padding are now applied to the input wrapper instead of the input per se, the autocomplete width is still using the input widths, I tried to find where the magic happens but I am not sure exactly how the width is determined here. Any ideas?
> 
> image

Probably the ref which is on the input instead of the input container. 

Providing the ref to TextInput using the `wrapperProps` should do the tricks...

```
<TextInput wrapperProps={{ ref: YOUR_REF }}  />
```

patricklafrance:
> > Border color is not the same has the input:
> 
> never has been, it's up to the consumer to set the right border color

Ok! that's weird?

PR #923, opened 2022-03-16
Feature/ssr

alexasselin008:
## Summary

Add what's document, window and useLayoutEffect checks so Orbit works properly in an SSR environement.

## What I did

Added checks and tests to support SSR

## How to test

Multiple Tests have been added to properly test each component in an SSR environement

No


jest/setup-jest.js
```diff
@@ -1 +1,37 @@
 global.ResizeObserver = require("resize-observer-polyfill");
+
+const ERROR_PATTERNS_WE_SHOULD_FIX_BUT_ALLOW = [];
+const WARNING_PATTERNS_WE_SHOULD_FIX_BUT_ALLOW = [];
+
```
alexasselin008, line 5: This was taken from React Spectrum. This ensure that the SSR errors that are only logged in the console "useLayoutEffect not supported on the server" are throwed and fail the tests

patricklafrance: Orbit already have https://github.com/gsoft-inc/sg-orbit/blob/master/jest/utils/errorHandling.tsx for this purpose.

packages/components/src/accordion/tests/jest/Accordion.ssr.test.tsx
```diff
@@ -0,0 +1,30 @@
+/**
+ * @jest-environment node
```
patricklafrance, line 2: What is it used for? Non SSR jest tests doesn't requires this.

packages/components/src/accordion/tests/jest/Accordion.ssr.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+ */
+import { Accordion } from "@components/accordion";
+import { Content } from "@components/placeholders";
+import { H3 } from "@components/typography";
+import { Item } from "@components/collection";
+import { renderToString } from "react-dom/server";
+
+test("Server-side rendering: renders on a server without crashing", () => {
```
patricklafrance, line 10: Please follow a naming convention similar to the other Jest tests. Should be something like "can render on the server"

packages/components/src/accordion/tests/jest/Accordion.ssr.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+                    <H3>Header</H3>
+                    <Content>Content</Content>
+                </Item>
+                <Item>
+                    <H3>Header</H3>
+                    <Content>Content</Content>
+                </Item>
+            </Accordion>
+        );
+    expect(renderOnServer).not.toThrow();
```
patricklafrance, line 28: line break before the `expect` please

patricklafrance:
Curious why you didn't add the SSR test to the existing test file? Could have been added into a `describe` section.

The advantage is that it would have been very easy to re-use the same setup has the non SSR tests.

packages/components/src/overlay/src/Overlay.tsx
```diff
@@ -76,7 +76,13 @@ export function InnerOverlay({
         </Transition>
     );
 
-    return createPortal(content, containerElement || document.body);
+    const host = containerElement ?? (isBrowser ? document.body : undefined);
+
+    if (host && isBrowser) {
```
patricklafrance, line 81: Why check `isBrowser` again?

Should use `!isNil(host)`

packages/components/src/overlay/src/Overlay.tsx
```diff
@@ -76,7 +76,13 @@ export function InnerOverlay({
         </Transition>
     );
 
-    return createPortal(content, containerElement || document.body);
+    const host = containerElement ?? (isBrowser ? document.body : undefined);
```
patricklafrance, line 79: Not sure `host` is the right name here, I suggest `_containerElement`

packages/components/src/overlay/src/useOverlayPosition.ts
```diff
@@ -45,15 +45,15 @@ export function useOverlayPosition({
             enabled: allowFlip,
             name: "flip",
             options: {
-                boundary: boundaryElement
+                boundary: boundaryElement ?? (isBrowser ? document.body : undefined)
```
patricklafrance, line 48: Should you add an utility function `getBodyElement` ? Maybe in the same file as `isBrowser`

packages/components/src/shared/src/dom.ts
```diff
@@ -0,0 +1,9 @@
+export function canUseDOM(): boolean {
```
patricklafrance, line 1: I honestyle hate this code lol. Not a big fan of `!!`, should `isNil` to be consistant with the codebase.

packages/components/src/shared/src/useSafeLayoutEffect.tsx
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * React currently throws a warning when using useLayoutEffect on the server.
+ * To get around it, we can conditionally useEffect on the server (no-op) and
+ * useLayoutEffect in the browser.
+ *
+ * @see https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85
+ */
+export const useSafeLayoutEffect = isBrowser
+    ? React.useLayoutEffect
+    : React.useEffect;
```
patricklafrance, line 16: Euh!!! Not sure about this.... useEffect is very different than useLayoutEffect. One is synchronous and one is asychronous. Event when 100% on the client they are not interchangeable.

packages/components/src/styling/src/BreakpointProvider.tsx
```diff
@@ -37,6 +36,8 @@ export function BreakpointProvider({
     children,
     unsupportedMatchMediaBreakpoint = "lg"
 }: BreakpointProvider) {
+    // Ensure that matchMedia function exists. In a jest environement or in SSR, this function is not available.
+    const supportsMatchMedia = typeof window !== "undefined" && typeof window.matchMedia === "function";
```
patricklafrance, line 40: Why not keep this keep this check is `useMediaQuery` ? Also, it should use `isNil` and `isFunction`

packages/components/src/styling/src/theming/createThemeVars.ts
```diff
@@ -103,12 +103,14 @@ function appendColorSchemes<C, L, D>(
 }
 
 function renderBucket(scope: string, bucket: VarsBucket) {
-    const element = document.createElement("style");
+    if(isBrowser) {
```
patricklafrance, line 106: missing a space between `if` and `(`

patricklafrance: Instead of scoping the whole code, you should exit at the beginning if it's on the server side.

patricklafrance (changes requested):
Blocking because of useSafeLayoutEffect

packages/components/src/styling/src/BreakpointProvider.tsx
alexasselin008: We can not use isNil, since we don't check the value of window (if we try to access the value of window, it breaks) so we check the type of it. isNil would not work

packages/components/src/accordion/tests/jest/Accordion.ssr.test.tsx
alexasselin008: This is use to simulate a "server side" environement. This prevents the use of document and window

packages/components/src/shared/src/useSafeLayoutEffect.tsx
alexasselin008: This is a commonly use pattern: 
Chakra: https://github.dev/chakra-ui/chakra-ui/blob/8e201b9f84a914e58cdc9ee90b3e03269fabf4d7/packages/hooks/src/use-safe-layout-effect.ts#L12
Spectrum has something similar: https://github.dev/adobe/react-spectrum/blob/91ca94fe52840b7a32b961ec08208f5fbdf65697/packages/%40react-aria/utils/src/useLayoutEffect.ts#L21
Dan Abramov suggests it : https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

the idea is that
"If you use server rendering, keep in mind that neither useLayoutEffect nor useEffect can run until the JavaScript is downloaded."

alexasselin008:
> Curious why you didn't add the SSR test to the existing test file? Could have been added into a `describe` section.
> 
> The advantage is that it would have been very easy to re-use the same setup has the non SSR tests.

Adding new files allow me to add the  @jest-environment node comment at the top of the file, simulating an "SSR" environement, where document and window are not available

packages/components/src/styling/src/BreakpointProvider.tsx
patricklafrance: Got it, still why not keep it in `useMediaQuery` ?

patricklafrance (approved):

Issue #921, opened 2022-03-09
💡 An Image src prop should accept responsive value


patricklafrance:
## Problem to solve

An application might want to provide a different image based on a breakpoint.

## Idea

`useResponsiveValue` can be used but for covenience the `src` prop could also accept responsive value.

PR #927, opened 2022-03-28
Feature/921

fraincs:
Issue: 

#921 An image src prop should accept a responsive value

## Summary

Image and SVGImage src were not supporting a responsive value as their src prop value.

## What I did

Added support for src responsive prop.


packages/components/src/image/src/SvgImage.tsx
```diff
@@ -11,28 +11,28 @@ export interface InnerSvgImageProps extends
...
-     * See [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill).
-     */
+         * See [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill).
+         */
     fill?: FillProp;
     /**
-    * The image height.
-    */
+        * The image height.
+        */
```
patricklafrance, line 23: Not sure why all these comments doesn't follow the same spacing as the other comments

patricklafrance (approved):

patricklafrance (approved):

PR #930, opened 2022-03-30
Generate manually the icons instead of using a babel plugin

alexasselin008:
Issue: 

## How to test


alexasselin008:
Documentation is missing

scripts/helper.js
```diff
@@ -0,0 +1,45 @@
+const fs = require("fs");
```
patricklafrance, line 1: To keep things consistant you should use `shelljs` like the other scripts

patricklafrance: I wouldn't create an "helper" scripts it's too generic. Instead I would create an invidual file per helper like the other helpers. With shelljs I don't even think you really need to define helpers for `clearDir`, `checkFolderExist` etc...

packages/icons/scripts/copy-and-optimize-to-dist.js
```diff
@@ -0,0 +1,31 @@
+const chalk = require("chalk");
```
patricklafrance, line 1: You could rename this for `optimize-svgs`. Since you optimize the SVG you have to copy the output somewhere it's a consequence of optimisation and doesn't really have to be explicitly specified

patricklafrance: To be consistant with the other scripts, this script should probably use "meow"

packages/icons/scripts/optimize.js
```diff
@@ -0,0 +1,107 @@
+const { optimize: svgoOptimize } = require("svgo");
```
patricklafrance, line 1: "optimize" is not specific, it could optimize a lot of things, maybe rename to "optimize-svg" ?

packages/icons/scripts/optimize.js
```diff
@@ -0,0 +1,107 @@
...
+        return true;
+    } else {
+        console.error(
+            `The size of ${name} is not correct. width: ${width} height: ${height}`
+        );
+        process.exit(1);
+    }
+};
+
+const optimize = async dir => {
```
patricklafrance, line 76: Rename to `optimizeSvg`

patricklafrance: Should you separate the validation from the optimization?

You could load all the files, proceed all files through validation, if they are all valid process all files though optimization. Separation of concerns.

packages/icons/scripts/copy-and-optimize-to-dist.js
```diff
@@ -0,0 +1,31 @@
...
+const { checkFolderExist } = require("../../../scripts/helper");
+const fs = require("fs");
+const { optimize } = require("./optimize");
+
+const SRC_PATH = "src";
+const DIST_PATH = "dist";
+
+const generateSvg = (svgList, dir) => {
```
patricklafrance, line 9: Are you really "generating" an svg? It seems more like create (or writing) a file.

packages/icons/scripts/copy-and-optimize-to-dist.js
```diff
@@ -0,0 +1,31 @@
...
+
+    svgList.map(async svg => {
+        const { data, name } = svg;
+
+        fs.writeFile(`${dir}/${name}`, data, err => {
+            if (err) {throw err;}
+        });
+    });
+
+    console.log("✨ The icons has been saved!");
```
patricklafrance, line 21: I like the icons but the other scripts doesn't have any :(

packages/icons/scripts/copy-and-optimize-to-dist.js
```diff
@@ -0,0 +1,31 @@
...
+
+const generateSvg = (svgList, dir) => {
+    checkFolderExist(dir);
+    console.log("\n⚙️  Svg generation...");
+
+    svgList.map(async svg => {
+        const { data, name } = svg;
+
+        fs.writeFile(`${dir}/${name}`, data, err => {
+            if (err) {throw err;}
```
patricklafrance, line 17: Braces should not be inlined according to our guidelines

packages/icons/scripts/copy-and-optimize-to-dist.js
patricklafrance: "Icons optimization completed!... " with a chalk.green

packages/components/scripts/generate-icons.js
```diff
@@ -0,0 +1,20 @@
+const chalk = require("chalk");
```
patricklafrance, line 1: Rename to "generate-icon-components" ?

packages/components/scripts/generate-react-components.js
```diff
@@ -0,0 +1,90 @@
...
+
+${iconNames.map(icon => `import { ReactComponent as Inner${icon} } from "./${icon}";`).join("\n")}
+
+${iconNames.map(icon => `export const ${icon} = createOrbitIcon(Inner${icon}, "${icon}");`).join("\n")}
+
+${Object.entries(groupedIcons).map(([key, group])=> {
+        const name = getComponentName(key);
+
+        return `export const ${name} = createOrbitMultiVariantIcon(${group.map(icon => `Inner${getComponentName(icon.name, icon.sizeInName)}`).join(", ")}, "${name}");`;
+    }).join("\n")}
```
patricklafrance, line 32: Indentation issue

packages/components/scripts/generate-react-components.js
```diff
@@ -0,0 +1,90 @@
...
+import { createOrbitIcon, createOrbitMultiVariantIcon } from "../createOrbitIcon";
+
+${iconNames.map(icon => `import { ReactComponent as Inner${icon} } from "./${icon}";`).join("\n")}
+
+${iconNames.map(icon => `export const ${icon} = createOrbitIcon(Inner${icon}, "${icon}");`).join("\n")}
+
+${Object.entries(groupedIcons).map(([key, group])=> {
+        const name = getComponentName(key);
+
+        return `export const ${name} = createOrbitMultiVariantIcon(${group.map(icon => `Inner${getComponentName(icon.name, icon.sizeInName)}`).join(", ")}, "${name}");`;
```
patricklafrance, line 31: What is a "sizeInName" ?

packages/components/scripts/generate-react-components.js
```diff
@@ -0,0 +1,90 @@
...
+`;
+
+function getComponentName(fileName, size) {
+    const name = fileName.replace(/\.svg$/, "");
+
+    const options = { pascalCase: true };
+    let formatedName = camelCase(name, options);
+
+    formatedName = formatedName.replace(/^(Icon)/, "");
+    if(size){
```
patricklafrance, line 43: Space after an "if"

packages/components/scripts/generate-react-components.js
```diff
@@ -0,0 +1,90 @@
...
+        formatedName = formatedName.replace(new RegExp(`${size}$`, "g"), `Icon${size}`);
+    } else {
+        formatedName += "Icon";
+    }
+
+
+    return formatedName;
+}
+
+function generateSvgComponents(icons, dir){
```
patricklafrance, line 53: I would use "Icon" instead of "Svg"

packages/components/scripts/generate-icons.js
```diff
@@ -0,0 +1,20 @@
...
+const { clearDir } = require("../../../scripts/helper");
+const { optimize } = require("../../icons/scripts/optimize");
+const { generateSvgComponents, generateIndexFile } = require("./generate-react-components");
+
+const SRC_PATH = "../icons/src";
+const GENERATED_ICON_PATH = "./src/icons/src/generated-icons";
+
+console.log("\nBuild started...");
+
+optimize(SRC_PATH)
```
patricklafrance, line 11: Something weird with the naming here, I would have expect this code to start by calling something call "optimized" I would have expect something retrieving all the svg files.

Maybe `optimize` is having too many concerns?

packages/icons/scripts/optimize.js
alexasselin008: Since the checks are done after the file has been parsed by SVGO, i've took your adviced and separated the concerns. I load and optimize the svgs, then i validate the results before writing the files

packages/components/scripts/generate-react-components.js
alexasselin008: Not sure what the issue your seeing is

patricklafrance: image

packages/components/scripts/generate-react-components.js
```diff
@@ -0,0 +1,91 @@
...
+        ref: true,
+        svgoConfig: config
+    }, {
+        componentName: `Inner${name}`
+    });
+
+    fs.writeFileSync(`${path}/${name}.tsx`, iconTemplate(code));
+}
+
+function generateIndexFile(icons, dir) {
```
patricklafrance, line 80: I think it should be in his own file

patricklafrance:
The babel config at build/babel.config.js still have `babel-plugin-named-asset-import` configured, we shouldn't use it anymore.

It's probably because of the Storybook. I don't think the Storybook should still depend on this babel plugin. If we want to keep it because the document have a few custom svg, at least move `babel-plugin-named-asset-import` in a specific babel config for the Storybook project.

I would prefer not having to still depend on the plugin thought

patricklafrance:
Do we still need jest/svgr-mock.js with the new icons?

patricklafrance (approved):

Issue #932, opened 2022-03-31
🐛  Input placeholder font-size too small on Safari


fraincs:
## Describe the bug

In Safari the rendered placeholder input font-size is not right, it should be on par with the rendered one in Chrome.

![image](https://user-images.githubusercontent.com/361632/161060919-6fcf38fb-ce2c-47c2-afc7-01373bfb0d70.png)

## Steps to reproduce

open /?path=/story/text-input--default-story on Safari

## Expected results

The placeholder font size is the same across all browsers.

PR #933, opened 2022-03-31
Added css to placeholder in order to make Safari placeholder font-size the same as others browsers

fraincs:
Issue: 

## Summary

This aim to fix issue #932 

## What I did

Added a specific rule on the placeholder style declaration.

## How to test

Using Safari visit : /?path=/docs/text-input--default-story

### DIFF
--- a/packages/components/src/input/src/Input.css
+++ b/packages/components/src/input/src/Input.css
@@ -72,6 +72,7 @@
 .o-ui-input textarea::placeholder {
     color: var(--o-ui-text-alias-input-placeholder);
     opacity: 1;
+    font-size: var(--o-ui-fs-3);
 }
 
 .o-ui-input input:disabled::placeholder,


LaurentDario (approved):

packages/components/src/input/src/Input.css
```diff
@@ -72,6 +72,7 @@
 .o-ui-input textarea::placeholder {
     color: var(--o-ui-text-alias-input-placeholder);
     opacity: 1;
+    font-size: var(--o-ui-fs-3);
 }
```
patricklafrance, line 76: Why does a generic "Input.css" file have selectors specific to a textarea? It should go in a "TextArea.css" file.

Also, was it a bug specific to TextArea? Other inputs should have the same bug.

fraincs:
```
.o-ui-input input::placeholder,
.o-ui-input textarea::placeholder {
...
}
```

The font-size is applied to both the textarea and the input, since it's common to all inputs the CSS is in input.css. Managing two rules in two files that does the same thing is prone to errors when changes occurs, having to remember to change it in two files is counter-intuitive.

patricklafrance: Ah! The joy of seeing partial code 😂

PR #939, opened 2022-04-04
added background color to input wrapper

fraincs:
Issue: 

## Summary

When passing a backgroundColor style prop the input was receiving the value, not the wrapper this PR aims at fixing this problematic.

## What I did

Added backgroundColor to adaptInputStylingProps.


fraincs:
> If you add the backgroudColor prop you also have to add the backgroundColorActive/Focus/Hover props

Done

Issue #934, opened 2022-03-31
🐛  A card header doesn't center it's content vertically


fraincs:
## Describe the bug

When something other than text is present in a Card Header it's content is not centered versus the Heading. 

## Steps to reproduce

Use a component smaller than 24px in height in a ```<Heading>``` section. The Component is aligned to the top.

## Expected results

The Component is centered vertically. 

## Reproducible demo

Paste this in an Orbit sample code. And notice that the switch is not centered.

```
<Card>
    <Heading>NASA Headquarters</Heading>
    <Header><Switch></Switch></Header>
    <Content>
        <Paragraph>
            NASA Headquarters, officially known as Mary W. Jackson NASA Headquarters or NASA HQ and
            formerly named Two Independence Square, is a low-rise office building in the
            two-building Independence Square complex at 300 E Street SW in Washington, D.C.
        </Paragraph>
    </Content>
</Card>
```

PR #941, opened 2022-04-04
made sure that every elements in a card header and heading are centered

fraincs:
Issue: 

## Summary

#934 - A card header doesn't center it's content vertically 

## What I did

Set display flex to a heading and header and centered the items.

## How to test

Add a Switch in an Card heading and header


packages/components/src/card/src/Card.css
```diff
@@ -98,25 +103,25 @@
...
+.o-ui-card-vertical>.o-ui-card-aside {
     grid-column-start: 1;
     grid-column-end: 3;
     padding: var(--o-ui-sp-5);
 }
 
-.o-ui-card-vertical > .o-ui-card-illustration + .o-ui-card-aside,
-.o-ui-card-vertical > .o-ui-card-image + .o-ui-card-aside {
+.o-ui-card-vertical>.o-ui-card-illustration+.o-ui-card-aside,
+.o-ui-card-vertical>.o-ui-card-image+.o-ui-card-aside {
```
alexasselin008, line 124: There is a lot of css files with spaces around > and +. Should you fix them too?

fraincs: space should be there actually easier to read

patricklafrance: @fraincs have a look also in https://github.com/gsoft-inc/sg-orbit/pull/940/files then, space have been removed around the +

patricklafrance:
How can I test this if this now work? Was there something like a failing chromatic tests before the fix?

patricklafrance:
Most of the time when fixing something it's a good idea to had a test before to ensure we do not have any regression in the future.

LaurentDario (approved):

Issue #951, opened 2022-04-11
❗ Upload fonts in a CDN instead of self-hosting them in our applications.


alexasselin008:
## Description

The issue: https://stackoverflow.com/questions/71592690/next-js-preload-self-hosted-fonts-from-node-modules

I found no way to preload font when they are self hosted. It would be easier if we load the fonts from this url instead of hosting them : 
https://assets.sharegate.com/shared/brand/fonts/interphases/TT-Interphases-Var-Roman.woff2

PR #952, opened 2022-04-11
Use Fonts in a CDN instead of self-hosting them

alexasselin008:
PR for #951


fraincs:
What happens if the fonts are moved in the CDN, the whole solution becomes broken? Do we have the rights to host them and expose them here as anyone can link to them, we have a license since it's a paid font, lees of a concern since we already have the font available on Github.

patricklafrance:
I am also fine with the copyright risk. As @alexasselin008 mentionned, they are already available publicly.

I do share the risk corcerning a potential break if someone move the fonts on the CDN. The initial intent of these fonts wasn't to be used on our consumer facing apps so it might not be known by everyone that we are moving this way.

As we discussed privately, I would be more confortable using the fonts on this CDN if we formalize the ownership of those assets e.g. what use them, who can manage them, how to manage them etc..

fraincs (approved):

PR #962, opened 2022-05-02
Feature/958

fraincs:
Issue: 

## Summary

- Many components were missing a Guidelines section, this PR fixes this for components that have been documented by the design system team.
- Some components were documented but were not using the right structure.

## What I did

- Reworked the components documentation according to a new structure.
- Added documentation for newly written components documentation.

### DIFF
--- a/packages/components/src/accordion/docs/Accordion.stories.mdx
+++ b/packages/components/src/accordion/docs/Accordion.stories.mdx
@@ -26,11 +26,11 @@ import { InnerItem, Item } from "@components/collection";
 ### Accessibility
 
 - The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
-- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit heading element accordingly.
+- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit [Heading](?path=/docs/heading--default-story) element accordingly.
 
 ## Usage
 
-We do recommend that you specify a unique `key` prop for every `<Item>`. If you choose to omit the `key` prop, a key matching the `<Item>` position will be generated. For example, `"0"` would be the generated key for the first `<Item>` of an `<Accordion>` component without keys.
+> We do recommend that you specify a unique `key` prop for every `<Item>`. If you choose to omit the `key` prop, a key matching the `<Item>` position will be generated. For example, `"0"` would be the generated key for the first `<Item>` of an `<Accordion>` component without keys.
 
 ### Default
 
--- a/packages/components/src/accordion/docs/Accordion.stories.mdx
+++ b/packages/components/src/accordion/docs/Accordion.stories.mdx
@@ -21,11 +21,12 @@ import { InnerItem, Item } from "@components/collection";
     githubPath="/packages/components/src/accordion/src"
 />
 
-## Accessibility
+## Guidelines
 
-The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
+### Accessibility
 
-That heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit heading element accordingly.
+- The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
+- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit heading element accordingly.
 
 ## Usage
 
--- a/packages/components/src/accordion/docs/Accordion.stories.mdx
+++ b/packages/components/src/accordion/docs/Accordion.stories.mdx
@@ -26,11 +26,11 @@ import { InnerItem, Item } from "@components/collection";
 ### Accessibility
 
 - The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
-- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit [Heading](?path=/docs/heading--default-story) accordingly.
+- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit [heading](?path=/docs/heading--default-story) accordingly.
 
 ## Usage
 
-> We do recommend that you specify a unique `key` prop for every `<Item>`. If you choose to omit the `key` prop, a key matching the `<Item>` position will be generated. For example, `"0"` would be the generated key for the first `<Item>` of an `<Accordion>` component without keys.
+We do recommend that you specify a unique `key` prop for every `<Item>`. If you choose to omit the `key` prop, a key matching the `<Item>` position will be generated. For example, `"0"` would be the generated key for the first `<Item>` of an `<Accordion>` component without keys.
 
 ### Default
 
--- a/packages/components/src/accordion/docs/Accordion.stories.mdx
+++ b/packages/components/src/accordion/docs/Accordion.stories.mdx
@@ -26,7 +26,7 @@ import { InnerItem, Item } from "@components/collection";
 ### Accessibility
 
 - The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
-- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit [Heading](?path=/docs/heading--default-story) element accordingly.
+- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit [Heading](?path=/docs/heading--default-story) accordingly.
 
 ## Usage
 
--- a/packages/components/src/alert/docs/Alert.stories.mdx
+++ b/packages/components/src/alert/docs/Alert.stories.mdx
@@ -20,9 +20,11 @@ import { Heading, InnerHeading } from "@components/typography";
     githubPath="/packages/components/src/alert/src"
 />
 
-## Accessibility
+## Guidelines
 
-An alert comes with the `alertdialog` role and should be used to display a confirmation message, an alert, an error, or a warning.
+### Accessibility
+
+- An alert comes with the `alertdialog` role and should be used to display a confirmation message, an alert, an error, or a warning.
 
 It requires:
 - The focus to be trapped inside the alert.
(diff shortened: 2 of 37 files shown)


patricklafrance:
Not sure about the formatting here

image

patricklafrance:
I don't have a strong opinion if we want to uppercase or not the components name but for most of the documentation I wrote, other than a few exceptions, I lowercase the components name. I think it should be consistent.

image

patricklafrance:
image

patricklafrance:
For pages with "Guidelines" I wonder where this type of information should stand at in the page hierarchy? I guess it's the first information a dev should read?

image

patricklafrance:
Seems to have an extra "to"

image

patricklafrance:
Components name casing should be consistent.

image

patricklafrance:
I am not sure this is useful information for a dev?

image

packages/components/src/accordion/docs/Accordion.stories.mdx
```diff
@@ -21,15 +21,16 @@ import { InnerItem, Item } from "@components/collection";
...
 
-That heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit heading element accordingly.
+- The accordion header that shows and hides the content must be an [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).
+- The heading element should have a level that is appropriate for the information architecture of the page. You can choose the appropriate level by selecting one of Orbit [Heading](?path=/docs/heading--default-story) element accordingly.
 
 ## Usage
 
-We do recommend that you specify a unique `key` prop for every `<Item>`. If you choose to omit the `key` prop, a key matching the `<Item>` position will be generated. For example, `"0"` would be the generated key for the first `<Item>` of an `<Accordion>` component without keys.
+> We do recommend that you specify a unique `key` prop for every `<Item>`. If you choose to omit the `key` prop, a key matching the `<Item>` position will be generated. For example, `"0"` would be the generated key for the first `<Item>` of an `<Accordion>` component without keys.
```
LaurentDario, line 33: If we want to put the emphasis on Accessibility within Orbit, we should avoid these Accessibility failures.
Markdown is not the best for styling content as it has limited option to highlight or indent content, but that solution (`>`) is a `blockquote`.
And the usage is a [failure of SC 1.3.1](https://www.w3.org/TR/2007/WD-WCAG20-TECHS-20070517/Overview.html#G115)
> Failure Example 3: Using blockquote elements to provide additional indentation

Blockquotes should only be used to refer to sourced content

Would it be ok to leave it as a plain text?
(This applies to other usage in the PR)

packages/components/src/alert/docs/Alert.stories.mdx
```diff
@@ -20,24 +20,24 @@ import { Heading, InnerHeading } from "@components/typography";
...
-Label and description can be overwritten with the use of `aria-label` and `aria-describedby` props on the alert.
+### Accessibility
 
-### Focus
+- `aria-label` and `aria-describedby` defaults to the alert `<Heading>` and `<Content>` values.
+- You can overwrite them by setting `aria-label` and `aria-describedby` props on the alert.
 
-If the alert calls for an action, make sure to set the focus correctly to avoid any unwanted action.
-To do so, you can use the `autoFocusButton` prop on the alert to specify which control should have the initial focus.
+[aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label), [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)
```
LaurentDario, line 34: Un truc qui pourrait être testé ici, c'est le [Footnotes]([aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)).

```
- `aria-label`[^1] and `aria-describedby`[^2] defaults to the alert `<Heading>` and `<Content>` values.
- You can overwrite them by setting `aria-label` and `aria-describedby` props on the alert.

[^1]: [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)
[^2]: [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)
```

donnerai

- `aria-label`[^1] and `aria-describedby`[^2] defaults to the alert `<Heading>` and `<Content>` values.
- You can overwrite them by setting `aria-label` and `aria-describedby` props on the alert.

[^1]: [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) (mdn web docs)
[^2]: [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) (mdn web docs)

packages/components/src/image/docs/Image.stories.mdx
```diff
@@ -18,9 +18,13 @@ import { Inline } from "@components/layout";
     githubPath="/packages/components/src/image/src"
 />
 
-## Accessibility
+## Guidelines
 
-An image must have an `alt` attribute.
+### Accessibility
+
+- An image must have an `alt` attribute.
```
LaurentDario, line 25: While this is true I think it's important to specify that the `alt` attribute **can** be empty.
It is written in the documentation linked bellow but we know everyone's not gonna click it
> If the image doesn't require a fallback (such as for an image which is decorative or an advisory icon of minimal importance), you may specify an empty string ("").

packages/components/src/icons/docs/Icon.stories.mdx
```diff
@@ -19,9 +19,13 @@ import IndexFileUsage from "./IndexFileUsage.mdx";
     githubPath="/packages/components/src/icons/src"
 />
 
-## Accessibility
+## Guidelines
 
-When applicable, an accessible name should be provided through `aria-label` to indicate the **purpose** of the icon in it's **current context**. For more information see [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute).
+### Accessibility
+
+- When applicable, an accessible name should be provided through `aria-label` to indicate the **purpose** of the icon in it's **current context**.
```
LaurentDario, line 26: I would also add that if the icon is not adding any information and doesn't need to be announced by a screen reader, just like Img uses an empty alt, icons should use `aria-hidden="true"` to be ignored by screenreaders while still displaying in the UI.
cf: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden#example

packages/components/src/alert/docs/Alert.stories.mdx
LaurentDario: You could also decide to make all the links footnotes and have the footnote at the very very bottom of the page like:

Source:
```
## Guidelines

### When to use

- To display a confirmation message, an error, or a warning. This is a link[^when] of the first section

### Accessibility

- `aria-label`[^a11y] and `aria-describedby`[^a11y] defaults to the alert `<Heading>` and `<Content>` values.
- You can overwrite them by setting `aria-label` and `aria-describedby` props on the alert.

## Usage

Use it like you want to use it!

[^when]: [my random link](*)
[^a11y]: [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label), [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)
```

## Guidelines

### When to use

- To display a confirmation message, an error, or a warning. This is a link[^when] of the first section

### Accessibility

- `aria-label`[^a11y] and `aria-describedby`[^a11y] defaults to the alert `<Heading>` and `<Content>` values.
- You can overwrite them by setting `aria-label` and `aria-describedby` props on the alert.

## Usage

Use it like you want to use it!

[^when]: [my random link](*)
[^a11y]: [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label), [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)

LaurentDario (approved):
A few comments on accessibility, links and use of blockquote throughout the doc.
Something that would be good to have is an icon or a way to differentiate links that are bringing you to external documentation (like mdn) from the links that make you jump inside the content of Orbit itself (within component or from component to component)

fraincs:
> Not sure about the formatting here

the idea is to be able to have a reference to these elements without adding a link in the `text` as it's not supported, would you push it to the right?

packages/components/src/accordion/docs/Accordion.stories.mdx
patricklafrance: You're right.

What do you suggest as an alternative to highlight a particular piece of content?

packages/components/src/image/docs/Image.stories.mdx
LaurentDario: With the new branding, we use mostly SVG illustration so it's not so much the case anymore, but in most apps we build images are used in a decorative way and should use an empty `alt`. Some examples I found:
`Image` components used
image
image
`SvgImage` components used
image
image

fraincs:
> I don't have a strong opinion if we want to uppercase or not the components name but for most of the documentation I wrote, other than a few exceptions, I lowercase the components name. I think it should be consistent.

The goal was to be able to easily guess if we are talking about an HTML element or an Orbit Component, the Capitalized word makes it easy to differentiate. A tooltip is not a Tooltip.

image

packages/components/src/alert/docs/Alert.stories.mdx
fraincs: With the fact that linking in a `code` tag would you still suggest footnotes?

image

LaurentDario: I think that that in-code link is a good option

packages/components/src/image/docs/Image.stories.mdx
patricklafrance: Make sense then!

patricklafrance:
> > I don't have a strong opinion if we want to uppercase or not the components name but for most of the documentation I wrote, other than a few exceptions, I lowercase the components name. I think it should be consistent.
> 
> The goal was to be able to easily guess if we are talking about an HTML element or an Orbit Component, the Capitalized word makes it easy to differentiate. A tooltip is not a Tooltip.
> 
> image

Personally, it's more intuitive if you suffix the name with "element" rather than an implicit convention. For example, if I read "header element" I understand it's an html element and not a React component.

If you want to go this way fine but in order for the reader to understand you'll have to use this convention everywhere.

fraincs:
> > > I don't have a strong opinion if we want to uppercase or not the components name but for most of the documentation I wrote, other than a few exceptions, I lowercase the components name. I think it should be consistent.
> > 
> > 
> > The goal was to be able to easily guess if we are talking about an HTML element or an Orbit Component, the Capitalized word makes it easy to differentiate. A tooltip is not a Tooltip.
> > image
> 
> Personally, it's more intuitive if you suffix the name with "element" rather than an implicit convention. For example, if I read "header element" I understand it's an html element and not a React component.
> 
> If you want to go this way fine but in order for the reader to understand you'll have to use this convention everywhere.

I was heading in this direction (suffix the name with "element"). Since most of the time we mention the component name this won't render the copy too verbose. Approved.

packages/components/src/accordion/docs/Accordion.stories.mdx
patricklafrance: It depends, the idea is to highlight the important stuff the user should see first. Without any specific I am not sure the result would be the same.

Furthermore, the solution you proposed might works for components pages but not for the other documentation pages like "Style props", or "installation"

Issue #964, opened 2022-05-04
🚀  Recover documentation highlights that have been lost in an accessibility purge


fraincs:
## Description

When the doc has been reworked some information that was highlighted using blockquotes `> ...` has lost it's appearance. Therefore we need a way to put the focus on important pieces of information be it a colored block, an emoji 👉🏻 , a NOTE: or some other way.

Attached some inspiration: 

![image](https://user-images.githubusercontent.com/361632/166725678-a84485ad-50d5-43c6-b0ae-d51b38c70493.png)
![image](https://user-images.githubusercontent.com/361632/166725697-48a28c69-9f92-47f0-bc96-e1a72626d22a.png)
![image](https://user-images.githubusercontent.com/361632/166725717-6ebdb384-fcde-4cfc-9c65-c1c05c863346.png)

pierre-lambert:
Isn't "Message" fulfilling this need? https://orbit.sharegate.design/?path=/docs/message--default-story

fraincs:
@pierre-lambert in some instances I think a Message is not fulfilling it's duty let's take this : 

In Orbit, 1 rem unit = 16 pixels.

This isn't a message for the user but more of a Tip what do you think?

Wouldn't something like this make more sense?

💡  In Orbit, 1 rem unit = 16 pixels.

versus this

![image](https://user-images.githubusercontent.com/361632/166934924-16913b9c-a01f-4e2d-8226-334caf2c36e9.png)

PR #966, opened 2022-05-06
Feature/964

fraincs:
Issue: 

Fix for issue #964 

## Summary

Recovered documentation highlights that have been lost in an accessibility purge.

## What I did

Re-highlighted the important sections of the doc with markdown. The final result is a div instead of a blockquote since a blockquote was not the right intent.

### DIFF
--- a/storybook/styles/docs.css
+++ b/storybook/styles/docs.css
@@ -332,9 +332,11 @@
 
 /* ELEMENTS | BLOCKQUOTE */
 .sbdocs.sbdocs-blockquote {
-    padding: var(--o-ui-sp-3) var(--o-ui-sp-4) var(--o-ui-sp-3) var(--o-ui-sp-3);
-    background-color: var(--o-ui-purple-1);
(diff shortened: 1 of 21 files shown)


storybook/styles/docs.css
```diff
@@ -332,9 +332,11 @@
 
 /* ELEMENTS | BLOCKQUOTE */
 .sbdocs.sbdocs-blockquote {
```
patricklafrance, line 334: Is the idea here to still support blockquote? If so, you can't really render any since the markdown will defer to the Highlight component. Therefore, I would remove the CSS.

patricklafrance (approved):

storybook/styles/docs.css
fraincs: We are not planning to support it as quoting someone is not in the projects ;) I'll remove it. If we ever need to use a blockquote, which we doubt, this might change for now removing.

Issue #977, opened 2022-05-24
🚀 Document Button Usage specs


fraincs:
Button Documentation need to be revised and new sections needs to be added: Related, Accessibility, Best Practices and Anatomy.

PR #978, opened 2022-05-25
added documentation to the button mdx

fraincs:
Issue: 

## Summary

Fix for issue #977. 

## What I did

Added sections (Accessibility / when to use / VS  as well as content).

## How to test

?path=/docs/button--default-story#toggle-button

### DIFF
--- a/packages/components/src/button/docs/Button.stories.mdx
+++ b/packages/components/src/button/docs/Button.stories.mdx
@@ -34,6 +34,28 @@ import { Text } from "@components/typography";
     githubPath="/packages/components/src/button/src"
 />
 
+## Guidelines
+
+### When to use
+
+- To trigger an action such as submitting a form, exporting data.
+
+### Content
+
+Buttons should:
+
+- Use strong, actionable verbs.
+- Be positionned in consistent locations.
+
+### Accessibility
+
+- Use the disabled prop to set the disabled state of the button.
+
+### Button vs Link vs Switch
+
+- When you need to link to another page, use a [link](?path=/docs/link--default-story) component.
+- To activate or deactivate a feature, use the [switch](?path=/docs/switch--default-story) instead of a single [toggle button](?path=/docs/button--default-story#toggle-button).
+
 ## Usage
 
 ### Default
@@ -48,7 +70,7 @@ A default button.
 
 ### Icon
 
-A button can contain [icons](?path=/docs/icon-gallery--page).
+A button can contain [icons](?path=/docs/icon-gallery--page). We should avoid using icons on small buttons.
 
 <Preview>
     <Story name="icon">
@@ -100,7 +122,7 @@ A button can vary in size.
 
 ### Disabled
 
-A button can be disabled.
+Used to disable a button that isn’t usable.
 
 <Preview>
     <Story name="disabled">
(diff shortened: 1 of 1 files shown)


packages/components/src/button/docs/Button.stories.mdx
```diff
@@ -34,6 +34,28 @@ import { Text } from "@components/typography";
...
+### When to use
+
+- To trigger an action such as submitting a form, exporting data.
+
+### Content
+
+Buttons should:
+
+- Use strong, actionable verbs.
+- Be positionned in consistent locations.
```
patricklafrance, line 48: Is it really related to content?

packages/components/src/button/docs/Button.stories.mdx
```diff
@@ -34,6 +34,28 @@ import { Text } from "@components/typography";
...
+
+### Accessibility
+
+- Use the disabled prop to set the disabled state of the button.
```
patricklafrance, line 52: Is it really an accessibility tip? It seems quite obvious to use a disabled prop to disabled a button

packages/components/src/button/docs/Button.stories.mdx
```diff
@@ -34,6 +34,28 @@ import { Text } from "@components/typography";
...
+
+### Button vs Link vs Switch
+
+- When you need to link to another page, use a [link](?path=/docs/link--default-story) component.
+- To activate or deactivate a feature, use the [switch](?path=/docs/switch--default-story) instead of a single [toggle button](?path=/docs/button--default-story#toggle-button).
```
patricklafrance, line 57: I would split those in 2 sections since a basic "button" is not related at all with a switch.

I would do

- Button vs Link
- ToggleButton vs Switch

packages/components/src/button/docs/Button.stories.mdx
```diff
@@ -100,7 +122,7 @@ A button can vary in size.
 
 ### Disabled
 
-A button can be disabled.
+Used to disable a button that isn’t usable.
```
patricklafrance, line 125: Not sure about this change. The previous one seems better. I mean, disabling a button is always because he should not be usable.

packages/components/src/button/docs/Button.stories.mdx
```diff
@@ -48,7 +70,7 @@ A default button.
 
 ### Icon
 
-A button can contain [icons](?path=/docs/icon-gallery--page).
+A button can contain [icons](?path=/docs/icon-gallery--page). We should avoid using icons on small buttons.
```
patricklafrance, line 73: I would be more directive: "Avoid using icons on small buttons"

packages/components/src/button/docs/Button.stories.mdx
fraincs: The position no, you are right.

packages/components/src/button/docs/Button.stories.mdx
fraincs: It's indeed probably too obvious, I was trying to avoid not to have accessibility tips.

patricklafrance: Could you write something about the tooltips here then? It makes sense to me to mention that here

patricklafrance (approved):

PR #1002, opened 2022-07-04
Feature/976

fraincs:
Issue: 

## Summary

Trying to override a Lozenge color and backgroundColor was throwing a typescript error as well as not working in some situations (when a lozenge had the highlight prop).

## What I did

Removed the typescript omissions in the component declaration. Modified the CSS of lozenge to work when highlighted.

## How to test

/?path=/story/chromatic-lozenge--inherit-parent-text-properties and by adding a color and backgroundColor prop to a lozenge

### DIFF
--- a/packages/components/src/lozenge/docs/Lozenge.stories.mdx
+++ b/packages/components/src/lozenge/docs/Lozenge.stories.mdx
@@ -19,7 +19,6 @@ import { Text } from "@components/typography";
     githubPath="/packages/components/src/lozenge/src"
 />
 
-
 ## Guidelines
 
 ### When to use
--- a/packages/components/src/lozenge/src/Lozenge.css
+++ b/packages/components/src/lozenge/src/Lozenge.css
@@ -54,7 +54,7 @@
 }
 
 .o-ui-lozenge.o-ui-lozenge-informative.o-ui-lozenge-highlight .o-ui-lozenge-text {
-    color: var(--o-ui-white);
+    color: inherit;
 }
 
 .o-ui-lozenge.o-ui-lozenge-informative.o-ui-lozenge-highlight .o-ui-lozenge-icon {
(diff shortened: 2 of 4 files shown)


packages/components/src/lozenge/src/Lozenge.css
```diff
@@ -54,7 +54,7 @@
 }
 
 .o-ui-lozenge.o-ui-lozenge-informative.o-ui-lozenge-highlight .o-ui-lozenge-text {
-    color: var(--o-ui-white);
+    color: inherit;
```
tjosepo, line 57: Do we need this property?

packages/components/src/lozenge/docs/Lozenge.stories.mdx
```diff
@@ -19,7 +19,6 @@ import { Text } from "@components/typography";
     githubPath="/packages/components/src/lozenge/src"
 />
 
-
```
LaurentDario, line 22: Should we add a mention that changing the background or color of the Lozenge can lead to accessibility failures? Something along the lines of 
> Lozenge's variants follow the contrast guidelines. Changing the background or text color might cause contrast check failures.
(Not sure `contrast check failures` is the right way to put it)

LaurentDario (approved):
Just a documentation suggestion.

packages/components/src/lozenge/docs/Lozenge.stories.mdx
wasim-abuzaher-sg: errors? issues? noncompliance?

fraincs: actually we don't want people to do it but we understand that we might need to do it and at least it won't fail. I will add it to the accessibility section of the lozenge doc, but we don't document it for other components like button / tag who also support changing the backgroundColor and color.

Issue #979, opened 2022-05-25
🚀 Small Lozenge should not use lowercase text as it's hard to read


fraincs:
## Description

When in a small Lozenge, text is hard to read if not in all caps. All caps should be enforced in all small Lozenge.

## Checklist

- [ ] Chromatic tests
- [ ] Documentation
- [ ] Figma

PR #1005, opened 2022-07-08
small lozenge text should be uppercase when small / adjusted the doc …

fraincs:
…and the component

Issue: 

#979 - Small Lozenge should not use lowercase text as it's hard to read.

## Summary

A small Lozenge should not use lowercase text as it makes it hard to read for the users. We now enforce an uppercase on small lozenge text.

## What I did

When lozenge component size is small we set the textTransform property to `uppercase` by default.

### DIFF
--- a/packages/components/src/lozenge/src/Lozenge.tsx
+++ b/packages/components/src/lozenge/src/Lozenge.tsx
@@ -45,6 +45,10 @@ const textSize = createSizeAdapter({
 });
 /* eslint-enable sort-keys, sort-keys-fix/sort-keys-fix */
 
+const textTransform = function (size) {
+    return size === "sm" ? "uppercase" : null;
+};
+
 export function InnerLozenge({
     as = DefaultElement,
     children,
@@ -68,7 +72,8 @@ export function InnerLozenge({
         },
         text: {
             className: "o-ui-lozenge-text",
-            size: textSize(sizeValue)
+            size: textSize(sizeValue),
+            textTransform: textTransform(sizeValue)
         }
     }), [sizeValue]));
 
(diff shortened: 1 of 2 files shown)


packages/components/src/lozenge/src/Lozenge.tsx
```diff
@@ -45,6 +45,10 @@ const textSize = createSizeAdapter({
 });
 /* eslint-enable sort-keys, sort-keys-fix/sort-keys-fix */
 
+const textTransform = function (size) {
+    return size === "sm" ? "uppercase" : null;
```
LaurentDario, line 49: Not familiar with this but does that mean that we would set `textTransform: null`? Do we want it to be `textTransform: undefined`?

LaurentDario (approved):

packages/components/src/lozenge/src/Lozenge.tsx
```diff
@@ -68,7 +72,8 @@ export function InnerLozenge({
         },
         text: {
             className: "o-ui-lozenge-text",
-            size: textSize(sizeValue)
+            size: textSize(sizeValue),
+            textTransform: textTransform(sizeValue)
```
patricklafrance, line 76: Pretty sure we decided when we implemented style props that Orbit component wouldn't use them to prevent rendering components with 47567 inline props.

Here you could easily do it in CSS, you already have the `o-ui-lozenge-text` class and also a size classe rendered.

fraincs: Totally, wanted to do the "clean" way as in CSS this could easily be overridden but I can do it in CSS no problem.

fraincs: Changed it to CSS

PR #1006, opened 2022-07-08
converted 3 stories to CSF

fraincs:
Issue: 

## Summary

Converted 3 stories to the CSF format.


patricklafrance:
Pretty straighforward so far.

You should do Button before going farther, this is one with which you'll hit a wall :)

https://github.com/gsoft-inc/sg-orbit/blob/master/packages/components/src/button/tests/chromatic/Button.chroma.jsx

Issue #1019, opened 2022-08-08
🐛 TextArea's text doesn't have the right color in dark mode


alexasselin008:
## Describe the bug

When using a TextArea in dark mode, the text is black instead of being primary.
![image](https://user-images.githubusercontent.com/38871812/183451896-b15d89ce-d7b0-4d8d-9b00-8be46e13249e.png)

## Steps to reproduce

1- Open the doc https://orbit.sharegate.design/?path=/docs/textarea--default-story&globals=colorScheme:dark#default
2- Change the color scheme to dark mode
3- type something in the text area
4- notice the text is black

## Expected results

The text should be color primary

alexasselin008:
A visual regression test should be added in order for this to never happen again!

fraincs:
IIRC we are not testing dark mode as it would double the number of Chromatic tests

PR #1021, opened 2022-08-16
textarea now inherit input color

fraincs:
Issue: 

## Summary

#1019 - TextArea's text doesn't have the right color in dark mode

## What I did

Textarea now inherits the input text color, since the hiearchy is different from the text input where an input has it's style directly applied, a textarea has a wrapper.


alexasselin008 (approved):

alexasselin008 (changes requested):
add a chromatic test ?

alexasselin008 (approved):

PR #1023, opened 2022-09-06
Fix CSS @import declaration ordering

tjosepo:
Issue: Fix CSS @import declaration ordering

## Summary

According to [the specs](https://drafts.csswg.org/css-cascade-5/#at-import
): 
> Any `@import` rules must precede all other valid at-rules and style rules in a style sheet (ignoring `@charset` and `@layer` statement rules)

The `components/src/index.css` file didn't follow this rule.

While most build tools are able to "do what the developer expects", it's still invalid code, and some tools (notably, Vite) will give out warning messages because of this.

## What I did

Moved the content of `index.css` around to respect the specs.

### DIFF
--- a/packages/components/src/index.css
+++ b/packages/components/src/index.css
@@ -1,3 +1,56 @@
+@import "./styling/index.css";
+
+/*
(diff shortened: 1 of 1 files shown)


fraincs (approved):
Good job.

PR #1029, opened 2022-10-03
Feature/1028

fraincs:
Reworked the css declarations using first-child as it could be unsafe in SSR. This has a downside as first-of-type(which is what Next is suggesting using) only works with elements and not classes. Rendering the CSS more flimsy, this is not an issue as we control the elements in Orbit, but something to be aware of if we ever change the DOM of certain components.

### DIFF
--- a/packages/components/src/accordion/src/Accordion.css
+++ b/packages/components/src/accordion/src/Accordion.css
@@ -24,10 +24,6 @@
     padding-bottom: var(--o-ui-sp-2);
 }
 
-.o-ui-accordion-header:first-of-type {
-    margin-top: 0;
-}
-
 /* TRIGGER */
 .o-ui-accordion-trigger {
     outline: transparent;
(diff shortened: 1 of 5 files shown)


patricklafrance:
Very cool that you are handling this issue.

I wonder thought how safer it is by specifying the type like this instead of classes? I feel less safe with this changes than before haha.

fraincs:
> Very cool that you are handling this issue.
> 
> I wonder thought how safer it is by specifying the type like this instead of classes? I feel less safe with this changes than before haha.

I understand your concern, that said, the *-of-type selectors are suppose to be for elements and not classes, it kinda works although not officially supported (https://developer.mozilla.org/docs/Web/CSS/:last-of-type), which means that it could break anytime. Doing a class:last-of-type will select the last element with the class, which might be exactly what we want but since it's not to spec, I'm a little frisky to use and hope browsers does not change the way they treat this case.

fraincs:
@patricklafrance looking at this it seems it's more a Storybook / Emotion issue (https://github.com/storybookjs/storybook/issues/18103?fbclid=IwAR3N3p-xr2VYoLjjWhII_ErTfdLT_uiTtwCB6jWxnCJvrrytJg_anen1SXI) I don't see why compiled CSS would be affected by this.

patricklafrance:
Ahhh, great find @fraincs 

It's not an Orbit issue then.

PR #1030, opened 2022-10-07
added fluid prop to checkbox switch and radio

fraincs:
Added the option to use a fluid prop on Switch/Checkbox and Radio. This is meant to be undocumented as it's not a common use case nor one we want to promote.

### DIFF
--- a/packages/components/src/checkbox/src/Checkbox.css
+++ b/packages/components/src/checkbox/src/Checkbox.css
@@ -215,3 +215,8 @@
     margin-top: var(--o-ui-sp-1);
     margin-bottom: var(--o-ui-sp-1);
 }
+
+/* FLUID */
+.o-ui-checkbox-fluid {
+    width: 100%;
+}
--- a/packages/components/src/checkbox/src/Checkbox.tsx
+++ b/packages/components/src/checkbox/src/Checkbox.tsx
@@ -1,7 +1,7 @@
 import { AbstractInputProps } from "../../input";
 import { Box } from "../../box";
 import { ChangeEvent, ChangeEventHandler, ComponentProps, forwardRef, useMemo } from "react";
-import { OmitInternalProps, isNil, mergeProps, omitProps, resolveChildren, useChainedEventCallback, useCheckableProps, useSlots } from "../../shared";
+import { OmitInternalProps, cssModule, isNil, mergeProps, omitProps, resolveChildren, useChainedEventCallback, useCheckableProps, useSlots } from "../../shared";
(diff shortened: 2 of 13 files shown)


patricklafrance:
I wonder why are you adding this option if it is something which is not a common use case nor something we want to promote?

fraincs:
> I wonder why are you adding this option if it is something which is not a common use case nor something we want to promote?

image

We have some situations like here that still would benefit from it.

alexasselin008:
> I wonder why are you adding this option if it is something which is not a common use case nor something we want to promote?

Also, the property is already available on Checkboxes, so having it on radios makes the API more consistent

patricklafrance:
Got it.

Honnestly, I don't think making them fluid is the way to accomplish this.

I think it should be a new component which composed a checkbox/radio with other elements. Something like a RadioTile and CheckboxTile.

Does it make sense?

fraincs:
> Got it.
> 
> Honnestly, I don't think making them fluid is the way to accomplish this.
> 
> I think it should be a new component which composed a checkbox/radio with other elements. Something like a RadioTile and CheckboxTile.
> 
> Does it make sense?

Makes sense, and upon further discussions the issue will have to be closed and rephrased since a fluid Form pass fluid to all its children and it's not something you want to do(rendering all checkbox / switch radio as fluid in a fluid form).

patricklafrance:
Ok!

While thinking about it thought, RadioTile and CheckboxTile might not be the best names as it would be confusing with our current Tile component which render a Radio or Checkbox depending if it's a single select or multiselect behavior.

I wonder thought if it make sense to keep in the DS the actual Selectable tiles:

image

image

And also add the one you referenced in your screenshot. In my opinion, they seem to serve the same purpose and should probably be merged.

fraincs:
They indeed share the same DNA and could probably be merged down the line where one could use a checkbox tile / radio tile or switch tile(if we judge it makes sense)

patricklafrance:
Still, you have shown in your screenshot a component which should probably be part of Orbit. Is there any plan to integrate it? Is it part of a backlog somewhere?

fraincs:
> I wonder thought if it make sense to keep in the DS the actual Selectable tiles:

Tiles are selectables entities, you are suggesting/wondering if we should split these in two (RadioTiles and SelectTiles)? And deprecate the Tile component? Is my understanding right?

fraincs:
An issue has been opened regarding our discussions

https://github.com/gsoft-inc/sg-orbit/issues/1039

PR #1037, opened 2022-10-11
BREAKING: Add support for React 18

tjosepo:
Issue: N/A

## Summary

This PR adds support for React 18, and drops support for React <=17.

## What I did

- Update Storybook to support React 18.
- Update `@types/react` to support React 18.
- Update `@testing-library` to support React 18.
- Removed `@testing-library/react-hooks`, since the version that supports React 18 has been merged into `@testing-library/react`
- Fix tests broken by React 18 and the new version of `@testing-library`, mostly caused by `act(() => ...)` not being awaited.
- Removed `@reach/useId`, to use React 18's `useId`.
- ~Add a polyfill for `useId`, for React versions lower than 18.~ Not anymore. We're okay with this being a **breaking change.**

## How to test

- Is this testable with Jest or Chromatic screenshots? ✅

- Does this need an update to the documentation? ❌


packages/components/src/shared/src/useId.ts
```diff
@@ -1,7 +1,23 @@
...
+        return undefined;
+    }
+
+    return id;
+};
+
+const useAutoId = useIdReact ?? useIdPolyfill;
 
 export function useId(userId?: string, prefix?: string) {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
alexasselin008, line 20: this disable is probably not necessary anymore

packages/components/src/overlay/tests/jest/usePopup.test.tsx
```diff
@@ -455,8 +455,8 @@ test("closing the popup with esc keypress return the focus to the trigger", asyn
         />
     );
 
-    act(() => {
-        userEvent.click(getByTestId("trigger"));
+    await act(() => {
```
alexasselin008, line 458: all other lines have the format: 

```
    await act(() => userEvent.click(getByTestId("trigger")););
```

no need for the brackets and the return

packages/components/src/collection/src/getItemText.tsx
```diff
@@ -3,7 +3,7 @@ import { getRawSlots, isNil } from "../../shared";
 
 export function getItemText(item: CollectionNode): string {
     if (isDivider(item) || isItem(item)) {
-        const { stringValue, text } = getRawSlots(item?.content, ["text"]);
+        const { stringValue, text } = getRawSlots(<>{item?.content}</>, ["text"]);
```
alexasselin008, line 6: This is odd, i don't understand why we need to specify fragments.

alexasselin008: After investigation, I saw that the item.content can by either a ReactElement or a ElementType[].
getRawSlots receive a ReactNode

 ElementType is not part of ReactNode, so wrapping it makes it work, but the issue should be fixed

The typing here should be fixed another way

packages/components/src/shared/src/useId.ts
```diff
@@ -1,7 +1,23 @@
 import { isNil } from "./assertions";
-import { useId as useAutoId } from "@reach/auto-id";
+import { useId as useIdReact, useState } from "react";
+
+// Polyfill to support useId on React versions below 18.0.0, doesn't work on server
+let currentId = 0;
```
alexasselin008, line 5: wouldn't it be safer to still use the import { useId as useAutoId } from "@reach/auto-id"; in the meantime as the polyfill?

alexasselin008:
We whould probably update the Installation documentation to show the React 18 way of rendering the app with the new root.

storybook/main.js
```diff
@@ -43,7 +43,10 @@ const config = {
             }
         }
     ],
-    webpackFinal: customizeWebpack
+    webpackFinal: customizeWebpack,
+    reactOptions: {
+        strictMode: isDebug
```
alexasselin008, line 48: shouldnt this be true all the time since its a debug only configuration?

packages/components/src/shared/src/useId.ts
alexasselin008: As mentionned, i also wouldn't mind not having a polyfill and force all the clients to update to react 18. We rarely support older version of the apps

packages/components/src/tag/src/TagList.tsx
```diff
@@ -67,7 +67,7 @@ function TagItem({
                 }
             )}
         >
-            {item.content}
+            <>{item.content}</>
```
patricklafrance, line 70: Hmm, why?

storybook/decorators/withBackgroundMatchingColorScheme.jsx
```diff
@@ -36,5 +36,5 @@ export function withBackgroundMatchingColorScheme(Story, context) {
         }
     }, [colorScheme, context]);
 
-    return <Story />;
+    return story();
```
patricklafrance, line 39: That's an odd one. Is it related to something changed in a newer version of SB?

packages/components/src/shared/src/useId.ts
patricklafrance: I agree with Alex regarding not supporting app using React 17 and lower. Maybe it could work for useId but what about using other new features from React 18?

I think it's quite limiting and apps could afford using an older version of Orbit until they migrate to React 18 (which should not be a significant investment anyway)

patricklafrance:
@alexasselin008 wasn't there issues you found out earlier while investigating React 18 `useId` function?

patricklafrance:
I was expecting to saw usage of the new transition API. Is there any reason why you haven't use it?

packages/components/src/collection/src/getItemText.tsx
tjosepo: Do you have any suggestions for a fix?

I'm not too familiar with ElementType, ReactElement, etc.

packages/components/src/tag/src/TagList.tsx
tjosepo: Same issue as in `getTextItem`. The type of `item.content` isn't `ReactNode`, but wrapping it in a fragment casts it into a `ReactNode`.

storybook/decorators/withBackgroundMatchingColorScheme.jsx
tjosepo: Yep, there were some rendering issues when using `<Story />` instead of `story()`.

If I remember correctly, the error "Rendered more hooks than during the previous render" appeared in React 18 when using the new root API, Strict Mode, and `<Story />`.

tjosepo:
> @patricklafrance  I was expecting to see usage of the new transition API. Is there any reason why you haven't use it?

I felt that using `useTransition` was out of scope for this PR. I wanted to focus on updating having React 18 work, without using any of the new features. I made an exception for useId, since it was trivial to add, and I had to do something about `@reach/auto-id` not supporting React 18 without the use of  `--legacy-peer-deps`.

tjosepo:
> @patricklafrance I agree with Alex regarding not supporting app using React 17 and lower. Maybe it could work for useId but what about using other new features from React 18?
>
> I think it's quite limiting and apps could afford using an older version of Orbit until they migrate to React 18 (which should not be a significant investment anyway)

I agree that we should drop React 17 ASAP.

I initally felt that having a transitional release that supports both React 18 and React 17 might be a good thing to have, but I'm not too sure about it anymore 🤔I don't have any strong feelings about it one way or another.

The next release should definitely drop React 17,

storybook/decorators/withBackgroundMatchingColorScheme.jsx
patricklafrance: It's odd since Storybook official docs example still say tp use `<Story />` not `story()`

https://storybook.js.org/docs/react/writing-stories/decorators#global-decorators

patricklafrance:
> > @patricklafrance I agree with Alex regarding not supporting app using React 17 and lower. Maybe it could work for useId but what about using other new features from React 18?
> > I think it's quite limiting and apps could afford using an older version of Orbit until they migrate to React 18 (which should not be a significant investment anyway)
> 
> I agree that we should drop React 17 ASAP.
> 
> I initally felt that having a transitional release that supports both React 18 and React 17 might be a good thing to have, but I'm not too sure about it anymore 🤔I don't have any strong feelings about it one way or another.
> 
> The next release (after this one) should definitely drop React 17,

I think the release containing this PR should already drop React 17 support as Orbit have been stable for a while. If there unreleased fixes which should be made available to apps, they can be released before the PR is merged.

tjosepo:
>I think the release containing this PR should already drop React 17 support as Orbit have been stable for a while. If there unreleased fixes which should be made available to apps, they can be released before the PR is merged.

Let's go for that! I'll update this PR tomorrow to drop React 17.

patricklafrance:
> > @patricklafrance  I was expecting to see usage of the new transition API. Is there any reason why you haven't use it?
> 
> I felt that using `useTransition` was out of scope for this PR. I wanted to focus on updating having React 18 work, without using any of the new features. I made an exception for useId, since it was trivial to add, and I had to do something about `@reach/auto-id` not supporting React 18 without the use of `--legacy-peer-deps`.

Ok sure, thought, please create a backlog issue to keep a log about adding `useTransition` as I do fear no one will tackle the issue otherwise.

storybook/decorators/withBackgroundMatchingColorScheme.jsx
tjosepo: The documentation mentions both syntaxes, but don't seem to tell when to use one over another 🤔

patricklafrance: Oh you are right, I haven't notice the tab before you mentionned it.

Well it says the following:

// Replacing the <Story/> element with a Story function is also a good way of writing decorators.
// Useful to prevent the full remount of the component's story.

Preventing a full remount seems like a good idea then.

I wonder thought why it cause rendering issue when using the component syntax. Is there an underlying issue we will hide by patching with the function syntax?

packages/components/src/tag/src/TagList.tsx
tjosepo: I think I figured it out.

I changed the type of  `item.content` to be `ReactNode`, since it was only used to store `children` props.

alexasselin008:
> @alexasselin008 wasn't there issues you found out earlier while investigating React 18 `useId` function?

I forgot. The issue i had was that the useId function can not be used for css selectors 

As per the documentation: 
`useId generates a string that includes the : token. This helps ensure that the token is unique, but is not supported in CSS selectors or APIs like querySelectorAll.`

tjosepo:
> The issue i had was that the useId function can not be used for css selectors

I don't think we ever apply styling to class names with auto-generated IDs in them.

To be honest, I'm not sure what the point of these class names even are 🤔 `useId` mostly seems useful when trying to associate a field to a label.

patricklafrance:
@tjosepo I would have expect those cases with the fragments to be fixed before you merge the PR.

> const { avatar, "end-icon": endIcon, icon, stringValue, text } = useRawSlots(<>{selectedItem?.content}</>, ["icon", "avatar", "text", "end-icon"]);

Could you submit another PR asap to fix those?

patricklafrance:
> > The issue i had was that the useId function can not be used for css selectors
> 
> I don't think we ever apply styling to class names with auto-generated IDs in them.
> 
> To be honest, I'm not sure what the point of these class names even are 🤔 `useId` mostly seems useful when trying to associate a field to a label.

For classes I am not sure, the common use case is to generate UUID to generate the required attributes to fulfill accessibility requirements

alexasselin008:
Ill take a look into useRawSlots

PR #1048, opened 2022-10-14
Fix header anchor links missing

tjosepo:
In Storybook, headers no longer displayed anchor links when you hovered on them.

image

This issue was fixed in  Storybook `6.5.13-alpha.0`.


fraincs (commented):
Can we live without anchors until 6.5.13 is non alpha? Is the alpha stable, by definition it shouldn't?

fraincs:
They are not rendered at all

patricklafrance:
> They are not rendered at all
> […](#)

It would broke the whole docs navigation. I doubt it’s an option to release without anchors.

fraincs:
> > They are not rendered at all
> > […](#)
> 
> It would broke the whole docs navigation. I doubt it’s an option to release without anchors.

I agree I was just wondering if we were ok with an alpha build of SB.

patricklafrance:
> > > They are not rendered at all
> > > […](#)
> > 
> > 
> > It would broke the whole docs navigation. I doubt it’s an option to release without anchors.
> 
> I agree I was just wondering if we were ok with an alpha build of SB.

An alpha of a minor release is not a big deal IMO but still a due diligence should be done to validate what is included in this minor release.

fraincs (approved):

PR #1052, opened 2022-11-04
Fix dark mode toggle

tjosepo:
## Summary

An update to Storybook broke our dark mode toggle.

Updating the value longer updated the `globals` variable. The page had to be refreshed for the changes to be reflected.

## What I did

I created an addon to handle dark mode.

The addon stores dark mode value to localstore, and emits an event when the colorscheme changes. This ensures that previews and stories are always up to date with the toggle.


fraincs (approved):

patricklafrance:
@tommysanterre @fraincs I don't think merging this is right.

I mean, `globals` is an [official feature of SB](https://storybook.js.org/docs/react/essentials/toolbars-and-globals) maybe the latest version included a bug but they will release a fix.

I doubt it worth going rogue with our things. It's additional development and additional maintenance.

I believe this PR should be reverted

tjosepo:
I don't mind reverting if we're okay with the feature temporarily not working.

I'll check the Storybook repo to see if this issue was flagged.

From my time spent debugging, it seems like the feature still works as intended inside previews, but not inside docs. (Changing 
the value causes the the docs rerender, but the `global` variable stays the same as before 😕)

tjosepo:
Seems like the issue was flagged in https://github.com/storybookjs/storybook/issues/18477

> Ah! Ok, that's because you are reading `context.globals` which is a deprecated field and will be removed in 7.0.
> 
> I guess the correct way to access them in a container would be:
> ```js
>   const story = context.storyById(context.id);
>   const { globals } = context.getStoryContext(story);
> ```

> (Note in 7.0 context.id will be removed and you'd change the first line to `context.storyById()` -- but that doesn't work in 6.5 🤔 )

We can revert this PR, and do the changes suggested in the issue.

It's kind of a bummer that the official solution to this issue is only temporary, and will become deprecated again in version 7.0 😕

patricklafrance:
Hmm yeah indeed! Let’s discuss our next step on monday then?

patricklafrance:
So I read the mentionned SB issue.

If I understand correctly, the issue for Orbit is not that the globals value doesn't work anymore, the issue is that the template (specifically, in our case the `withBackgroundMatchingColorScheme` decorator) doesn't re-render anymore when a global value mutate?

If this is the case and the following works:

```
 const story = context.storyById(context.id);
 const { globals } = context.getStoryContext(story);
```

I don't think it's a big deal to update:

```
const story = context.storyById(context.id);
```

for:

```
const story = context.storyById(context.storyById());
```

when SB 7.0 is out

tjosepo:
Great! 

We can merge https://github.com/gsoft-inc/sg-orbit/pull/1053, and then I'll make another PR with these changes: https://github.com/gsoft-inc/sg-orbit/commit/b46e6e3b380054bc898a34cde08178e378f57bb0

PR #1055, opened 2022-11-09
Add missing icon button link docs example

patricklafrance:
## Summary

A doc examples for IconButtonAsLink was missing.


patricklafrance:
After discussion with @fraincs I will ignore the failing chromatic test as it's a flacky which often fail.

PR #1086, opened 2022-12-20
Updated Stylelint version

alexasselin008:
## Summary

There are multiple inconsistencies in the css files. This PR is a proposition of stylistic rules we could add to the sg-stylelint configs to standardise the writing of CSS

## What I did

- Update stylelint rules
- Fix files with broken rules


packages/components/src/button/src/Button.css
```diff
@@ -351,12 +351,12 @@ a.o-ui-button {
 .o-ui-button-inherit-style.o-ui-button-primary .o-ui-button-end-icon {
     fill: var(--o-ui-white);
     color: var(--o-ui-white);
-    opacity: .9;
+    opacity: 0.9;
```
alexasselin008, line 354: @fraincs Any opinion on this? at the moment, we have both 0.X and .X. Should we settle on one. Do you prefer with or without the 0?

packages/components/src/card/src/Card.css
```diff
@@ -118,7 +118,7 @@
     padding: var(--o-ui-sp-6);
 }
 
-.o-ui-card-horizontal > .o-ui-card-illustration+.o-ui-card-aside {
+.o-ui-card-horizontal > .o-ui-card-illustration + .o-ui-card-aside {
```
alexasselin008, line 121: @fraincs the spacing around these operator is inconsistent. Is this the way you would like to write them, or without the spaces?

.stylelintrc.js
```diff
@@ -1,6 +1,43 @@
+/** @type {import('stylelint').Config} */
 module.exports = {
     extends: "@sharegate/stylelint-config-recommended",
     rules: {
         "linebreaks": "unix",
+
+        "color-hex-length": "long",
```
alexasselin008, line 7: Will be in generic config file when accepted

packages/components/src/button/src/Button.css
fraincs: No strong opinion, although in the case of opacity it has no use, as the maximum is 1.0

packages/components/src/card/src/Card.css
fraincs: This is what happens on save on my end, been fighting with them a lot ` + ` is perfect.

fraincs (changes requested):
I would remove the trailing 0 rule as its not adding anything.

fraincs (approved):

PR #1129, opened 2023-01-27
Convert CSF Stories to CSF 3 + TypeScript

alexasselin008:
## Summary

chroma.jsx that were in the CSF were: 
- renamed to stories.tsx
- converted to TypeScript
- converted to CSF3*

Even if the stories are in the CSF3 format, that all specify a render function, which could be optimised by:
- if the render function only forward props to compnents, it could be remove and replaced with args
- if the render function is similar to another test, spread the other test in the story instead of re-writing it entirely

Mutiple test have been modified because they were using property that aren't available anymore.


alexasselin008:
States story where "focus" "selected" and "hover" were added on items are valid. It's their typescript types that should of been modified.

alexasselin008:
40 stories were added in Grid, since the viewport chromatic setting were never set up properly

fraincs (approved):

Issue #1141, opened 2023-02-07
🐛 Tooltips often stay open when hovered fast


alexasselin008:
## Describe the bug

![tooltip war](https://user-images.githubusercontent.com/38871812/217299022-01b251ab-25c2-4929-ab48-c87521373a86.gif)

## Steps to reproduce

I've had luck reproducing this issue when hovering fast over a stack of icons

```
<Stack gap={1} alignY="center" height="200px" overflowY="scroll" paddingTop="50px">
    <TooltipTrigger flexShrink={0}>
        <EmailIcon  flexShrink={0} />
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0} />
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0}/>
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0}/>
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0}/>
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon flexShrink={0} />
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0}/>
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0}/>
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
        <TooltipTrigger>
        <EmailIcon  flexShrink={0}/>
        <Tooltip>Send an email to the orbital space station.</Tooltip>
    </TooltipTrigger>
    
</Stack>
```

alexasselin008:
Had to revert my initial fix since it broke that possiblity to mouseover a tooltip. During the investigation, i found out what the real problem was. 

The dom structure is like this: 
```
<Overlay>
     <Tooltip />
     <OverlayArrow />
</Overlay>
```
[...]

PR #1142, opened 2023-02-08
#1141 Tooltips often stay open when hovered fast

alexasselin008:
## Summary

Tooltips now open one at the time, never multiple at once. Also fixed an issue where mouseLeave event would not be detected anymore when hovering the overlay

## What I did

The condition 
```
if (!isTargetParent((event as FocusEvent).relatedTarget, overlayRef)) {
```
really was the issue. It was intended for blur events, but was preventing mouseLeave event from firing if the mouseleave was on the overlay. 

added a check to make sure only blur event were taking into consideration for this logic

### DIFF
--- a/packages/components/src/tooltip/src/TooltipTrigger.tsx
+++ b/packages/components/src/tooltip/src/TooltipTrigger.tsx
@@ -7,7 +7,6 @@ import {
     isNil,
     mergeProps,
     resolveChildren,
-    useControllableState,
     useEventCallback,
     useId,
     useMergedRefs
@@ -17,6 +16,7 @@ import { useResponsiveValue, useThemeContext } from "../../styling";
 
 import { Div } from "../../html";
 import { TooltipTriggerContext } from "./TooltipTriggerContext";
+import { useTooltipTriggerState } from "./useTooltipTriggerState";
 
 const DefaultElement = "div";
 
@@ -95,7 +95,7 @@ export function InnerTooltipTrigger({
 
     const { themeAccessor } = useThemeContext();
 
-    const [isOpen, setIsOpen] = useControllableState(open, defaultOpen, false);
+    const { isOpen, setIsOpen } = useTooltipTriggerState(open, defaultOpen, false);
 
     const updateIsOpen = useCallback((event: SyntheticEvent, newValue: boolean) => {
         if (isOpen !== newValue) {
@@ -126,7 +126,7 @@ export function InnerTooltipTrigger({
         isDisabled: disabled,
         onHide: useEventCallback((event: SyntheticEvent) => {
             // Prevent from closing when the focus goes to an element of the overlay on opening.
-            if (!isTargetParent((event as FocusEvent).relatedTarget, overlayRef)) {
+            if (event.type !== "blur" || !isTargetParent((event as FocusEvent).relatedTarget, overlayRef)) {
                 updateIsOpen(event, false);
             }
         }),
--- a/packages/components/src/tooltip/src/useTooltipTriggerState.ts
+++ b/packages/components/src/tooltip/src/useTooltipTriggerState.ts
@@ -1,0 +1,55 @@
+import { useEffect, useMemo } from "react";
+import { useControllableState } from "../../shared";
+
+export interface OverlayTriggerState {
+    readonly isOpen: boolean;
+    setIsOpen: (isOpen: boolean) => void;
+}
+
+const tooltips:Record<string, () => void> = {};
+let tooltipId = 0;
+
+
+/*
+ * Manages state for a tooltip trigger. Tracks whether the tooltip is open, and provides
+ * methods to toggle this state. Ensures only one tooltip is open at a time
+*/
+export function useTooltipTriggerState(defaultOpen: boolean, open: boolean, defaultValue = false): OverlayTriggerState {
+    const [isOpen, setIsOpen] = useControllableState(open, defaultOpen, defaultValue);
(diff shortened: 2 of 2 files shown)


fraincs (approved):

patricklafrance:
Sorry for commenting a closed PR.

I am glad that you are working on this issue and seems to have found a fix.

However, I am not sure this is the right one.

For a tooltip, there's a use case where it could be opened by focusing a trigger element with a tab keypress. When it happens, since the tooltip is not activated with the mouse, the mouse can hover something else. If the mouse hover another tooltip trigger element, with this fix, it would actually closed the currently tab focused trigger element. I would personally consider this as a bug.

Here's an example:

tooltip

While this use case is not frequent, I believe it highlights that it might not be the right fix for this issue.

For a design system component, I usually wouldn't recommend tracking unknown component states. I think components should only know about theirself or sometimes, when in a parent/child relationship, integrates with an opt-in context mecanism.

I was personally expecting the fix to rather involve reacting to specific DOM events. Doing it this way rather seems like a patch because the cause of the issue is not well understood.

Anyway, I wouldn't do it this way but if you still want to, here's a few other things I noticed:

- You should use a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) rather than an object literal to store the close functions. Kind of weird, but I think that with a `WeakMap`, the key could even be the `setIsOpen` instead of the id.

- A tooltip already have a unique id, you could potentially use this instead of a generating a new one. (`const tooltipId = useId(tooltip.props.id, "o-ui-tooltip");`

- `closeOpenTooltips` -> `closeOpenedTooltips`

- `ensureTooltipEntry `, the word "ensure", usually means to validate that something already exist or is in a specific state. In this case, you are adding a new entry to the tooltip bucket. I would rather go with something like `add` or `set`

- It might something that you wish to remove from the DS, but at the moment those `if` conditions are usually using the `isNil` function.

```
if (!isNil(tooltip))
```

alexasselin008:
Hi @patricklafrance , 

Thanks for your insights on this PR, all comments are welcome, even on closed PR. 

This PR contains 2 things: 
- The "right" fix on where we adress the issue when reacting to a specific DOM event. 

```diff
-if (!isTargetParent((event as FocusEvent).relatedTarget, overlayRef)) {
+if (event.type !== "blur" || !isTargetParent((event as FocusEvent).relatedTarget, overlayRef)) {
```

- The rest of the code is more "future proofing" errors that could occur with tooltips. ~Most of~ All the the time, we only want one tooltip opened at a time. The gif in the issue should multiple tooltip being opened at once (my fix should fix the issue) but should an other bug happen, at least it will prevent multiple tooltip to be shown. 

> For a design system component, I usually wouldn't recommend tracking other components state. I think components should only know about theirself or sometimes, when in a parent/child relationship, integrates with an opt-in context mecanism.

This code was directly taken from `@react-stately/tooltip`. The only reason i didnt take the package directly is because they have a notion of "warm up" and "delay" before showing tooltips, and we don't necessary want that concept.

So React-Spectrum do think it's recommended to track other component state. 

For the other concerns: 
- Using a weakmap here doesn't bring any value, since there is a useEffect that makes sure that the tooltip is removed from the map, and when a tooltip open, it deletes all other entries. There is no possibility of this object litteral to be the only one having a reference on the tooltip
- There is no need for the tooltipID to be passed, since we don't use this map to retreive specific tooltips. 
- I could do the renaming, but since they are named this way in stately, i would prefer to keep their original name, since they are scoped inside a method and not public. 
- Good catch for the isNil, i'll fix it in another PR

patricklafrance:
Thanks for the clarifications!

Glag to know that you found the underlying issue, I misread the change and read it as you were "removing the "blur" check rather than adding it.

I would recommend to update the comment on the previous line accordingly to remember why the blur check has been added. That's the kind of fixes that most people tend to forget after a few days.

I would also recommend adding interaction tests to ensure the issue isn't reintroduced accidently later on.

> The rest of the code is more "future proofing" errors that could occur with tooltips. Most of All the the time, we only want one tooltip opened at a time. The gif in the issue should multiple tooltip being opened at once (my fix should fix the issue) but should an other bug happen, at least it will prevent multiple tooltip to be shown.

I dont understand what you mean here. Are you inferring that the use case in the gif isn't valid?

Furthermore, if adding the "blur" check fixed the issue, why adding code that you don't need? Adding code only for "future proofing" IMO is more like adding complexity and futur code to maintain.

> So React-Spectrum do think it's recommended to track other component state.

Well, as far as I know you are not aware of the context and the reasons behind the addition of this code to react-spectrum. So you can't say for sure that they think it's "recommended". I personally still think it's a bad idea to track other components state directly in DS land.

> Using a weakmap here doesn't bring any value, since there is a useEffect that makes sure that the tooltip is removed from the map, and when a tooltip open, it deletes all other entries. There is no possibility of this object litteral to be the only one having a reference on the tooltip

Granted, I was preaching for the safe side on this one but you're right it's useless since there's a `useEffect`. I wonder how it works thought with SSR? Since `userEffect` is a no-op on the server, I guess the dispose function won't be called? Is it? Not sure how SSR manage memory.

> There is no need for the tooltipID to be passed, since we don't use this map to retreive specific tooltips.

The idea here is that you already have a unique id for a tooltip, why use a new one.

> I could do the renaming, but since they are named this way in stately, i would prefer to keep their original name, since they are scoped inside a method and not public.

I am not sure why it matters that stately is already using this name. Anyway, up to you!

Issue #1018, opened 2022-08-05
🐛 The Button loading state does not prevent onClick


odrouin:
## Describe the bug

When adding the disabled state on a button it prevents the onClick from being called when clicking it but when adding the loading state, the onClick is called anyway even if both states have pretty much the same style with a big red circle mouse cursor when hovering it.

## Steps to reproduce

1. `<Button disabled onClick={() => alert("test")}>Click</Button>` does nothing when clicking it.
2. `<Button loading onClick={() => alert("test")}>Click</Button>` triggers the alert when clicking it.

## Expected results

The button should not call the onClick when it is in a loading state.

alexasselin008:
I agree that this is a bug, since it is really expected by the user, since we grey out the control AND add the specific "not-allowed" cursor.
![image](https://user-images.githubusercontent.com/38871812/183108722-ce5be21e-11f6-45ae-8793-4dd2ffb7d100.png)

patricklafrance:
Shopify Polaris is using internally a [use-disable-interaction](https://github.com/Shopify/polaris/blob/main/polaris-react/src/utilities/use-disable-interaction.ts) hook to provide this behavior to components.

We could do something similar.

PR #1145, opened 2023-02-09
Loading button should not trigger onClick events

alexasselin008:
## Summary

PR for https://github.com/gsoft-inc/sg-orbit/issues/1018


package.json
```diff
@@ -20,8 +20,8 @@
     "build:sb": "build-storybook -o storybook-static/sb",
     "build:sb-chroma": "cross-env STORYBOOK_IS_CHROMATIC=true build-storybook -o storybook-static/chromatic",
     "build:sb-docs": "cross-env STORYBOOK_IS_DOCS_SITE=true build-storybook -o storybook-static/docs --docs",
-    "build:netlify-sb": "yarn build && yarn build:sb",
-    "build:netlify-docs": "yarn build && yarn build:sb-docs",
+    "build:netlify-sb": "yarn && yarn build && yarn build:sb",
+    "build:netlify-docs": "yarn && yarn build && yarn build:sb-docs",
```
fraincs, line 24: pkoi le changement?

fraincs (approved):

packages/components/src/shared/src/useDisableInteraction.tsx
```diff
@@ -0,0 +1,23 @@
+/**
+ * useDisableInteraction provides the original event handler but disables interaction
+ * if the boolean passed is true.
+ */
+import { useCallback, MouseEventHandler } from "react";
+
+export function useDisableClick<T = Element>(disabled?: boolean, handleClick?: MouseEventHandler<T>) {
```
patricklafrance, line 7: Are those parameters really optional? I would expect the caller to always provide those

patricklafrance, line 7: Since there is only a single function in the file, I would recommend matching the name of file and the name of the function.

patricklafrance, line 7: Are you planning on using this hook for other components?

If not, I recommend moving the file in the `button` folder until another component needs it.

packages/components/src/button/src/useDisableClick.tsx
```diff
@@ -0,0 +1,23 @@
+/**
+ * useDisableInteraction provides the original event handler but disables interaction
+ * if the boolean passed is true.
+ */
+import { useCallback, MouseEventHandler } from "react";
+
+export function useDisableClick<T = Element>(disabled: boolean, handleClick: MouseEventHandler<T>) {
+    const handleClickWrapper = useCallback<MouseEventHandler<T>>(
+        event => {
+            if (disabled) {
```
tjosepo, line 10: Won't this always be `true`?

The function `handleClickWrapper` is only referenceable if `disabled` is true. If the value of `disabled` changes, there will be a stale reference to when `disabled` was true. If `disabled` changes, it won't affect the function until the next rerender, at which point the original `handleClick` will be returned and the new `handleClickWrapper` will be un-referenceable.

Should a `ref` be used instead?

Or just remove the `if` statement and inline the content inside of it?

package.json
alexasselin008: je l'avais enlever avant, pis la les previews ne marchait plus sur la branche. Donc revert + je trouve c'Est quoi le issue quand je l'enleve a un autre moment

packages/components/src/button/src/useDisableClick.tsx
alexasselin008: The if looks redundant you're right. I'll remove it in another PR

PR #1152, opened 2023-02-20
[Draft]

alexasselin008:
## Summary

Fix for issue https://github.com/gsoft-inc/sg-orbit/issues/1141

## Description of the issue

Here is 3 scenario of how the tooltip works before the fix

### Scenario 1 (works): 
- MouseEnter on trigger -> Tooltip shows
- MouseLeave on Trigger  -> MouseEnter on white background -> Tooltip disappear because of the event handlers on the trigger. 
image

### Scenario 2 (works):
- MouseEnter on trigger -> Tooltip shows
- MouseLeave on trigger ->  MouseEnter Overlay -> Trigger's Event handler ignore the mouseLeave because it's on the overlay. and that is allowed 
- MouseEnter on Tooltip -> Tooltip stays open
- MouseLeave Tooltip -> Tooltip closes because of the event handlers on the tooltip
image

### Scenario 3 (the issue):
- MouseEnter on trigger -> Tooltip shows
- MouseLeave on Trigger -> MouseEnter on Overlay -> Trigger's Event handler ignore the mouseLeave because it's on the overlay. and that is allowed 
- MouseLeave Overlay -> MouseEnter on white background -> since the mouse never went in the tooltip, the popup stays open because nobody listen to overlay mouseLeave. The bug is here
image

## What I did

Instead of adding spacing in the overlay and that the overlay adds an arrow, I've move spacing/arrow logic in the tooltip and the popover. This allows better control over how to position and handle the arrow

### DIFF
--- a/packages/components/src/tooltip/src/Tooltip.css
+++ b/packages/components/src/tooltip/src/Tooltip.css
@@ -1,3 +1,16 @@
+.o-ui-tooltip-wrapper {
+    max-width: var(--o-ui-sz-16);
+
+    --o-ui-tooltip-arrow-border-color: var(--o-ui-focus-ring-color-alias-default);
+    --o-ui-tooltip-arrow-dimension: 10px;
+    /* since we take a 10px by 10px and add a transformation of 45, we get a lozenge. the height of that lozenge can be calculated using
+       sqrt(var(--o-ui-tooltip-arrow-dimension) ^ 2 + var(--o-ui-tooltip-arrow-dimension) ^ 2) = 14.142135623730951. Since calc doesn't
+       allow complex calculation like this, we pre-compute it in --o-ui-tooltip-lozenge-dimension
+    */
+    --o-ui-tooltip-lozenge-dimension: 14px;
+    --o-ui-tooltip-spacing-with-trigger: var(--o-ui-sp-3);
+}
+
 .o-ui-tooltip {
     background-color: var(--o-ui-bg-alias-surface);
     border-radius: var(--o-ui-br-3);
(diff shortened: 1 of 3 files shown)


packages/components/src/tooltip/src/TooltipTrigger.tsx
```diff
@@ -175,18 +174,13 @@ export function InnerTooltipTrigger({
                     rest,
                     {
                         as,
-                        borderOffset: themeAccessor.getSpace(3),
```
alexasselin008, line 178: we got rid of this "workaround" so thats  a really good thing.

packages/components/src/tooltip/src/TooltipTrigger.tsx
```diff
@@ -175,18 +174,13 @@ export function InnerTooltipTrigger({
...
                         ref: overlayRef,
                         show: isOpen,
                         zIndex
                     }
                 )}
             >
                 {tooltipMarkup}
-                <OverlayArrow
```
alexasselin008, line 186: i feel like the overlay arrow is better located in the tooltip than in the trigger, so i feel like this is a good thing too

packages/components/src/tooltip/src/Tooltip.css
```diff
@@ -8,18 +21,104 @@
...
+    bottom: calc(var(--o-ui-tooltip-lozenge-dimension) / 2);
+}
+
+.o-ui-overlay[data-popper-placement^="bottom"] .o-ui-tooltip-arrow {
+    top: calc(var(--o-ui-tooltip-lozenge-dimension) / 2);
+}
+
+.o-ui-overlay[data-popper-placement^="left"] .o-ui-tooltip-arrow {
+    right: calc(var(--o-ui-tooltip-lozenge-dimension) / 2);
+}
```
alexasselin008, line 93: those styles used to be located in the Overlay.css. I'm not SUPER happy that i reference o-ui-overlay here, but i don't see any other way. 

We can't trust the "position"/"placement" property set on tooltip, since even if we set "top", if there is no room to display the tooltip at the top, popperjs will show it under, and the data-popper-placement will be updated to bottom, but the property stays at "top"

packages/components/src/tooltip/src/Tooltip.tsx
```diff
@@ -41,20 +43,27 @@ export function InnerTooltip({
     });
 
     return (
-        <Text
+        <Div
             {...mergeProps(
                 rest,
                 {
                     as,
-                    className: "o-ui-tooltip",
```
alexasselin008, line 49: Changing the structure of the tooltip break 6 chromatic test. Before we were able to modify the border color of the tooltip for instance. It was broken because the style was not propagated to the arrow. 

Now the style is forwarded to the wrapper, so the border doesn't target the tooltip's border. 

I think that it's ok and that i don't need to change anything

packages/components/src/tooltip/src/TooltipTrigger.tsx
```diff
@@ -159,6 +157,9 @@ export function InnerTooltipTrigger({
     ));
 
     const tooltipMarkup = augmentElement(tooltip, {
+        arrowProps: {
```
alexasselin008, line 160: could of simply passe arrowRef, but popover passes "arrowProps" so i felt like it would be better if they both had the same name

packages/components/src/popover/src/Popover.tsx
```diff
@@ -194,10 +198,26 @@ export function InnerPopover({
...
+                    dismissable && "dismissable"
+                )}
+                >
+                    {dismissButtonMarkup}
+                    {headerSectionMarkup}
+                    {content}
+                    {footerSectionMarkup}
+                </Div>
+
+                {arrowProps && <Div
```
alexasselin008, line 212: i don't really like that, introduce new prop HasArrow in addition to arrowProps?

packages/components/src/tooltip/src/Tooltip.tsx
alexasselin008: should forward rest to the tooltip not the wrapper

packages/components/src/popover/src/Popover.tsx
alexasselin008: fixed

alexasselin008:
We won't go this way for now, so i'm closing the PR

PR #1162, opened 2023-02-27
fix: Remove fill color from icons

tjosepo:
## Summary

The `fill` value was hardcoded in the SVG, which made it impossible to overwrite the color, and made the icon not adapt to dark mode.


alexasselin008:
Thank you for the PR!. 

The generated-icons folder is full of component auto generated when generating the react component from the SVgs. Modifiing this file means next build its going to be overwritten. 

Also, looking at the svgs, multiple icons have a fill icon so they probably need to be fixed as well. @fraincs can you take a look a this? 
image

alexasselin008:
You can check/test all icons here: 

https://sg-storybook.netlify.app/?path=/docs/chromatic-icons--icon-folder-major by toggling the dark mode

fraincs:
@tjosepo can you fix those while at it?

tjosepo:
I went ahead and removed the `fill="none"` attribute on the icons. Seems like they were unecessary, and it made previewing the icons harder.

PR #1166, opened 2023-02-28
Remove Babel and favor TSC for compiling

alexasselin008:
## Summary

- Removed babel from being used for projects transpilation. 
- Removed babel from  the storybook config
- Removed `svgr/webpack`. The few svgs that were previously dynamically loaded are not pre-compiled as react components
- The script `create-dir` is no longer necessary since tsc creates the folder by default
- Replaced babel-jest with ts-jest
- testPathIgnorePatterns is no longer necessary since the TSC only builds what is being published, so the tests are no longer in the build folder
- Jest config now uses pathsToModuleNameMapper in order to not duplicate the aliases defined in the ts config


.storybook/webpack.config.js
```diff
@@ -11,6 +11,7 @@ function addWebpackAliases(config) {
         "@css": path.resolve(__dirname, "../packages/css/src"),
         "@components": path.resolve(__dirname, "../packages/components/src"),
         "@orbit-ui/components": path.resolve(__dirname, "../packages/components/src"),
+        "@sharegate/orbit-ui": path.resolve(__dirname, "../packages/bundle/src"),
         "@experimental": path.resolve(__dirname, "../packages/experimental/src"),
         "@stories/components": path.resolve(__dirname, "./components"),
         "@stories/mdx": path.resolve(__dirname, "./mdx"),
```
patricklafrance, line 17: Do you actually still need those aliases?

Maybe those related to CSS files but those related to JS files (ts, tsx, ...) should be managed by TSC, not Webpack

patricklafrance: Ah it's for SB, I guess it's out of scope of TSC :(

alexasselin008: Yeah, its related to the new-ish Experimental package. I noticed that storybook was failing if we started it from a "fresh" install. It worked if the bundle was built once.

patricklafrance (approved):

PR #1167, opened 2023-03-01
Feature/icon design updates

fraincs:
Issue: 

## Summary

Some spacing issues had to be solved post icon rework.

## Accordion
- removed margin left on icon
- Spacing between icon and title is now 8px

## Listbox

- Listbox checkmark position is now 10px from the right
- Icon has no margin left anymore.
- endIcon has 10px margin on the left

## Message

- Padding left is now 40px - icon is positioned inside the padding at 12px

## Radio

- icon has now a 4px margin from the text

## Checkbox

- icon has now a 4px margin from the text

## Switch

- icon has now a 4px margin from the text

### DIFF
--- a/packages/components/src/listbox/src/Listbox.css
+++ b/packages/components/src/listbox/src/Listbox.css
@@ -71,7 +71,7 @@ IMPORTANT: The Listbox component hardcoded a few CSS values to enable dynamic sc
     mask-size: contain;
     background-color: var(--o-ui-bg-alias-accent);
     position: absolute;
-    right: 0.875rem;
+    right: calc((var(--o-ui-sp-8) - var(--o-ui-listbox-option-checkmark-size)) / 2);
 }
 
 /* OPTION | STATE | DISABLED */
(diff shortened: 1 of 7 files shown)


packages/components/src/listbox/src/Listbox.css
```diff
@@ -144,7 +143,7 @@ IMPORTANT: The Listbox component hardcoded a few CSS values to enable dynamic sc
 
 /* OPTION | ICONS | END ICON */
 .o-ui-listbox-option-end-icon {
-    margin-left: calc(var(--o-ui-sp-1) + 0.125rem);
+    margin-left: calc(var(--o-ui-sp-2) + 0.125rem);
```
patricklafrance, line 146: Values not on scale values are a smell ;)

fraincs: The 0.125rem could be removed here actually and would not make a huge difference, I considered this one not a smell since its not on the vertical rhythm scale. It was to compensate the fact that icons are now 20x20 missing 2px each side, that spacing was included in the icon before.

alexasselin008 (approved):

PR #1168, opened 2023-03-01
Feature/size icon

fraincs:
Issue: 

## Summary

Size Icons support only one size it was counter intuitive to support size inherit only for the TextLink component. This PR is about removing size from Icon once and for all and use another strategy to support the icon in a small textLink component.

## What I did

Removed size prop from Icon component. Reworked the textLink component to pass it's size to the link anchor as well as add come CSS to make sure inherit worked on the icon size in that situation.


packages/components/src/link/src/Link.css
```diff
@@ -157,19 +165,25 @@
 }
 
 /* ICON */
-.o-ui-link-start-icon,
-.o-ui-link-end-icon {
+.o-ui-text-link .o-ui-text-link-start-icon,
+.o-ui-text-link .o-ui-text-link-end-icon {
+    width: 1em;
```
alexasselin008, line 170: shouldn't we use --o-ui-sz-2 here?

fraincs: No since rem's are based on the root font size while em's are based on the parent font size. That would make all icons the same width.

alexasselin008: my bad did not notice

alexasselin008 (approved):

PR #1171, opened 2023-03-06
DeprecationWarning: Default PostCSS plugins are deprecated.

alexasselin008:
## Summary

image

## What I did

Added a postcss config with the sharegate shared postcss config.


alexasselin008:
tests are broken, investigating

Issue #1024, opened 2022-09-07
🐛 Tilde "~" imports in CSS not valid in all CSS pre-processors


tjosepo:
## Describe the bug

The tilde "~" import used in Orbit to import styles from the `node_modules` folder is not a standard CSS feature and most CSS pre-processors do not support it.

Here's a list of how the common pre-processors handle imports in CSS:

| Preprocessor | `@import "~package";` | `@import "package";` | `@import "./relative.css";` | `url("~package")` | `url("package")`| `url("./relative.png");`
| --- | --- | --- | --- | --- | --- | --- |
| css-loader (Webpack default) | ✅ Valid | ✅ Valid | ✅ Valid | ✅ Valid | ⚠️ Requires resolve-url-loader | ⚠️ Requires resolve-url-loader
| sass-loader |  ⚠️ Marked as deprecated | ✅ Valid | ✅ Valid | ⚠️ Marked as deprecated | ⚠️ Requires resolve-url-loader | ⚠️ Requires resolve-url-loader
| postcss-loader |  ❌ Invalid | ✅ Valid | ✅ Valid | ❌ Invalid | ✅ Valid | ✅ Valid
| postcss-import (used by Vite) | ❌ Invalid | ✅ Valid | ✅ Valid | ❌ Invalid | ✅ Valid | ✅ Valid
| esbuild | ❌ Invalid | ✅ Valid | ✅ Valid | ❌ Invalid | ✅ Valid | ✅ Valid
| parcel-css | ❌ Invalid | ❌ Requires `npm:` url scheme |  ✅ Valid | ❌ Invalid | ❌ Requires `npm:` url scheme | ✅ Valid

## Potential fix

The ideal fix would be:

- Make all `@imports` relative.
- Make all `url(...)` imports relative and ask to use `resolve-url-loader`.

alexasselin008:
After looking at this issue once again today, this could be fixed simply. 

There is 2 issues: 

## imports

We could simply change the 2 occurences of 
```diff
-@import "~@orbit-ui/css";
-@import "~@orbit-ui/components/index.css";
+@import "@orbit-ui/css";
+@import "@orbit-ui/components/index.css";
```

## url(~)

This one is trickier. We only use url(~) to position checkmarks on:
- Checkbox
- Listbox
- Menu
- Tiles

We could replace the CSS with React components that are positionned at the same place. This way we would not have to reach into assets in the node_modules.

This is the way chakra and mantine does it.
[...]

PR #1178, opened 2023-03-08
Fix for Tilde "~" imports in CSS not valid in all CSS pre-processors

fraincs:
Issue: 

## Summary

#1024 - Tilde "~" imports in CSS not valid in all CSS pre-processors 

## What I did

Remove ~ as it's unnecessary, reworked components using a CSS background url ~ as it's use in CSS is now widely supported.

### DIFF
--- a/packages/components/src/checkbox/src/Checkbox.css
+++ b/packages/components/src/checkbox/src/Checkbox.css
@@ -44,7 +44,7 @@
     transition: var(--o-ui-focus-ring-transition);
 }
 
-.o-ui-checkbox .o-ui-checkbox-box .o-ui-checkbox-checkmark {
+.o-ui-checkbox-checkmark {
     opacity: 0;
     position: absolute;
     top: -1px;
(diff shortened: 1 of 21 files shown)


packages/components/src/checkbox/src/Checkbox.css
```diff
@@ -44,46 +44,32 @@
     transition: var(--o-ui-focus-ring-transition);
 }
 
-.o-ui-checkbox-box:after {
-    content: "";
+.o-ui-checkbox .o-ui-checkbox-box .o-ui-checkbox-checkmark {
```
patricklafrance, line 47: The selector could probably my shrinked to `.o-ui-checkbox-checkmark`

packages/components/src/listbox/src/Listbox.css
```diff
@@ -63,17 +63,19 @@ IMPORTANT: The Listbox component hardcoded a few CSS values to enable dynamic sc
 }
 
 /* OPTION | STATE | SELECTED */
-.o-ui-listbox-option[aria-selected="true"]::before {
-    content: "";
+.o-ui-listbox-option .o-ui-listbox-option-checkmark {
```
patricklafrance, line 66: The selector could be shrinked to `.o-ui-listbox-option-checkmark`

alexasselin008 (approved):

PR #1184, opened 2023-03-08
Spinners and Loaders components

alexasselin008:
## Summary
- A loader provides reassurance that content is currently being loaded, processing, or that change will occur on the page.
- A spinner indicates that a part of the product is currently loading or performing a task, and the duration of this process is unknown.


packages/components/src/input/src/useInputContent.tsx
```diff
@@ -19,3 +20,7 @@ export function useInputButton(button: ReactElement, isActive: boolean, props: R
         variant: "tertiary"
     });
 }
+
+export function useInputSpinner(loading: boolean) {
+    return loading && <Spinner className="o-ui-input-spinner" role="presentation" size="md" />;
```
patricklafrance, line 25: Do you really need to specify size="md" ? It's the default size

packages/components/src/spinner/src/Spinner.tsx
```diff
@@ -0,0 +1,82 @@
...
+                rest,
+                {
+                    as,
+                    className: cssModule(
+                        "o-ui-spinner",
+                        normalizeSize(sizeValue),
+                        active && "active"
+                    ),
+                    ref: forwardedRef,
+                    role: "status"
```
patricklafrance, line 66: I am puzzled by this one, it seems like the right role might "progressbar" according to React Aria progresse bar indeterminate example: https://react-spectrum.adobe.com/react-spectrum/ProgressCircle.html#indeterminate.

It kind of seems to make sense after reading the following https://www.digitala11y.com/progressbar-role.

If progress role is the right role, there be additional step taken in some of Orbits components to correctly set the `aria-describedby` and `aria-busy` attribute. The Spinner components documentation should also be updated to mention those.

How did you come up with the "status" role?

patricklafrance (changes requested):

packages/components/src/spinner/docs/Spinner.stories.mdx
```diff
@@ -0,0 +1,98 @@
...
+</Preview>
+
+
+### Over Background
+
+You can change a spinner style when over a background by setting a color property on the Spinner.
+
+<Preview>
+    <Story name="over background">
+        <Inline alignY="end" backgroundColor="alias-accent" padding={4}>
```
patricklafrance, line 81: Inline is not the right component here, you probably want Flex as you are not horizontally inlining multiple components

patricklafrance: In fact a regular Div might even be enough for this example

packages/components/src/spinner/src/Spinner.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return (
+        <Box
+            {...mergeProps(
+                rest,
+                {
+                    as,
+                    className: cssModule(
+                        "o-ui-spinner",
+                        normalizeSize(sizeValue),
+                        active && "active"
```
patricklafrance, line 63: In which case would a spinner component not be active if he is rendered?

packages/components/src/loader/src/Loader.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+        <Box
+            {...mergeProps(
+                rest,
+                {
+                    as,
+                    className:cssModule(
+                        "o-ui-loader",
+                        show && "show",
+                        active && "active"
```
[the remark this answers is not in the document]
alexasselin008: show means the loader is visible. 
active means the loader is animated. We use this to disable animation for chromatic. We could use the disableAnimation to reduce the spinner's api

packages/components/src/loader/src/Loader.tsx
```diff
@@ -0,0 +1,73 @@
...
+    ...rest
+}: InnerLoaderProps) {
+    const [show, setShow] = useState<boolean>(isNil(delay) || delay === false);
+
+    useEffect(() => {
+        if (isNil(delay) || delay === false) {
+            return;
+        }
+
+        const showTimer = setTimeout(() => setShow(true), isNumber(delay) ? delay : DEFAULT_LOADER_DELAY);
```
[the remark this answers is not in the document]
alexasselin008: Spinners are going to be widely use for all our network calls. I would expect the application devs to be able to choose if they want a longer or shorter delay, but most of the time I feel like people won't think about how long the delay should be. Providing a default options makes sense to me in that case. 

zendesk: 750ms
seek:  800ms
atlassian: 150-200 ms

I've been using 300ms in Home and Cloud Copy and it feels good. Most loading spinner aren't noticeable.

packages/components/src/loader/tests/jest/Loader.test.tsx
```diff
@@ -0,0 +1,47 @@
+import { Loader } from "@components/loader";
+import { createRef } from "react";
+import { renderWithTheme } from "@test-utils";
+import { waitFor } from "@testing-library/react";
+
+// ***** Refs *****
+
+test("ref is a DOM element", async () => {
```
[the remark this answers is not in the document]
alexasselin008: The active props is already used by chromatic to disable the animation. As for the delay props, i've added two tests, but the only way to test that is to track the presence of a css class. Good enough, even if its not the most reliable way

packages/components/src/loader/src/Loader.tsx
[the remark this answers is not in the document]
patricklafrance: Ok!

Then maybe the prop should be renamed to `delayed` ?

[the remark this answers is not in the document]
alexasselin008: Since its number or boolean, i'd keep delay

packages/components/src/loader/src/Loader.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export function InnerLoader({
+    as = DefaultElement,
+    active = true,
+    delay,
+    forwardedRef,
+    ...rest
+}: InnerLoaderProps) {
+    const [show, setShow] = useState<boolean>(isNil(delay) || delay === false);
```
patricklafrance, line 33: I don't think it should be named "show" as it's not a prop of the components API. It should be something like "isVisible" or even "isShow"

fraincs (approved):

patricklafrance (approved):

Issue #1072, opened 2022-11-28
🐛 `rem` units not resizeable by user agent unless `postcss-pxtorem` is used.


tjosepo:
## Describe the bug

By default, [Orbit sets the body height to 16px.](https://github.com/gsoft-inc/sg-orbit/blob/6159198917c38ccdb09829707d253f9fd3bbcb73/storybook/styles/app.css#L3) which overrides the user's preferred font-size.

Unless a package like `postcss-pxtorem` is used to convert the `16px` into `1rem`, the user is unable to change the font-size.

## Steps to reproduce

Go to https://orbit.sharegate.design/ and resize the font size from your browser's settings.

## Expected results

Font size should be resized to the desired font size.

patricklafrance:
Hey @tjosepo!

What do you mean by allowing the user to override the font-size?

Are you referring to zooming the page or it’s something else?

tjosepo:
Most browsers and operating systems have the option to let users increase the size of fonts without zooming on a page.

On Chrome, this can be changed in _Settings_ > _Accessibility_ > _Change font size_

However, this feature doesn't work if a non-scalable unit (like `px`) is used for font-size on the `body` element.

fraincs:
@patricklafrance resizing your whole browser default font size it a feature not that used, if a user does it it is because he has accessibility issues, it's a matter of how attached are we to everything being the size we want vs if we want to accomodate the users knowing that they are probably aware that some website would not look their "best".

It has to be tested, but since we test zooming in and out to an extent, I am confident that changing the root font-size would not break our components, only make them look bigger or smaller, the same with the zoom in / out.

Indeed supporting this would need to go hand in hand with the apps supporting this too/

patricklafrance:
@fraincs I wonder how that would affect the style system? As right now all the classes expect that the default base font-size is 16px. Would it break anything?

fraincs:
> @fraincs I wonder how that would affect the style system? As right now all the classes expect that the default base font-size is 16px. Would it break anything?

same that we expect the user not to zoom and we are ok since we are using rems, I will validate but I think it'll be ok.

fraincs:
@patricklafrance 
<img width="469" alt="image" src="https://user-images.githubusercontent.com/361632/214459742-a2493ec0-7669-4594-9dce-89d2cbc8d2ce.png">

[...]

PR #1195, opened 2023-03-13
Feature/1072

fraincs:
Issue: 

## Summary

rem units not resizeable by user agent unless postcss-pxtorem is used - #1072

## What I did

Converted px to rems in the base code.

## How to test

This requires changing your browser basic font size for now.


patricklafrance:
While it does make sense to use "rem" in Orbit components CSS codebase, I am not sure about using rem for documentation stories and event Chromatic stories.

fraincs:
> While it does make sense to use "rem" in Orbit components CSS codebase, I am not sure about using rem for documentation stories and event Chromatic stories as most people tend to think in term of pixels.

Great point. For the chromatic stories I was planning for the eventuality of testing rems in Chromatic too. It could be px too. As for documentation the idea is to support the users who change their font size even in documentation.

packages/components/src/accordion/src/Accordion.css
```diff
@@ -53,9 +53,9 @@
 /* TRIGGER | BORDERED */
 .o-ui-accordion-bordered .o-ui-accordion-trigger {
     background-color: var(--o-ui-bg-alias-surface);
-    padding: calc(var(--o-ui-sp-4) - 1px) var(--o-ui-sp-5) var(--o-ui-sp-4) var(--o-ui-sp-6);
+    padding: calc(var(--o-ui-sp-4) - 0.0625rem) var(--o-ui-sp-5) var(--o-ui-sp-4) var(--o-ui-sp-6);
     min-height: var(--o-ui-sz-8);
-    border-top: 1px solid var(--o-ui-b-alias-low-break);
+    border-top: 0.0625rem solid var(--o-ui-b-alias-low-break);
```
alexasselin008, line 58: does it make sense for borders to be in REM? last time you mentionned that border are probably the only one that you would prefer having in px, since you always want the border to be 1px

fraincs: For a border to become 2px you would need a font size of roughly 256px. I decided to go rem's everywhere for consistency sake.

PR #1212, opened 2023-05-02
fix: Stale closure preventing tooltip from closing on mouseleave

### LINKED ISSUE #1141: 🐛 Tooltips often stay open when hovered fast

tjosepo:
Closes #1141

## Summary

Seems like tooltips remaining open was related to performance, as it mostly appeared in performance intensive areas like virtualized tables.

My guess was that the issue was caused by calling the `onmouseenter` and `onmouseleave` callbacks in a single render.

My guess seems to have been confirmed after simulating a slow render by adding these lines to the `TriggerTooltip` 

```js
// We simulate a render that takes a least 500ms to happen.
// This gives us enough time to trigger both onmouseenter and onmouseleave events in a single render.
const start = performance.now();
while(performance.now() - 500 < start) {};
```

## What I did

After looking into the code, I noticed that we had an if-statement that looked at the current value of `isOpen` before updating the state. `isOpen` can become stale if `onmouseenter` and `onmouseleave` happen in a single render. I modified the setter function to accept an _updater function_, which always passes the most up-to-date state as the parameter, which fixes the stale closure issue.

## How to test

No idea if we can make a test for this.

### DIFF
--- a/.changeset/popular-parrots-kiss.md
+++ b/.changeset/popular-parrots-kiss.md
@@ -1,0 +1,5 @@
+---
+"@orbit-ui/components": patch
+---
+
+Fix tooltip staying open when moving the mouse quickly
--- a/packages/components/src/tooltip/src/TooltipTrigger.tsx
+++ b/packages/components/src/tooltip/src/TooltipTrigger.tsx
@@ -101,9 +101,8 @@ export function InnerTooltipTrigger({
     const updateIsOpen = useCallback((event: SyntheticEvent, newValue: boolean) => {
         // Ideally, we would only update the state if we transition from open
         // to closed and vice-versa, but comparing the current value of
-        // `isOpen` with the `newValue` results in a stale closure issues,
-        // which can result in the tooltip being left open after moving the
-        // cursor away.
+        // `isOpen` with the `newValue` results in a stale closure, which can
+        // result in the tooltip being left open after moving the cursor away.
         setIsOpen(newValue);
 
         if (!isNil(onOpenChange)) {
--- a/packages/components/src/tooltip/src/TooltipTrigger.tsx
+++ b/packages/components/src/tooltip/src/TooltipTrigger.tsx
@@ -99,14 +99,17 @@ export function InnerTooltipTrigger({
     const [isOpen, setIsOpen] = useControllableState(open, defaultOpen, false);
 
     const updateIsOpen = useCallback((event: SyntheticEvent, newValue: boolean) => {
-        if (isOpen !== newValue) {
-            setIsOpen(newValue);
+        // Ideally, we would only update the state if we transition from open
+        // to closed and vice-versa, but comparing the current value of
+        // `isOpen` with the `newValue` results in a stale closure issues,
+        // which can result in the tooltip being left open after moving the
+        // cursor away.
+        setIsOpen(newValue);
 
-            if (!isNil(onOpenChange)) {
-                onOpenChange(event, newValue);
-            }
+        if (!isNil(onOpenChange)) {
+            onOpenChange(event, newValue);
         }
-    }, [onOpenChange, isOpen, setIsOpen]);
+    }, [onOpenChange, setIsOpen]);
 
     const close = useCallback((event: SyntheticEvent) => {
         updateIsOpen(event, false);
--- a/packages/components/src/tooltip/tests/jest/TooltipTrigger.test.tsx
+++ b/packages/components/src/tooltip/tests/jest/TooltipTrigger.test.tsx
@@ -92,6 +92,24 @@ test("when hovering the tooltip, do not close if hovering the trigger", async ()
     expect(screen.getByRole("tooltip")).toBeInTheDocument();
 });
 
+test("when unhovering the tooltip, close tooltip", async () => {
+    renderWithTheme(
+        <TooltipTrigger>
+            <Button data-testid="trigger">Trigger</Button>
+            <Tooltip>Content</Tooltip>
+        </TooltipTrigger>
+    );
+
+    await userEvent.hover(screen.getByTestId("trigger"), { });
+
+    expect(screen.getByRole("tooltip")).toBeInTheDocument();
+
+    await userEvent.unhover(screen.getByRole("tooltip"));
+
+    expect(screen.queryByRole("tooltip")).not.toBeInTheDocument();
+});
+
+
 // ***** Aria *****
 
 test("when an id is provided for the tooltip, it is used as the tooltip id", async () => {


patricklafrance:
Does it still work is `TooltipTrigger` is controlled by externally managing the `isOpen` value?

tjosepo:
Good catch. I think you might be right. I'll add a test to verify it.

patricklafrance:
An old school trick to fix those kind of issue is to add a `setTimeout()` of 0. It usually hide a deeper issue thought but it might be fine in this case.

Using `setTimeout`, you try something like the following:

```
    const updateIsOpen = useCallback((event: SyntheticEvent, newValue: boolean) => {
        setTimeout(() => {
                if (isOpen !== newValue) {
                    isOpenRef.current = newValue;
                    setIsOpen(newValue);

                    if (!isNil(onOpenChange)) {
                        onOpenChange(event, newValue);
                   }
                }
        }, 0);
    }, [onOpenChange, isOpen, setIsOpen]);
```

I think the `setTimeout` would opt out everything inside from the current React event "transaction" thought. Might not be that much of a big deal in this case, not sure.

I wonder if https://react.dev/reference/react/useTransition could help here? Not sure how it would integrate with Orbit controlled/uncontrolled support thought. Might not worth it if it requires to add an alternate hook to `useControllableState`

patricklafrance:
> Good catch. I think you might be right. I'll add a test to check what happens when the tooltip is controlled and `onOpenChange` is triggered.
> 
> I think I might just turn the ref into a simple mutable variable that gets reset on every render.

There's already a hook for this: https://github.com/gsoft-inc/sg-orbit/blob/master/packages/components/src/shared/src/useCommittedRef.ts

Another thing you could do is use the `useControllableState` optionall `onChange` callback to update your ref value when the state value change.

That being said, that `ref` does looks more like a patch than really fixing underlying issue.

tjosepo:
I decided to change approach and modify the setter of `useControllableState` to accept an _updater function_.

It works just like `useState`'s updater function.

I added a few tests to make sure the `onOpenChange` fires properly when the Tooltip component is controlled.

I also added tests to verify that the new updater function works as intended.

Tell me if you like this approach.

tjosepo:
>I wonder if https://react.dev/reference/react/useTransition could help here? Not sure how it would integrate with Orbit controlled/uncontrolled support thought. Might not worth it if it requires to add an alternate hook to useControllableState

I don't think it would in this case, but I wonder how useTransition and controlled/uncontrolled components would play out. 🤔

The new React documentation has [a very vague warning about using `ref.current` during rendering](https://react.dev/reference/react/useRef#caveats):
> Do not write or read `ref.current` during rendering, except for initialization. This makes your component’s behavior unpredictable.

No idea what it means by "makes your component’s behavior unpredictable", but I think it might be linked to supense & transitions. This may cause issues with hooks like `useControllableState` which rely a lot on refs. 

<sub>Maybe replacing the ref with a hook like [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) can solve the problem?</sub>

patricklafrance:
It's an interesting change.

The `updater` function idea kind of make sense. There's an issue thought in the current implementation because it doesn't take into account that a provided `onChange` function could return an updated state value.

The `onChange` function is kind of a mess, the type of issues that it solves should usually be adressed with a state machine but I never really understood how to integrate such a thing with the controlled/uncontrolled behavior.

Anyway, that's an issue for another time :)

Also, do you think it's a good idea that an `updater` function calls an handler? I feel like such a function should be pure?

tjosepo:
I'm having issues determining what should happen some issues with the behavior of `onOpenChange` when the tooltip is controlled.

Assuming we only want to fire onOpenChange when the value changes (e.g. when the tooltip is controlled and open, we only want to signal closing changes), but when the state changes twice in a single render, we don't know if we should fire the onOpenChange event again, because won't know if the `open` prop changed until the next render.

image


I looked at all our codebases with the RegEx `/<TooltipTrigger.*onOpenChange.*>/` to see if there was any usage of the TooltipTrigger component in one of our projects, but I couldn't find any.

I'm considering changing the behavior of onOpenChange events to always fire on mouseenter and mouseleave, regardless of the current `open` state.

tjosepo:
If we don't fire both change events, the mouseleave event _might_ be skipped causing and the tooltip will remain open indefinitely.

patricklafrance:
Hey @tjosepo 👋🏻 

> Assuming we only want to fire onOpenChange when the value changes (e.g. when the tooltip is controlled and open, we only want to signal closing changes)

Is it an assumption you are making to explain your proposition or are you considering it's always the case? I can imagine scenarios when an handler is listening to open/close state changes to track the state of an uncontrolled tooltip component.

> but when the state changes twice in a single render

AFAIK it shoudn't be possible, whenever a state prop change, a re-render is happening, the next state change is handled is a subsequent re-render. Maybe you are not refering to actual React state change here?

tjosepo:
> AFAIK it shoudn't be possible, whenever a state prop change, a re-render is happening, the next state change is handled by a subsequent re-render. Maybe you are not refering to actual React state change here?

Right, but because the callback functions of `mouseenter` and `mouseleave` compare the current state of `isOpen` with the new value **before** update the state, it's possible to skip updating the state when `onmouseenter` and `onmouseleave` happen in quick succession, faster than React has time to complete a full re-render and update the value of `isOpen`.

This is why I think the best approach is to change the state without doing a comparison with `isOpen` beforehand, as the value might be stale and there is no way to know if the state prop `open` will change value in the middle of a re-render.

In my latest commit, I did just that:

```diff
const updateIsOpen = useCallback((event: SyntheticEvent, newValue: boolean) => {
- if (isOpen !== newValue) {
    setIsOpen(newValue);
    
    if (!isNil(onOpenChange)) {
        onOpenChange(event, newValue);
    }
- }
}
```

alexasselin008:
> > AFAIK it shoudn't be possible, whenever a state prop change, a re-render is happening, the next state change is handled by a subsequent re-render. Maybe you are not refering to actual React state change here?
> 
> Right, but because the callback functions of `mouseenter` and `mouseleave` compare the current state of `isOpen` with the new value **before** update the state, it's possible to skip updating the state when `onmouseenter` and `onmouseleave` happen in quick succession, faster than React has time to complete a full re-render and update the value of `isOpen`.
> 
> This is why I think the best approach is to change the state without doing a comparison with `isOpen` beforehand, as the value might be stale and there is no way to know if the state prop `open` will change value in the middle of a re-render.
> 
> In my latest commit, I did just that:
> 
> ```diff
> const updateIsOpen = useCallback((event: SyntheticEvent, newValue: boolean) => {
> - if (isOpen !== newValue) {
>     setIsOpen(newValue);
>     
>     if (!isNil(onOpenChange)) {
>         onOpenChange(event, newValue);
>     }
> - }
> }
> ```

@patricklafrance Is the risk of firing the event twice really possible? the updateIsOpen method is only called from event handler, such as onShow, onHide or close. To me, if the event is fired twice with the same value, it would be an issue with the previous event handler firing twice and i would rather fix taht issue than having a redundant check?

patricklafrance:
Sounds reasonable to me, we can try it.

Could you somehow document your fix so there's a trace in the code? Not sure how thought since the fix is to remove code 😆 

Also, to accept this PR, could you remove changes to `useControllableState` ? It should be added in a distinct PR if we want to go this way.

tjosepo:
I updated the PR!

patricklafrance (approved):

Issue #1229, opened 2023-09-05
🐛 "An input component must either have a 'aria-label' attribute..." error when a <label> is present. 


tjosepo:
## Describe the bug

Input elements log an error about accessibility when `aria-label`, `aria-labelledby` and `placeholder` are missing.

There is no check for `<label>` elements, which makes some accessible inputs still log an error.

## Steps to reproduce

Create an element with this return:

```jsx
<>
    <Label htmlFor="test">Label</Label>
    <TextInput id="test" />
</>
```

## Expected results

There should be no errors. The text input has an accessible label.

PR #1230, opened 2023-09-05
Fix "An input component must either have a 'aria-label' attribute..." error when a <label> is present

tjosepo:
Issue: #1229

## Summary

Removes the warning "An input component must either have a 'aria-label' attribute..." when a `<label>` is present.

## What I did

I moved the validation logic inside a `useEffect` and add a check for `isNilOrEmpty(input.labels)`.

I had to modify `isNilOrEmpty` to support arrays (and NodeList). Instead of checking `value === ""`, it checks `value.length === 0` (this works for strings, arrays and NodeLists).

## How to test

Unit tests were added.

### DIFF
--- a/packages/components/src/number-input/src/NumberInput.tsx
+++ b/packages/components/src/number-input/src/NumberInput.tsx
@@ -1,6 +1,6 @@
 import { AbstractInputProps, adaptInputStylingProps, useInput, useInputIcon, useInputSpinner } from "../../input";
 import { Box, BoxProps } from "../../box";
-import { ChangeEvent, ComponentProps, FocusEvent, FocusEventHandler, MouseEvent, ReactElement, Ref, SyntheticEvent, forwardRef, useCallback, useMemo } from "react";
+import { ChangeEvent, ComponentProps, FocusEvent, FocusEventHandler, MouseEvent, ReactElement, Ref, SyntheticEvent, forwardRef, useCallback, useEffect, useMemo } from "react";
 import { Div, HtmlButton } from "../../html";
 import {
     OmitInternalProps,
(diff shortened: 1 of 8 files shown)


packages/components/src/number-input/src/NumberInput.tsx
```diff
@@ -335,6 +331,14 @@ export function InnerNumberInput(props: InnerNumberInputProps) {
         value: inputValueRef.current
     });
 
+    useEffect(() => {
```
patricklafrance, line 334: I believe this is textbook what React is recommending NOT to do with `useEffect`: https://react.dev/reference/react/useEffect#caveats

patricklafrance:
I am not sure I understand the use case correctly.

Does the following code would print a warning?

```
    <Field required>
        <Label>Username</Label>
        <TextInput placeholder="john@spacex.com" />
    </Field>
```

When there's a `Field` and a `Label`, the label id should be retrieved from the `FieldContext` by the following code:

```
const [fieldProps] = useFieldInputProps();
```

`fieldProps` would then add to the Input props the following properties:

```
"aria-describedby": !isGroup ? (!isNil(messageId) ? messageId : undefined) : undefined,
"aria-labelledby": !isGroup ? (!isNil(labelId) ? labelId : undefined) : undefined,
```

Which should prevent the warning from being displayed.

From what I can see by the [following test](https://github.com/gsoft-inc/sg-orbit/pull/1230/files#diff-aa97ee3489af05c30c9078e147382567006cd46edcc17349e3a4b866514d20a1R206), the issue might be that the Input and the Label are not wrapped within a `Field` component.

tjosepo:
> Does the following code would print a warning?

✅ I confirm that using `<Field>` worked without printing a warning.

I'll add it to the test scenarios.

---

I didn't know the `<Field>` component automatically added the accessibility properties.

To me, it still feel like a bug because using a label with `htmlFor` is accessible HTML, yet Orbit logs an error.

```tsx
<>
    <Label htmlFor="test">Label</Label>
    <TextInput id="test" />
</>
```

I understand that labels are this is not as easily verifyable as using `aria-label`, `aria-labelledby`, and `placeholder`, since this forces us to move the validation logic after the component has been rendered on the page. This prevents us from checking accessibility during SSR. Altough, it is just as unit-testable.

I think it's a valid trade-off.

patricklafrance:
Thank for the reply @tjosepo. Yes the intent is to always use the `<Field>` component to group together a Label and a Input. Great idea to update the warning message!

tjosepo:
Alright! I'll update my PR and go for it.

patricklafrance (approved):

PR #1231, opened 2023-09-26
feat(eslint-plugin): Created an eslint plugin to add rules and configuration

alexasselin008:
## Summary

Created a new eslint plugin that allows to use the design system with jsx-a11y plugin. In the future, this eslint plugin will include accessibility pointers, deprecation notices, and other helpful tips

I updated the installation documentation to reflect this new plugin

### DIFF
--- a/docs/getting-started/Installation.stories.mdx
+++ b/docs/getting-started/Installation.stories.mdx
@@ -43,6 +43,29 @@ To import Orbit experimental components styles, also add the following import de
 @import "@orbit-ui/experimental/index.css";
 ```
 
+## Set up your tooling environment
+
+Orbit offer an ESLint plugin to see in-context help in your IDE. This includes accessibility pointers, deprecation notices, and other helpful tips.
+
+First, add the ESlint plugin to your dependencies:
+
+<PackageInstallationSnippet workspaceFolder="tooling" packageName="eslint-plugin" />
+
+Second, add the plugin to your ESlint config. For example, your .eslintrc.json file may look like this:
+
+```json
+{
+    "$schema": "https://json.schemastore.org/eslintrc",
+    "plugins": ["@orbit-ui"],
+    "extends": [
+        // ...
+        "plugin:@orbit-ui/recommended"
+    ]
+}
+```
+
+More about [ESlint configuration](https://eslint.org/docs/latest/use/configure/configuration-files)
+
 ## Configure your application
 
 Below is an example of how to configure an application with a [pre-constructed](?path=/story/theming--page#option-3-retrieve-a-pre-constructed-theme-from-orbit) ShareGate theme object:
--- a/tooling/eslint-plugin/lib/config/jsx-a11y.ts
+++ b/tooling/eslint-plugin/lib/config/jsx-a11y.ts
@@ -1,0 +1,86 @@
+import type { Linter } from "eslint";
+import { sourceFiles } from "../utils/patterns";
+
+const config: Linter.Config = {
+    overrides: [
+        {
+            files: sourceFiles,
+            plugins: ["jsx-a11y"],
+            parserOptions: {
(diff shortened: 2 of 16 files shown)


docs/getting-started/Installation.stories.mdx
```diff
@@ -43,6 +43,29 @@ To import Orbit experimental components styles, also add the following import de
...
 
+## Set up your tooling environment
+
+Orbit offer an ESLint plugin to see in-context help in your IDE. This includes accessibility pointers, deprecation notices, and other helpful tips.
+
+First, add the ESlint plugin to your dependencies:
+
+<PackageInstallationSnippet workspaceFolder="tooling" packageName="eslint-plugin" />
+
+Second, add the plugin to your ESlint config. For example, your .eslintrc.json file may look like this:
```
patricklafrance, line 54: .eslintrc.json -> `.eslintrc.json`

docs/getting-started/Installation.stories.mdx
```diff
@@ -43,6 +43,29 @@ To import Orbit experimental components styles, also add the following import de
...
+
+```json
+{
+    "$schema": "https://json.schemastore.org/eslintrc",
+    "plugins": ["@orbit-ui"],
+    "extends": [
+        // ...
```
patricklafrance, line 61: Personally, for the new documentation that I am writting, I dont add those "..." comment anymore as it doesn't works well with copy/paste.

tooling/eslint-plugin/lib/config/jsx-a11y.ts
```diff
@@ -0,0 +1,86 @@
+import type { Linter } from "eslint";
+import { sourceFiles } from "../utils/patterns";
+
+const config: Linter.Config = {
+    overrides: [
+        {
+            files: sourceFiles,
```
patricklafrance, line 7: As there is only a single plugin and I doubt there would be others, I don,t think it's necessary to extract the files pattern in another file.

(As a side note, I am not a fan of this pattern, even in the web-configs project as it always force to be switch between 2 files to understand what files the sections is applying the rules for)

tooling/eslint-plugin/package.json
```diff
@@ -0,0 +1,69 @@
...
+    },
+    "dependencies": {
+        "eslint-plugin-jsx-a11y": "^6.7.1"
+    },
+    "peerDependencies": {
+        "eslint": "*"
+    },
+    "peerDependenciesMeta": {
+        "eslint": {
+            "optional": true
```
patricklafrance, line 45: I am not sure why `eslint` would be an optional dependency?

I notice that our web-configs projet also mark the `eslint` dependency as "optional". This is so annoying that package files doesn't allow comments!!!

docs/getting-started/Installation.stories.mdx
patricklafrance: Not a big deal but since I wrote a majority of the documentation and prefered "Then" to "Second", to keep the same tone accross our documentation site, it might be a good idea to prefer "Then".

Therefore, "Second" -> "Then"

patricklafrance:
The new orbit plugin should be configured in the Orbit repository to dogfood our own plugin :)

tooling/eslint-plugin/lib/config/jsx-a11y.ts
patricklafrance: Personally, this type of reusability enters the category of unnecessary coupling that could be replaced by copy/paste. I won't make a big deal of it thought for the web-configs project. Neither for this project, but it seems even more unnecessary here since it won't even be reused.

To answer your question, the only way I can think of that could make it easier for me would be if the name of the variables would include every file type that it covers.

patricklafrance (approved):

PR #1235, opened 2023-10-24
added banner to redirect to Orbiter when suitable

fraincs:
Issue: 

## Summary

As we have two "similar" Design System there was a need to make sure everyone uses the right Design System.

## What I did

Added a banner in the installation page warning users about which DS they should use.

### DIFF
--- a/.storybook/components/banner/Banner.css
+++ b/.storybook/components/banner/Banner.css
@@ -1,0 +1,5 @@
+.o-ui-sb-banner {
+    background: var(--o-ui-bg-alias-warning-light);
+    border-radius: var(--o-ui-br-3);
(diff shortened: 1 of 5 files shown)


vicky-comeau (approved):

.storybook/components/banner/Banner.css
```diff
@@ -0,0 +1,5 @@
+.o-ui-sb-banner {
+    background: var(--o-ui-bg-alias-warning-light);
+    border-radius: var(--o-ui-br-3);
+    color: var(--hop-decorative-option6-text);
```
vicky-comeau, line 4: shouldn't be a hopper color

vicky-comeau (approved):

